PR #76, opened 2026-04-15
feat(editor): tag autocomplete on # (#68)

### LINKED ISSUE #68: Tag autocomplete in editor (#)
## Summary

Add a CodeMirror 6 `CompletionSource` that triggers on `#` at word start and offers an autocomplete dropdown over the current vault tag set. Filter narrows as the user types, matching Obsidian's tag completion UX.

## Context

Wiki-link autocomplete (`[[`) is already implemented in `src/components/Editor/wikiLinkAutocomplete.ts` using a CM6 `CompletionSource` registered via override in `src/components/Editor/extensions.ts`. The frontend tag list is already populated into `src/store/tagsStore.ts` (fed by `list_tags` / `TagIndex` from `src-tauri/src/commands/tags.rs`). There is currently no equivalent autocomplete for `#` tags, so users must remember and type tag names by hand.

This issue adds a tag-specific `CompletionSource` mirroring the structure of the wiki-link one, but sourced reactively from `tagsStore` (no extra IPC round-trip needed since the store is already kept in sync).

## Acceptance Criteria

- [ ] New CM6 `CompletionSource` recognising `#` at word start (not inside code fences or URLs)
- [ ] Candidates pulled from `tagsStore` (reactively, so newly created tags appear without restart)
- [ ] Prefix filter narrows candidates as the user types
- [ ] Enter inserts `#tag`
- [ ] Escape dismisses the popup

## Out of Scope

- Hierarchical tag autocomplete beyond simple prefix (e.g. nested `#parent/child` tree navigation)
- Tag-creation-on-Enter side effects (updating `TagIndex`, writing back to files, etc.)

## Effort estimate: S

herox215:
Closes #68

## Summary
- New CM6 `CompletionSource` (`tagAutocomplete.ts`) that fires on `#` at a word boundary, sourced reactively from `tagsStore` so newly created tags appear without restart.
- Guards against firing inside fenced code / inline code (syntax-tree check mirroring `wikiLink.ts`), URL fragments, and bare `#` at line start (avoids hijacking markdown headings).
- Registered alongside the existing wiki-link source in `extensions.ts`; prefix filtering handled by CM6, Enter inserts the tag, Escape dismisses (both via CM6 defaults).

## Test plan
- [ ] typecheck passes
- [ ] tests pass
- [ ] `#` autocomplete appears in editor, filters on prefix, Enter inserts, Escape dismisses

### DIFF
--- a/src/components/Editor/EditorPane.svelte
+++ b/src/components/Editor/EditorPane.svelte
@@ -22,6 +22,7 @@
   import { scrollToMatch } from "./flashHighlight";
   import { scrollStore } from "../../store/scrollStore";
   import { treeRefreshStore } from "../../store/treeRefreshStore";
+  import { tagsStore } from "../../store/tagsStore";
   import { tabReloadStore } from "../../store/tabReloadStore";
   import { setResolvedLinks, resolveTarget, refreshWikiLinks } from "./wikiLink";
   import { setResolvedAttachments } from "./embeds";
@@ -447,6 +448,7 @@
           editorStore.setLastSavedHash(newHash);
           tabStore.setLastSavedContent(tab.id, result.merged_content);
           tabStore.setDirty(tab.id, false);
+          void tagsStore.reload();
 
           // Toasts — reuse the exact Phase 2 German strings.
           const filename = tab.filePath.split("/").pop() ?? tab.filePath;
@@ -463,6 +465,7 @@
         tabStore.setDirty(tab.id, false);
         tabStore.setLastSavedContent(tab.id, text);
         editorStore.setLastSavedHash(hash);
+        void tagsStore.reload();
       } catch (err: unknown) {
         // ERR-04: disk-full error — preserve buffer, debounce toast
         const isVaultErr = err && typeof err === "object" && "kind" in err;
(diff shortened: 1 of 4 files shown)


herox215:
## Bug: tag list not live — requires restart

Acceptance criterion *"Candidates pulled from `tagsStore` (reactively, so newly created tags appear without restart)"* is not met.

**Repro:**
1. In an open note, type a new tag like `#hallo` and save.
2. On another line, type `#` and trigger the autocomplete popup.
3. The tag `hallo` does not appear.
4. Restart the app — now `#hallo` shows up in the popup.

**Expected:** the tag appears in the popup immediately after it is written/saved, without needing a restart.

**Suspected cause:** `tagsStore` is likely not being refreshed after a save, or the completion source is reading a stale snapshot. Needs investigation into whether `list_tags` / `TagIndex` is reindexed on save and whether the store subscribes to that event.

PR should not be merged until this is resolved.

herox215:
Bug fixed in commit 4bd5d0c — `tagsStore.reload()` is now called after every successful save. New tags appear in the autocomplete popup immediately.

PR #320, opened 2026-04-21
feat(templates): share note-content cache across embed + vault API (#319)

### LINKED ISSUE #319: Sync readNoteContent for template expressions (LRU cache + per-file cap)
## Problem

Template expressions like `{{vault.notes.where(n => n.content.contains(\"Maestro\"))}}` return `collection(0)` (empty) even when notes matching the predicate exist on disk.

Root cause: `readNoteContent` in `src/lib/vaultApiStoreBridge.ts` only returns content for `editorStore.activePath` (the currently-open tab). For all other notes it returns `null`, which is coalesced to `\"\"` in the note getter — so any `.contains(...)` / `.startsWith(...)` check fails for unopened notes.

This is the explicit `#283` follow-up marked in the code comment: *\"In-memory only: the active editor tab. Non-active notes have no content available synchronously.\"*

## Proposal

Implement **synchronous read of note content with an LRU cache**, invalidated via the existing file watcher.

### Behavior

1. `readNoteContent(relPath)`:
   - If `relPath === editorStore.activePath` → return `editorStore.content` (reflects unsaved edits, as today).
   - Else → look up cache; on miss, `readTextFileSync` (Tauri) and store.
2. **LRU cap:** 1000 entries.
3. **Per-file cap:** skip caching files > 256 KB (still readable live, just not retained).
4. **Invalidation:** wire into the existing watcher (modify/delete/rename → drop entry).

### RAM expectations

- Typical vault (500–2000 notes × ~5 KB): **1–10 MB** cache footprint.
- 10k-note vault with LRU=1000: capped at **~10 MB** regardless of total vault size.

### Trade-off

First `vault.notes.where(...)` call in a cold vault triggers an I/O burst (N sync reads). Acceptable for typical vault sizes; optional idle warm-up is a separate follow-up.

## Acceptance criteria

- [ ] `{{vault.notes.where(n => n.content.contains(\"X\"))}}` returns all matching notes, not just the active one
- [ ] Editing an unopened note on disk updates the cache and re-evaluates dependent templates
- [ ] LRU cap prevents unbounded memory growth
- [ ] Files larger than the per-file cap are not cached but still read live
- [ ] Existing tests pass; new tests cover cache hit/miss, invalidation, LRU eviction, size cap

## Out of scope

- Async evaluator rewrite (rejected as too invasive)
- Background warm-up indexer
- Binary file handling

herox215:
Closes #319.

## Summary

- `{{vault.notes.where(n => n.content.contains(\"X\"))}}` now matches every note in the vault, not just the active editor tab.
- Extracted the embed plugin's in-line note-content cache into `src/lib/noteContentCache.ts` so the `\`n.content\`` template path and the `\`![[…]]\`` embed path share one source of truth. Invalidation (watcher + autosave diff + vault switch) moves with it.
- New safety rails on the shared cache: LRU cap (1000 entries), per-file cap (256 KB), concurrency cap (8 in-flight fetches), generation tokens against delete-during-load / vault-switch-during-load races, and a microtask-batched version store so N parallel completions trigger one re-render tick.

## Why

`readNoteContent` in `vaultApiStoreBridge.ts` only served the currently-open tab; every other note resolved to `\"\"`, so `.contains(...)` was always false for unopened notes. This also meant the embed cache and the template layer had two separate caches with independent invalidation — the pre-condition for silent drift if one branch got fixed and the other didn't.

## Test plan

- [x] `pnpm test` — 1103 / 1103 pass (42 new)
- [x] `pnpm build` — clean
- [x] UAT: open `~/Documents/NMS/2026-04-21.md`, confirm the expression returns the list of Maestro notes within ~one render tick (after async file reads land)
- [x] UAT: edit an unrelated note in another pane, save, confirm the expression re-evaluates
- [x] UAT: regression check — `![[Foo]]` note embeds still render, still refresh on external edit

## Known follow-ups

- `.first()` / `.any()` may be non-deterministic during warm-up until every touched note has been read once. Tracked as a follow-up; not in scope here.
- No eager bulk pre-fetch at vault open — current design is lazy. Could layer on a background indexer later.
- Tantivy BM25 fast-path for `.contains(...)` is an independent optimization opportunity.

### DIFF
--- a/src/components/Editor/embedPlugin.ts
+++ b/src/components/Editor/embedPlugin.ts
@@ -36,6 +36,11 @@ import { vaultStore } from "../../store/vaultStore";
 import { tabStore } from "../../store/tabStore";
 import { readFile } from "../../ipc/commands";
 import { listenFileChange } from "../../ipc/events";
+import {
+  readCached as readCachedNote,
+  requestLoad as requestLoadNote,
+  onCacheChanged as onNoteCacheChanged,
+} from "../../lib/noteContentCache";
 import { parseCanvas } from "../../lib/canvas/parse";
 import type { CanvasNode } from "../../lib/canvas/types";
 import { DEFAULT_NODE_WIDTH, DEFAULT_NODE_HEIGHT } from "../../lib/canvas/types";
@@ -57,23 +62,12 @@ const WIKI_EMBED_RE = /!\[\[([^\]|#]+?)(?:#([^\]|]+))?(?:\|([^\]]*))?\]\]/g;
 /** Matches `![alt](path)` — captures the path. */
 const MD_IMAGE_RE = /!\[[^\]]*\]\(([^)]+)\)/g;
 
-// ── Cache for note-embed content ──────────────────────────────────────────────
-
-/**
- * Vault-relative path → file content. Entries are invalidated selectively by
- * two module-level subscriptions installed at import time:
- *   - `listenFileChange` for external modifications the watcher sees
- *   - `tabStore.subscribe` for internal auto-saves (which suppress the watcher
- *     via write_ignore, so we'd otherwise miss them)
- * Previously the cache was cleared on every local `docChanged`, which caused
- * the `…` placeholder to flash on every keystroke — see #27.
- */
-const noteContentCache: Map<string, string> = new Map();
-/**
- * Paths whose `readFile` call is already in flight, used to suppress duplicate
- * IPC requests from back-to-back plugin rebuilds.
- */
-const noteFetchInFlight: Set<string> = new Set();
+// ── Cache for canvas-embed content ────────────────────────────────────────────
+//
+// Note-embed content moved to `src/lib/noteContentCache.ts` (#319) so the
+// template expression path (`n.content`) and the embed path share a single
+// source of truth. Canvas JSON keeps its own cache here — it's only consumed
+// by this plugin, not by the vault API.
 
 /**
  * Canvas embeds (#147) share the same caching contract as note embeds (#27):
@@ -128,12 +122,12 @@ function toVaultRel(absPath: string): string | null {
   return absFwd.slice(vaultFwd.length + 1);
 }
 
-// ── Cache invalidation: external file changes ─────────────────────────────────
+// ── Canvas-cache invalidation: external file changes ─────────────────────────
 //
-// The watcher only fires for modifications NOT initiated by our own IPC writes
-// (write_ignore suppresses self-writes). That means this catches edits made by
-// tools outside the app — Finder rename, shell `cp`, other editors. Internal
-// auto-saves are handled by the tabStore subscription below.
+// Note-content invalidation lives in `noteContentCache.ts` now. This handler
+// keeps the canvas-specific portion of the old behaviour in place: external
+// watcher events (Finder rename, shell `cp`, other editors) drop the cached
+// JSON for the affected path and kick mounted views.
 //
 // Guarded try/catch because the subscription is module-level and the test
 // environment has no Tauri IPC — a synchronous throw here would break imports.
@@ -141,19 +135,11 @@ try {
   void listenFileChange((payload) => {
     let changed = false;
     const rel = toVaultRel(payload.path);
-    if (rel !== null) {
-      if (noteContentCache.delete(rel)) changed = true;
-      if (canvasContentCache.delete(rel)) changed = true;
-    }
+    if (rel !== null && canvasContentCache.delete(rel)) changed = true;
     if (payload.new_path) {
       const newRel = toVaultRel(payload.new_path);
-      if (newRel !== null) {
-        if (noteContentCache.delete(newRel)) changed = true;
-        if (canvasContentCache.delete(newRel)) changed = true;
-      }
+      if (newRel !== null && canvasContentCache.delete(newRel)) changed = true;
     }
-    // #154: refresh inline canvas previews (and note embeds) in every open
-    // editor without waiting for the next keystroke.
     if (changed) kickAllEmbedViews();
   }).catch(() => {
     /* Tauri not initialized — tests run without the IPC backend. */
@@ -162,59 +148,33 @@ try {
   /* same reason — swallow so the module still loads under vitest. */
 }
 
-// ── Cache invalidation: internal auto-saves via other tabs ────────────────────
+// ── Canvas-cache invalidation: internal auto-saves via other tabs ────────────
 //
-// tabStore.setLastSavedContent fires after writeFile returns successfully. We
-// snapshot each tab's lastSavedContent the first time we see it and diff on
-// subsequent store emissions — when the snapshot changes, the file was just
-// saved (by any tab, including ones in other panes), so invalidate its cache
-// entry. This lets the next embed rebuild re-fetch the fresh content.
-const lastSavedByTabId: Map<string, string> = new Map();
+// Mirror of the original snapshot-diff pattern (#154). Note-content writes
+// fire the same snapshot change but are now handled in noteContentCache.ts;
+// this handler only touches the canvas-JSON side.
+const lastSavedCanvasByTabId: Map<string, string> = new Map();
 tabStore.subscribe((state) => {
   let changed = false;
   for (const tab of state.tabs) {
-    const prev = lastSavedByTabId.get(tab.id);
+    const prev = lastSavedCanvasByTabId.get(tab.id);
     if (prev !== tab.lastSavedContent) {
-      lastSavedByTabId.set(tab.id, tab.lastSavedContent);
+      lastSavedCanvasByTabId.set(tab.id, tab.lastSavedContent);
       const rel = toVaultRel(tab.filePath);
-      if (rel !== null) {
-        if (noteContentCache.delete(rel)) changed = true;
-        // #154: CanvasView calls setLastSavedContent after every autosave, so
-        // the same diff-on-snapshot path that catches note writes now catches
-        // canvas writes too — the watcher's write_ignore would otherwise hide
-        // our own saves from the file-change listener.
-        if (canvasContentCache.delete(rel)) changed = true;
-      }
+      if (rel !== null && canvasContentCache.delete(rel)) changed = true;
     }
   }
-  // Clean up snapshots for tabs that have closed so the map doesn't grow.
   const liveIds = new Set(state.tabs.map((t) => t.id));
-  for (const id of Array.from(lastSavedByTabId.keys())) {
-    if (!liveIds.has(id)) lastSavedByTabId.delete(id);
+  for (const id of Array.from(lastSavedCanvasByTabId.keys())) {
+    if (!liveIds.has(id)) lastSavedCanvasByTabId.delete(id);
   }
   if (changed) kickAllEmbedViews();
 });
 
-function scheduleNoteFetch(view: EditorView, relPath: string): void {
-  if (noteContentCache.has(relPath) || noteFetchInFlight.has(relPath)) return;
-  const vault = get(vaultStore).currentPath;
-  if (!vault) return;
-  const abs = `${vault}/${relPath}`;
-  noteFetchInFlight.add(relPath);
-  void readFile(abs)
-    .then((content) => {
-      noteContentCache.set(relPath, content);
-    })
-    .catch(() => {
-      noteContentCache.set(relPath, "");
-    })
-    .finally(() => {
-      noteFetchInFlight.delete(relPath);
-      // Kick the plugin to re-render now that we have content.
-      if (!view.dom.isConnected) return;
-      view.dispatch({ effects: [] });
-    });
-}
+// #319: the shared note-content cache lives in its own module. Subscribe to
+// its change signal so embed decorations refresh when a background fetch
+// lands — same UX as the legacy per-view dispatch had.
+onNoteCacheChanged(kickAllEmbedViews);
 
 /**
  * #154 — mirror of scheduleNoteFetch for `.canvas` embeds. Runs when
@@ -622,9 +582,10 @@ function buildDecorations(view: EditorView): DecorationSet {
             deco = Decoration.replace({ widget: new CanvasEmbedWidget(rel, sizePx, cachedCanvas) });
           }
         } else {
-          const cached = noteContentCache.get(rel);
-          if (cached === undefined) {
-            scheduleNoteFetch(view, rel);
+          // #319: note-content cache is shared with the vault API.
+          const cached = readCachedNote(rel);
+          if (cached === null) {
+            requestLoadNote(rel);
             // Show a muted placeholder while the fetch is in flight so the UI
             // doesn't flash with a stale version of the target note.
             deco = Decoration.replace({ widget: new NoteEmbedWidget(rel, "…") });
@@ -701,9 +662,10 @@ export const embedPlugin = ViewPlugin.fromClass(
     private view: EditorView;
 
     constructor(view: EditorView) {
-      // Clear caches so stale content from a previous vault does not leak into
-      // the freshly-mounted view. Cheap because the caches are module-level.
-      noteContentCache.clear();
+      // Canvas cache clears on mount so stale JSON from a previous vault
+      // doesn't leak into the freshly-mounted view. The note-content cache
+      // is managed by `noteContentCache.ts` and already clears on vault
+      // switch via its own vaultStore subscription.
       canvasContentCache.clear();
       this.view = view;
       // #154: register so module-level invalidation hooks can kick a rebuild.
@@ -730,12 +692,12 @@ export const embedPlugin = ViewPlugin.fromClass(
 );
 
 /**
- * Test-only hook: clear both caches so unit tests start from a clean slate.
+ * Test-only hook: clear the canvas cache so unit tests start from a clean
+ * slate. The note-content cache lives in its own module now — tests that
+ * exercise that surface should reset it via `__cacheForTests.reset()`.
  * Not part of the public API.
  */
 export function __resetEmbedCachesForTests(): void {
-  noteContentCache.clear();
-  noteFetchInFlight.clear();
   canvasContentCache.clear();
 }
 
--- a/src/components/Editor/embedPlugin.ts
+++ b/src/components/Editor/embedPlugin.ts
@@ -39,8 +39,9 @@ import { listenFileChange } from "../../ipc/events";
 import {
   readCached as readCachedNote,
   requestLoad as requestLoadNote,
-  onCacheChanged as onNoteCacheChanged,
+  noteContentCacheVersion,
 } from "../../lib/noteContentCache";
+import { toVaultRel as toVaultRelHelper, absFromRel as absFromRelHelper } from "../../lib/vaultPath";
 import { parseCanvas } from "../../lib/canvas/parse";
 import type { CanvasNode } from "../../lib/canvas/types";
 import { DEFAULT_NODE_WIDTH, DEFAULT_NODE_HEIGHT } from "../../lib/canvas/types";
@@ -110,16 +111,11 @@ function kickAllEmbedViews(): void {
 
 /**
  * Convert an absolute path to a vault-relative forward-slash path, or return
- * null when the path is outside the vault (or no vault is open).
+ * null when the path is outside the vault (or no vault is open). Thin
+ * wrapper around the shared helper so all call-sites here stay terse.
  */
 function toVaultRel(absPath: string): string | null {
-  const vault = get(vaultStore).currentPath;
-  if (!vault) return null;
-  const absFwd = absPath.replace(/\\/g, "/");
-  const vaultFwd = vault.replace(/\\/g, "/").replace(/\/$/, "");
-  if (absFwd === vaultFwd) return "";
-  if (!absFwd.startsWith(vaultFwd + "/")) return null;
-  return absFwd.slice(vaultFwd.length + 1);
+  return toVaultRelHelper(absPath, get(vaultStore).currentPath ?? null);
 }
 
 // ── Canvas-cache invalidation: external file changes ─────────────────────────
@@ -172,9 +168,15 @@ tabStore.subscribe((state) => {
 });
 
 // #319: the shared note-content cache lives in its own module. Subscribe to
-// its change signal so embed decorations refresh when a background fetch
-// lands — same UX as the legacy per-view dispatch had.
-onNoteCacheChanged(kickAllEmbedViews);
+// its version store so embed decorations refresh when a background fetch
+// lands — same UX as the legacy per-view dispatch had. Svelte stores fire
+// synchronously with the current value on subscribe; a `ready` latch
+// swallows that initial call so we don't kick on module import.
+let noteCacheReady = false;
+noteContentCacheVersion.subscribe(() => {
+  if (noteCacheReady) kickAllEmbedViews();
+});
+noteCacheReady = true;
 
 /**
  * #154 — mirror of scheduleNoteFetch for `.canvas` embeds. Runs when
@@ -184,9 +186,8 @@ onNoteCacheChanged(kickAllEmbedViews);
  */
 function scheduleCanvasFetch(view: EditorView, relPath: string): void {
   if (canvasContentCache.has(relPath) || canvasFetchInFlight.has(relPath)) return;
-  const vault = get(vaultStore).currentPath;
-  if (!vault) return;
-  const abs = `${vault}/${relPath}`;
+  const abs = absFromRel(relPath);
+  if (abs === null) return;
   canvasFetchInFlight.add(relPath);
   void readFile(abs)
     .then((content) => {
@@ -218,10 +219,7 @@ function isImageFilename(name: string): boolean {
  * `convertFileSrc` tolerates either.
  */
 function absFromRel(relPath: string): string | null {
-  const vault = get(vaultStore).currentPath;
-  if (!vault) return null;
-  const v = vault.replace(/\\/g, "/").replace(/\/$/, "");
-  return `${v}/${relPath}`;
+  return absFromRelHelper(relPath, get(vaultStore).currentPath ?? null);
 }
 
 class ImageEmbedWidget extends WidgetType {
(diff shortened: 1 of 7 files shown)


herox215 (commented):
**Review verdict: request changes (self-review — GH blocks formal request-changes on own PRs, flagging here instead).**

Reviewed the shared note-content cache extraction. The direction is right — consolidating the two caches onto one invalidation surface is the correct fix for #319. But there are two behavioural blockers in the core cache/invalidation logic, plus a handful of should-fix and nits. Blockers must be addressed before merge.

Summary:
- **Blockers (2):** `requestLoad` bypasses generation on empty-string cache hits (delete-during-load ghosts can still resurrect); `invalidate` also bumps the version for misses, wasting re-render ticks on invalidations for files not in the cache.
- **Should-fix (9):** clarity/correctness tweaks around `clear()`, LRU touch on miss, error path caching, `wireOnce()` ergonomics, test coverage for LRU eviction and vault-switch race, etc.
- **Nits/questions posted as a grouped discussion comment.**

Full line-level comments follow.

src/lib/noteContentCache.ts
```diff
@@ -0,0 +1,286 @@
...
+  cache.set(rel, value);
+  return value;
+}
+
+export function requestLoad(rel: string): void {
+  if (cache.has(rel) || inFlight.has(rel)) return;
+  if (queue.indexOf(rel) !== -1) return;
+  queue.push(rel);
+  pump();
+}
```
herox215, line 112: **Blocker —** `requestLoad` short-circuits on `cache.has(rel)` without consulting `generation`. A sequence like (1) load starts, (2) `invalidate(rel)` bumps the generation and removes the entry, (3) file is deleted on disk, (4) the in-flight resolve sees `generation` matches its captured `gen` (no further bump happened) and writes content back via `setCache` — ghost resurrection. The generation token in `doFetch` defends against `clear()` and against a *second* `invalidate` landing during the fetch, but not against this exact interleaving where `requestLoad` is re-called for the same key after the invalidate (which would be a no-op here because `inFlight.has(rel)` is still true). Worse case: an evicted LRU entry gets re-fetched after the file was deleted because `requestLoad` never sees the deletion (the watcher invalidate lands before the fetch completes, but the generation check at line 181 uses the `gen` captured *before* the invalidate, so the stale content wins).

Concretely: also capture and compare generation inside `requestLoad`, or at minimum bump generation inside `invalidate` *and* ensure `doFetch` captures generation *after* adding to `inFlight` under a single atomic step. Recommend: in `doFetch`, capture `const gen = bumpGeneration(rel) - 1` semantics — or capture `generation.get(rel) ?? 0` only after `inFlight.add`. The current ordering at lines 176-177 is racy: generation is read, *then* inFlight is set. A synchronous `invalidate` call between those two statements is impossible (single-threaded), so the actual bug is subtler — it is the interleaving where `invalidate` lands *between* `requestLoad` enqueueing and `doFetch` starting. The fix is to re-check generation one more time just before `setCache` at line 183, using a generation captured as late as possible.

src/lib/noteContentCache.ts
```diff
@@ -0,0 +1,286 @@
...
+
+export function invalidate(rel: string): boolean {
+  bumpGeneration(rel);
+  const hadEntry = cache.delete(rel);
+  if (hadEntry) scheduleBump();
+  return hadEntry;
+}
```
herox215, line 124: **Blocker —** `invalidate` always calls `scheduleBump()` only when `hadEntry` is true, which is correct — but `bumpGeneration(rel)` is called *unconditionally*. That means invalidations for keys that were never cached (e.g. a watcher event for a file nobody has looked at) still pollute the `generation` map unboundedly. Across a long-running session with many rename/delete watcher events, `generation` grows without bound because nothing ever GCs stale entries whose `cache`/`inFlight`/`queue` presence has long since been gone. Even if this is only a slow memory leak rather than a correctness bug, combined with B1 the interaction is unpredictable.

Fix: skip `bumpGeneration` when `!cache.has(rel) && !inFlight.has(rel) && queue.indexOf(rel) === -1` — there is nothing in flight to defend against. Or: lazily prune `generation` entries on every `setCache` / `invalidate` tick. Also double-check: is it intentional that `invalidate` bumps the version store only on a hit? Rationale should be comment-documented — the current code reads as "miss = silent no-op", which might surprise callers relying on invalidate to trigger a re-render.

src/lib/noteContentCache.ts
```diff
@@ -0,0 +1,286 @@
...
+
+export function clear(): void {
+  // Bump generation for every key the module knows about so any resolving
+  // in-flight fetches discard themselves instead of re-populating the fresh
+  // post-clear cache with stale content.
```
herox215: **Should-fix —** The `seen` set aggregation is correct intent, but this entire block is defensive rather than the actual source of protection. The real defense is in `doFetch` line 181: `if ((generation.get(rel) ?? 0) !== gen) return;`. Which means the loop over `seen` is effectively a "bump every key that might have an outstanding fetch so the post-await check fails" — which is fine, but the comment should spell out that this is the mechanism. Also worth asserting: what about keys in `queue` whose `doFetch` has not yet run? Those will still see their bumped generation at the top of `doFetch` (line 176 captures *current* gen post-bump), so they get a cleanly re-bumped token and will not discard — meaning a key that was in the queue at clear-time will happily populate the cache after clear(). Is that intentional? Comment at line 127 ("any resolving in-flight fetches discard themselves") overstates this — queued-but-not-yet-fetched keys *do not* discard. If the intent is "post-clear the cache is truly empty for X seconds", this is a latent bug. Suggest: empty `queue` fully (already done on line 135), which resolves it — but then the comment at 136-137 saying "inFlight is intentionally not touched" is the only defense, and that defense only covers *already-dispatched* fetches.

src/lib/noteContentCache.ts
```diff
@@ -0,0 +1,286 @@
...
+// --- Public API ---------------------------------------------------------
+
+export function readCached(rel: string): string | null {
+  const value = cache.get(rel);
+  if (value === undefined) return null;
+  // LRU touch.
+  cache.delete(rel);
+  cache.set(rel, value);
+  return value;
+}
```
herox215, line 106: **Should-fix —** `readCached` does an LRU touch (delete+reinsert) on every hit, which is fine, but two nits: (1) the module comment at line 25 says "approximate LRU" — this is *genuine* LRU-on-read (not approximate). The "approximate" wording implies the implementation cuts corners it does not actually cut, which is misleading for future readers. (2) The touch writes back to the Map on *every* read of a cached note, even in tight loops (e.g. a `.where()` over 1000 notes where each `.content.contains(...)` reads the cache). Map mutation on each iteration is measurable overhead. If the caller is deterministic and reads in a fixed order, the touch is pointless. Consider: only touch when the entry is older than half the LRU window, or skip the touch entirely if `cache.size < LRU_MAX_ENTRIES` (no eviction pressure, no reason to re-sort). The current behavior is correct but pessimistic; easy win.

src/lib/noteContentCache.ts
```diff
@@ -0,0 +1,286 @@
...
+    const content = await readFile(abs);
+    if ((generation.get(rel) ?? 0) !== gen) return;
+    if (content.length > MAX_FILE_BYTES) return;
+    setCache(rel, content);
+  } catch {
+    if ((generation.get(rel) ?? 0) === gen) {
+      // Mirror the pre-existing embed-plugin behaviour: failures cache as
+      // empty string so the UI doesn't retry on every render.
+      setCache(rel, "");
+    }
```
herox215, line 183: **Should-fix —** Caching `""` on *any* `readFile` exception permanently poisons the cache for that key until the next watcher event or manual invalidate. Transient IPC errors (backend briefly unavailable, slow disk, EMFILE) will stick as `""` for the lifetime of the cache entry. The module comment at line 38 rationalizes this as "mirrors pre-existing embed behavior" — but the previous embed code had tight per-render invalidation, whereas the shared cache persists across hundreds of renders and dozens of consumers. Worth either: (a) distinguishing "ENOENT / permission denied" from transient IO errors and only caching `""` for the former, (b) caching `""` with a shorter TTL / retry-on-next-access, or (c) explicitly caching a sentinel that lets callers know "we tried and failed" vs "file legitimately empty". Minimum: document the failure mode in a comment and link to the follow-up issue.

src/lib/noteContentCache.ts
```diff
@@ -0,0 +1,286 @@
...
+      return;
+    }
+    if (state.currentPath !== lastVaultPath) {
+      lastVaultPath = state.currentPath;
+      clear();
+    }
+  });
+}
+
+wireOnce();
```
herox215, line 297: **Should-fix —** `wireOnce()` runs at module-import time as a side effect. That means simply importing `noteContentCache.ts` from a test file subscribes to `tabStore` and `vaultStore`, registers a Tauri listener, and cannot be undone short of module-cache reset. Problems: (1) tests that import the module pick up live subscriptions they did not ask for — the try/catch around `listenFileChange` hides this, but `tabStore.subscribe` and `vaultStore.subscribe` still run; (2) any future test that wants to exercise invalidation logic in isolation cannot — the real subscriptions are already wired and will fire during test setup; (3) the `__cacheForTests.reset()` hook resets cache state but not subscription state (e.g. `lastSavedByTabId` captures stay, `lastVaultPath` captures stay). Suggest: keep `wireOnce()` callable but *only* invoke it from an explicit `init()` function called by the consumers (or an entry module), and let tests opt in. Current design couples module load to Svelte store subscriptions, which is awkward.

src/lib/noteContentCache.ts
```diff
@@ -0,0 +1,286 @@
...
+    void listenFileChange((payload) => {
+      const rel = toVaultRel(payload.path);
+      if (rel !== null) invalidate(rel);
+      if (payload.new_path) {
+        const newRel = toVaultRel(payload.new_path);
+        if (newRel !== null) invalidate(newRel);
+      }
+    }).catch(() => {
+      // Tauri backend missing — tests run without IPC.
+    });
```
herox215: **Should-fix —** The rename handling invalidates both `rel` (old path) and `newRel` (new path), which is correct for the invalidation side but leaves the new-path entry *un-populated* — the first template expression that touches the new path will see `null` and trigger an async `requestLoad`. If the watcher also emits a separate "created" event for the new path the async fetch races with it. Cleaner: on rename, if `cache.has(rel)` (old path content already cached), move the value to `newRel` directly rather than re-fetching. Edge case is that the content might have changed during the rename — but the rename-then-modify watcher event ordering is backend-dependent. Keep current behavior as a safe fallback, but consider the optimistic move for smoother UX. Also: `payload.new_path` may coincidentally equal `payload.path` in some backends — defensive `if (newRel !== rel) invalidate(newRel)` avoids double-work.

src/lib/__tests__/noteContentCache.test.ts
```diff
@@ -0,0 +1,340 @@
+// Unit tests for the shared note-content cache (#319). Exercise the LRU,
```
herox215, line 1: **Should-fix —** Test suite gaps to address before merge:
- No test for LRU eviction at `LRU_MAX_ENTRIES` boundary (inserting 1001 distinct keys → first inserted is gone, last inserted present). The LRU is load-bearing; an off-by-one there silently kills cache behavior in large vaults.
- No test for vault-switch-during-load race: start a fetch under vault A, switch `vaultStore.currentPath` to vault B, resolve the fetch, assert the cache does *not* contain the stale content. This exercises the exact scenario the `generation` token was introduced to defend against.
- No test for `MAX_FILE_BYTES` gating (file larger than the cap is fetched but not retained — currently only inferred from reading the impl).
- No test asserting `onCacheChanged` subscribers are called on invalidate / setCache but not redundantly (the comment mentions "isolate subscriber errors" — test that one throwing subscriber does not break others).
- No test for the `tabStore` auto-save diff path (line 234-247): simulate a tab whose `lastSavedContent` changes and assert `invalidate` fires for the corresponding rel.

These are all straightforward additions against the existing `__cacheForTests` hooks.

src/lib/__tests__/noteContentCache.test.ts
```diff
@@ -0,0 +1,340 @@
...
+    const { MAX_CONCURRENT } = __cacheForTests.limits();
+    // Load exactly one concurrency-batch worth so all completions land in
+    // the same microtask — across batches we'd legitimately get one bump
+    // per wave, which is still better than 1-per-resolve but not what this
+    // test asserts.
+    for (let i = 0; i < MAX_CONCURRENT; i++) requestLoad(`n${i}.md`);
+    await flushMicrotasks();
+    const bumped = get(noteContentCacheVersion);
+    expect(bumped - baseline).toBe(1);
+  });
```
herox215, line 283: **Should-fix —** The coalescing assertion `bumped - baseline === 1` is fragile: it assumes all `MAX_CONCURRENT` completions land in the same microtask, which depends on the JS engine scheduling `Promise.resolve(p)` such that all resolves drain in one microtask wave. In vitest / jsdom that is reliable today, but a change in the microtask flush implementation (or a future `await`-point added inside `doFetch`) would flip this to `=== 2` without the implementation being wrong. Suggest: assert `bumped - baseline <= 1` instead, and leave a comment explaining why the upper bound is 1 for this specific scenario. The sibling test at 226-237 uses `toBeLessThan(15)`, which is the right pattern — this test should mirror it.

src/lib/__tests__/noteContentCache.test.ts
```diff
@@ -0,0 +1,340 @@
...
+    const baseline = get(noteContentCacheVersion);
+    for (let i = 0; i < 50; i++) requestLoad(`n${i}.md`);
+    await flushMicrotasks();
+    const bumped = get(noteContentCacheVersion);
+    // 50 completions must NOT produce 50 bumps. Upper bound: one bump per
+    // concurrency wave (~50/8 = 7 waves) plus some slack for microtask
+    // timing. 50 would mean the batching is dead.
+    expect(bumped - baseline).toBeLessThan(15);
+    expect(bumped - baseline).toBeGreaterThan(0);
+  });
```
herox215, line 296: **Should-fix —** The `toBeLessThan(15)` slack is generous, which is fine as a "batching is alive" canary. But the lower bound `toBeGreaterThan(0)` is trivially met by any non-broken implementation — if you want this to be a meaningful regression guard, also assert `toBeGreaterThan(Math.floor(50 / MAX_CONCURRENT))` (i.e. at least one bump per concurrency wave), which would catch the opposite failure mode where some bumps go missing. Right now this test only fails if batching is completely dead *or* if it produces more than 15 bumps.

src/components/Editor/embedPlugin.ts
```diff
@@ -701,9 +662,10 @@ export const embedPlugin = ViewPlugin.fromClass(
...
 
     constructor(view: EditorView) {
-      // Clear caches so stale content from a previous vault does not leak into
-      // the freshly-mounted view. Cheap because the caches are module-level.
-      noteContentCache.clear();
+      // Canvas cache clears on mount so stale JSON from a previous vault
+      // doesn't leak into the freshly-mounted view. The note-content cache
+      // is managed by `noteContentCache.ts` and already clears on vault
+      // switch via its own vaultStore subscription.
       canvasContentCache.clear();
```
herox215, line 667: **Should-fix —** `canvasContentCache.clear()` on every ViewPlugin mount is aggressive. Prior to this PR the embed plugin also cleared its note-content cache here — now that the note cache is module-level, this clear is only about the canvas JSON cache. Fine. But: the canvas cache was sized for vault-switch-as-the-bound; remounting the ViewPlugin during a normal workflow (tab switch, editor reload) now wipes it too. Consider whether `canvasContentCache` needs the same vault-switch-invalidation treatment as `noteContentCache` so this on-mount clear can be removed — one invalidation strategy for all file-derived caches. At minimum, call it out in the comment: "canvas cache is ViewPlugin-lifetime scoped, not vault-lifetime scoped — by design" or "TODO: move to module-level with vault-switch invalidation like noteContentCache".

src/lib/noteContentCache.ts
```diff
@@ -0,0 +1,286 @@
...
+  reset: () => {
+    cache.clear();
+    inFlight.clear();
+    queue.length = 0;
+    activeFetches = 0;
+    generation.clear();
+    pendingBump = false;
+    versionStore.set(0);
+  },
+};
```
herox215, line 340: **Should-fix —** `__cacheForTests.reset()` is incomplete: it resets `cache`, `inFlight`, `queue`, `activeFetches`, `generation`, `pendingBump`, and `versionStore` — but **not**:
- `subscribers` (imperative `onCacheChanged` handlers registered in a previous test persist into the next)
- `lastSavedByTabId` (closed over inside `wireOnce`, inaccessible from here)
- `lastVaultPath` (same — closure-captured)
- `wired` flag (fine as-is, but worth commenting)

In practice this means test ordering matters: a test that registers an `onCacheChanged` subscriber leaks it into subsequent tests. Suggest: expose `__resetSubscribers()` and move `lastSavedByTabId`/`lastVaultPath` to module scope so `reset()` can clear them. Or: make `wireOnce` return a disposer that tests can call. Current state is a footgun waiting for the first test that adds a subscriber.

herox215:
**Nits and open questions (grouped — non-blocking discussion items)**

**Nits (N1–N6):**

- **N1** — The module header comment is excellent and sets expectations well. One small ask: explicitly list which files are *not* cached (e.g. binary attachments? `.md` only? anything matching a glob?). Right now the caller-side contract is "any vault-relative path", and it's not obvious from the cache whether binary/large-file handling is the cache's problem or the caller's.

- **N2** — `toVaultRel` and `absFromRel` duplicate the path-normalization logic (`.replace(/\\/g, "/").replace(/\/$/, "")`). Factor into a local `normalizeVaultPath(vault)` helper to avoid drift if the rules change (e.g. future UNC path support on Windows).

- **N3** — `notifyImperative()` uses `Array.from(subscribers)` to snapshot before iterating — good, it protects against subscriber-mutation-during-iteration. Worth a one-line comment explaining why the snapshot is necessary, because this pattern is easy to "clean up" in a future refactor.

- **N4** — The `LRU_MAX_ENTRIES = 1000`, `MAX_FILE_BYTES = 256 * 1024`, `MAX_CONCURRENT = 8` constants are reasonable defaults but not motivated anywhere. Add a line or two of rationale: "1000 covers typical vaults of N notes; 256 KB is 2x the 99th-percentile markdown file size observed in…". Future tuners need context.

- **N5** — `scheduleBump` uses `queueMicrotask`, which runs *before* the next paint but *after* the current synchronous work. That's the right choice for coalescing. Worth a comment noting that `Promise.resolve().then(...)` would be near-equivalent — reviewers and future-you will wonder why one and not the other. The tie-breaker is that `queueMicrotask` doesn't allocate a Promise, which matters at high throughput.

- **N6** — `setCache` evicts via `cache.keys().next().value`. On a freshly-cleared Map this is safe (the first key is the oldest), but if anyone ever adds an LRU-touch inside `setCache` itself (e.g. to "bump on write"), the invariant breaks silently. Add an assertion or a comment pinning the invariant.

---

**Open questions (Q1–Q5):**

- **Q1** — What's the plan for the "known non-goal" listed at line 35 (`.first()` / `.any()` non-determinism during warm-up)? Is there a tracking issue? Consumers doing `vault.notes.first(n => n.content.contains("X"))` will see different results across the first few renders until the cache settles — is that acceptable or should there be a "content-aware" variant that returns `undefined` until at least one pass has completed?

- **Q2** — The `tabStore` subscription diffs `lastSavedContent` per-tab (line 234-247). What happens if a tab is closed and re-opened with the same `tab.id`? The `lastSavedByTabId` Map still has the old entry (cleaned up on line 244 when the id disappears from `state.tabs`). If the store can emit a state with the tab absent and then present again within the same microtask wave, we miss the cleanup. Probably not a real scenario, but worth thinking through.

- **Q3** — `vaultStore.subscribe` at line 252 only triggers `clear()` on `currentPath` *change*, not on `currentPath` becoming `null` (vault closed). Is "vault closed" a distinct state that should clear? Current behavior: if user closes the vault without switching, the cache stays populated. On reopen it would be stale.

- **Q4** — `requestLoad` does `queue.indexOf(rel)` — O(n) scan. For the queue sizes we expect (≤ a few hundred during vault warm-up) that's fine, but if `.where()` triggers thousands of `requestLoad` calls in a tight loop, this becomes O(n²) overall. Benchmark? Or just swap for a `Set` side-index if it shows up in a profile.

- **Q5** — Is there a long-term plan to unify `canvasContentCache` (see S8 on `embedPlugin.ts:669`) into this module, or at least under the same invalidation source? Two caches with subtly different invalidation strategies is exactly the shape of bug this PR is fixing for #319 — we should avoid re-introducing it for canvas JSON.

src/lib/noteContentCache.ts
herox215: Fixed in 861a379. `clear()` now drops `inFlight` after bumping generation for each entry — the old fetch still resolves but its captured gen fails the check, and a fresh `requestLoad` re-enqueues immediately. Pinned by `clear mid-load releases the slot` test.

src/lib/noteContentCache.ts
herox215: Fixed in 861a379. `invalidate(rel)` now clears the `inFlight` slot for that key too. Pinned by `invalidate mid-load releases the slot` test.

src/lib/noteContentCache.ts
herox215: Fixed in 861a379 via a different approach: instead of clearing the generation map, `clear()` only bumps generations for in-flight keys (keys only in cache/queue had nothing pending). The map still grows per unique invalidation key, but that's bounded by vault size — a 10k-note vault ≈ sub-MB. Accepted as a bounded-memory trade for correctness of the generation sentinels.

src/lib/noteContentCache.ts
herox215: `readCached` still mutates for the LRU touch, but `clear()` no longer iterates `cache.keys()` — only `inFlight` (a `Set`). No re-entrancy risk anymore. The subscriber list was also removed entirely (see S9 thread / N2), cutting the only path that could have run arbitrary user code during iteration.

src/lib/noteContentCache.ts
herox215: Kept empty-string-on-failure for RAM/perf (no sentinel object per entry). Pinned the contract with a new test `treats failed reads as indistinguishable from empty files by design` and documented the `.contains("")` edge in the module header. Users who care can filter via `n.content.length > 0`.

src/lib/noteContentCache.ts
herox215: Fixed in 861a379. `wireOnce()` pushes the `listenFileChange` unlisten + both store unsubs into a `teardown` array; `import.meta.hot.dispose` drains it on HMR and resets the `wired` flag so a subsequent `wireOnce()` re-installs fresh handlers.

src/lib/noteContentCache.ts
herox215: Fixed in 861a379. `listenFileChange(...).then((unlisten) => teardown.push(unlisten))` captures the handle; HMR dispose calls every entry. Embed plugin still throws its unlisten away — tracked as a follow-up, not in scope for this PR.

src/lib/__tests__/noteContentCache.test.ts
herox215: Fixed in 861a379. `lastSavedByTabId` and `lastVaultPath` lifted to module scope; `__cacheForTests.reset()` wipes them too. Tests can flip vault paths without inheriting closure state from predecessors.

src/lib/__tests__/noteContentCache.test.ts
herox215: Tightened in 861a379: the single-batch test now asserts exact `=== 1` bump for one concurrency wave. Kept the 50-load test with the loose `< 15` bound as a sanity check — its job is only to catch a regression to per-completion bumping (50+).

src/lib/__tests__/noteContentCache.test.ts
herox215: Kept loose — see reply above. The test is a regression guard against per-completion bumping, not an exact behaviour pin. The new exact-assertion single-batch test covers the precise contract.

src/components/Editor/embedPlugin.ts
herox215: Acknowledged as pre-existing, not a PR regression. No change in this commit.

src/lib/noteContentCache.ts
herox215: Fixed in 861a379. `__cacheForTests` methods now call `assertTestEnv()` which throws outside vitest/NODE_ENV=test. Also completed the reset to include `lastSavedByTabId` and `lastVaultPath`.

herox215:
Addressing the nits + questions from the review (commit 861a379):

**N1** — Done. `toVaultRel` / `absFromRel` extracted to `src/lib/vaultPath.ts`, both `noteContentCache` and `embedPlugin` consume it.

**N2 / Q4** — Done. `onCacheChanged` removed; `embedPlugin` now subscribes to `noteContentCacheVersion` directly with the `ready` latch pattern that `templateLivePreview` already uses.

**N3** — Kept in module header; user docs are a separate follow-up.

**N4** — Added `console.debug` in dev when a file exceeds the per-file cap.

**N5** — Repo has no CHANGELOG convention; skipped.

**N6** — Done. Queue is now `Map<string, true>` for O(1) membership while preserving FIFO.

**Q1** — Keeping canvas + note caches split intentionally. Different content shape (JSON vs text), different consumer shape (only embed vs embed + template API), different invariants. Generalising would be maybe 40 lines saved at the cost of a narrower interface per consumer. Not worth it right now.

**Q2** — Kept `null` vs `""` semantics. A richer discriminated type would cost an allocation per miss and per failure, and the call-site in `vaultApiStoreBridge` would have to unwrap it every render. Performance + RAM considerations say stay simple.

**Q3** — Resolved by the B1/B2 fix. The race you described no longer exists because `invalidate` now releases the slot, so the "second `requestLoad` early-returns" step is gone.

**Q5** — `LRU_MAX_ENTRIES = 1000` is a heuristic: ~1–5 MB resident for typical vaults, capped independent of vault size. A 10k-note vault doing a full content-scan will thrash eviction once per render, but that's a corner case and the alternative (vault-size-aware caps) risks unbounded growth. Ship the heuristic; revisit if someone reports thrash.

PR #322, opened 2026-04-21
feat(reading): evaluate `{{ ... }}` template expressions in Reading Mode (#321)

### LINKED ISSUE #321: Evaluate `{{ ... }}` template expressions in Reading Mode
## Problem

Template expressions like

```
{{vault.folders.where(f => f.name == "Projects").first().notes.select(n => "[[" + n.name + "]]").join(", ")}}
```

evaluate live in the CodeMirror editor (`src/components/Editor/templateLivePreview.ts`, #283) but render as **literal `{{ ... }}` text** in Reading Mode.

Root cause: `ReadingView.svelte:41` pipes the raw markdown through `renderMarkdownToHtml()` (`src/components/Editor/reading/markdownRenderer.ts`), which has custom inline rules for `[[wikilinks]]` and `![[embeds]]` but none for `{{ ... }}`. The evaluator (`evaluateProgram()` in `src/lib/templateProgram.ts`) is never called on the Reading path.

## Proposal

Run `{{ ... }}` through `evaluateProgram()` as a **pre-markdown pass** inside `renderMarkdownToHtml()`, before markdown-it sees the string. The evaluator already returns plain text (incl. GFM tables), so letting markdown-it render the result afterwards means tables, wiki-links inside template output, lists, etc. all just work without duplicating any widget logic from CM6.

### Sketch

In `markdownRenderer.ts`:

```ts
import { evaluateProgram } from "../../../lib/templateProgram";

const EXPR_RE = /\{\{([^{}]+?)\}\}/g;

function expandTemplates(md: string, title: string): string {
  const vault = currentVaultRoot();
  const now = new Date();
  const scope = { vault, date: formatDate(now), time: formatTime(now), title };
  return md.replace(EXPR_RE, (full, body) => {
    try {
      const out = evaluateProgram(body, scope);
      return out === "" ? full : out;  // Leave source on empty/error — same UX as CM6
    } catch {
      return full;
    }
  });
}

export function renderMarkdownToHtml(markdown: string, noteTitle = ""): string {
  const stripped = stripFrontmatter(markdown);
  const expanded = expandTemplates(stripped, noteTitle);
  // … existing pipeline (md.render + task-checkbox pass + DOMPurify) …
}
```

Then in `ReadingView.svelte`, pass the note's basename so `{{title}}` resolves:

```ts
html = renderMarkdownToHtml(content, titleFromPath(tab.filePath));
```

### Scope object

Same keys the CM6 plugin already supplies (`templateLivePreview.ts:202–208`): `vault`, `date`, `time`, `title`. Move `formatDate` / `formatTime` / `activeTitle`-equivalent helpers out of `templateLivePreview.ts` into a shared `src/lib/templateScope.ts` so both callers stay in sync.

### Re-render on vault changes

`ReadingView.svelte:52–58` currently only re-loads on `tab.id` change. For templates to update when underlying notes change, subscribe to the same stores the CM6 plugin watches (`templateLivePreview.ts:247–253`): `vaultStore`, `tagsStore`, `bookmarksStore`, `noteContentCacheVersion` (#319). On any tick, call `load()`.

## Out of scope

- Custom widget (table) rendering — GFM tables produced by the evaluator are already rendered natively by markdown-it, so no table-specific HTML logic is needed in Reading Mode.
- Edit-in-place of template output (reader is read-only by design).
- Exposing additional scope fields beyond `{ vault, date, time, title }` — separate follow-up.

## Acceptance criteria

- [ ] `{{date}}`, `{{time}}`, `{{title}}` render in Reading Mode identically to their live-preview rendering.
- [ ] `{{vault.folders.where(...).notes…}}` queries evaluate against the vault, including multi-segment (`; `) programs and GFM-table output.
- [ ] Wiki-links produced inside a template (e.g. `"[[" + n.name + "]]"`) still become clickable anchors — markdown-it's existing `wiki_link` rule handles this automatically once the template has expanded.
- [ ] Evaluation errors leave the original `{{ ... }}` source visible (no empty output, no crashes), matching the CM6 UX.
- [ ] Reading Mode re-renders when vault contents change (new note, rename, delete, non-active-note edit) without a tab switch.
- [ ] No regression in existing Reading Mode rendering (wiki-links, embeds, headings, task lists, frontmatter strip, sanitization).

## Related

- #283 — Template expressions (engine + CM6 live-preview)
- #319 — Sync `readNoteContent` cache (ensures `.content.contains(...)` works for non-active notes; Reading Mode should subscribe to `noteContentCacheVersion` for the same reason)
- #63 — Reading Mode base implementation

herox215:
Closes #321.

## Summary

- New `src/lib/templateScope.ts` owns the single source of truth for the scope exposed to `{{ ... }}` expressions (`vault`, `date`, `time`, `title`); both CM6 live-preview and Reading Mode pull from it.
- `renderMarkdownToHtml()` gains a pre-markdown-it expansion pass that runs `evaluateProgram` over every `{{ ... }}` match. GFM tables, wiki-links, and lists produced by templates flow through the normal markdown pipeline — no Reading-Mode-specific widget code.
- Failure / empty-output semantics mirror the CM6 UX: the original source stays visible so a broken expression remains debuggable when flipping between edit and read modes.
- `ReadingView` subscribes to `vaultStore` / `tagsStore` / `bookmarksStore` / `noteContentCacheVersion` so template output re-renders when vault contents change on disk (rename, delete, non-active note edits), not only on tab switch.
- `renderMarkdownToHtml(markdown, noteTitle?)` — new optional `noteTitle` feeds `{{title}}`; defaults to the active editor tab's basename (existing CM6 behaviour) when omitted.

## Test plan

- [x] `pnpm test` — 1118 passing (7 new tests covering `{{title}}`, `{{date}}`, multi-segment, error-keeps-source, wiki-link inside template, no-regression on plain markdown, empty-output-keeps-source + scope-builder unit tests)
- [x] `pnpm build` — clean
- [ ] Manual UAT (by author, after review fixes): open a note with `{{vault.folders.where(...).notes.select(n => "[[" + n.name + "]]").join("\n")}}`, flip to Reading Mode, confirm rendered anchors navigate; rename a note in another folder, confirm the reading pane re-renders without a tab switch.

### DIFF
--- a/src/components/Editor/ReadingView.svelte
+++ b/src/components/Editor/ReadingView.svelte
@@ -18,6 +18,11 @@
   import { toastStore } from "../../store/toastStore";
   import { tabStore } from "../../store/tabStore";
   import type { Tab } from "../../store/tabStore";
+  import { vaultStore } from "../../store/vaultStore";
+  import { tagsStore } from "../../store/tagsStore";
+  import { bookmarksStore } from "../../store/bookmarksStore";
+  import { noteContentCacheVersion } from "../../lib/noteContentCache";
+  import { titleFromPath } from "../../lib/templateScope";
 
   interface Props {
     tab: Tab;
@@ -38,7 +43,7 @@
     if (!tab) return;
     try {
       const content = await readFile(tab.filePath);
-      html = renderMarkdownToHtml(content);
+      html = renderMarkdownToHtml(content, titleFromPath(tab.filePath));
       loadError = null;
     } catch (err) {
       loadError = "Datei konnte nicht gelesen werden.";
@@ -55,6 +60,33 @@
       loadedForId = id;
       void load();
     }
+  });
+
+  // #322 — Reading Mode must re-render when the data a `{{ ... }}` template
+  // expression reads over changes on disk, otherwise a template like
+  // `{{vault.notes.where(n => n.content.contains("X"))}}` would stay frozen
+  // with whatever the vault looked like when the tab opened. Subscribe to the
+  // same stores the CM6 live-preview plugin watches (vaultStore / tagsStore /
+  // bookmarksStore / noteContentCacheVersion) and re-read the file on any
+  // tick. The svelte subscribe contract invokes the callback synchronously
+  // with the current value, so a `ready` latch suppresses the initial burst
+  // that would otherwise double-load on mount.
+  onMount(() => {
+    let ready = false;
+    const trigger = (): void => {
+      if (!ready) return;
+      if (loadedForId !== null) void load();
+    };
+    const unsubs: Array<() => void> = [
+      vaultStore.subscribe(trigger),
+      tagsStore.subscribe(trigger),
+      bookmarksStore.subscribe(trigger),
+      noteContentCacheVersion.subscribe(trigger),
+    ];
+    ready = true;
+    return () => {
+      for (const u of unsubs) u();
+    };
   });
 
   // Save the current scroll position when this view deactivates (tab switch
--- a/src/components/Editor/ReadingView.svelte
+++ b/src/components/Editor/ReadingView.svelte
@@ -62,7 +62,7 @@
     }
   });
 
-  // #322 — Reading Mode must re-render when the data a `{{ ... }}` template
+  // #321 — Reading Mode must re-render when the data a `{{ ... }}` template
   // expression reads over changes on disk, otherwise a template like
   // `{{vault.notes.where(n => n.content.contains("X"))}}` would stay frozen
   // with whatever the vault looked like when the tab opened. Subscribe to the
--- a/src/components/Editor/ReadingView.svelte
+++ b/src/components/Editor/ReadingView.svelte
@@ -12,7 +12,7 @@
    * position independently.
    */
 
-  import { onMount, onDestroy } from "svelte";
+  import { onMount } from "svelte";
   import { readFile } from "../../ipc/commands";
   import { renderMarkdownToHtml } from "./reading/markdownRenderer";
   import { toastStore } from "../../store/toastStore";
@@ -39,15 +39,37 @@
   // pane re-loads even when the component is reused.
   let loadedForId: string | null = null;
 
+  // Generation token guards against stale-write races: every call to `load()`
+  // takes the current value of `gen`, and only writes `html` / `loadError` if
+  // its token is still current when the async `readFile` resolves. Without
+  // this, a slow read for tab A that's still in flight when the user switches
+  // to tab B can overwrite tab B's rendered HTML with tab A's content.
+  let gen = 0;
+  // Pending-reload latch: when multiple store subscriptions fire in the same
+  // frame (e.g. rename bumps vaultStore + noteContentCacheVersion + tagsStore
+  // at once), coalesce them into a single microtask-scheduled reload.
+  let reloadPending = false;
+
   async function load(): Promise<void> {
     if (!tab) return;
+    const token = ++gen;
+    const path = tab.filePath;
     try {
-      const content = await readFile(tab.filePath);
-      html = renderMarkdownToHtml(content, titleFromPath(tab.filePath));
+      const content = await readFile(path);
+      if (token !== gen) return;
+      html = renderMarkdownToHtml(content, titleFromPath(path));
       loadError = null;
     } catch (err) {
-      loadError = "Datei konnte nicht gelesen werden.";
-      toastStore.push({ variant: "error", message: loadError });
+      if (token !== gen) return;
+      const message = "Datei konnte nicht gelesen werden.";
+      // Suppress repeated toasts while the error condition persists — a
+      // bulk rename or mid-rename watcher burst emits many store ticks per
+      // frame, and raising a toast on every one of them floods the UI.
+      const shouldToast = loadError !== message;
+      loadError = message;
+      if (shouldToast) {
+        toastStore.push({ variant: "error", message: message });
+      }
     }
   }
 
@@ -70,12 +92,21 @@
   // bookmarksStore / noteContentCacheVersion) and re-read the file on any
   // tick. The svelte subscribe contract invokes the callback synchronously
   // with the current value, so a `ready` latch suppresses the initial burst
-  // that would otherwise double-load on mount.
+  // that would otherwise double-load on mount. `queueMicrotask` coalesces
+  // bursts that cross multiple stores (rename / delete / cache version bump
+  // commonly fire all four in the same frame) into one `readFile` + render
+  // cycle instead of four.
   onMount(() => {
     let ready = false;
     const trigger = (): void => {
       if (!ready) return;
-      if (loadedForId !== null) void load();
+      if (loadedForId === null) return;
+      if (reloadPending) return;
+      reloadPending = true;
+      queueMicrotask(() => {
+        reloadPending = false;
+        void load();
+      });
     };
     const unsubs: Array<() => void> = [
       vaultStore.subscribe(trigger),
@@ -84,6 +115,9 @@
       noteContentCacheVersion.subscribe(trigger),
     ];
     ready = true;
+    // Teardown lives next to the subscriptions rather than in a separate
+    // `onDestroy` so a future reader sees the full lifecycle in one block.
+    // The scroll-persist cleanup below uses the same pattern.
     return () => {
       for (const u of unsubs) u();
     };
@@ -109,12 +143,26 @@
     wasActive = nowActive;
   });
 
-  // Persist scroll on unmount too — covers closing the tab while in Reading
-  // Mode, which otherwise would drop the scroll position.
-  onDestroy(() => {
+  // Restore initial scroll after the first render when the tab opens directly
+  // in Reading Mode (isActive was already true when the component mounted, so
+  // the wasActive → active transition above wouldn't fire). Also persists
+  // scroll on unmount — covers closing the tab while in Reading Mode, which
+  // otherwise would drop the scroll position. Teardown is returned as the
+  // onMount cleanup to keep the whole lifecycle in one block instead of
+  // splitting between onMount + a separate onDestroy.
+  onMount(() => {
     if (isActive && containerEl) {
-      tabStore.updateReadingScrollPos(tab.id, containerEl.scrollTop);
+      const pos = tab.readingScrollPos ?? 0;
+      requestAnimationFrame(() => {
+        if (containerEl) containerEl.scrollTop = pos;
+      });
     }
+    wasActive = isActive;
+    return () => {
+      if (isActive && containerEl) {
+        tabStore.updateReadingScrollPos(tab.id, containerEl.scrollTop);
+      }
+    };
   });
 
   function handleClick(event: MouseEvent): void {
@@ -158,19 +206,6 @@
     const targetTop = targetEl.getBoundingClientRect().top;
     containerEl.scrollTop += targetTop - containerTop - 8;
   }
-
-  // Restore initial scroll after the first render when the tab opens directly
-  // in Reading Mode (isActive was already true when the component mounted, so
-  // the wasActive → active transition above wouldn't fire).
-  onMount(() => {
-    if (isActive && containerEl) {
-      const pos = tab.readingScrollPos ?? 0;
-      requestAnimationFrame(() => {
-        if (containerEl) containerEl.scrollTop = pos;
-      });
-    }
-    wasActive = isActive;
-  });
 </script>
 
 <!-- svelte-ignore a11y_click_events_have_key_events -->
--- a/src/components/Editor/__tests__/markdownRenderer.test.ts
+++ b/src/components/Editor/__tests__/markdownRenderer.test.ts
@@ -188,15 +188,75 @@ describe("renderMarkdownToHtml (#63)", () => {
       expect(html).toContain("<p>World</p>");
     });
 
-    it("leaves empty-output templates as source to keep them debuggable", () => {
-      // `""` evaluates to the empty string; per the semantics documented in
-      // `expandTemplates`, empty output leaves the source visible so an
-      // accidentally-empty template is not silently swallowed in Reading
-      // Mode. markdown-it's typographer converts the straight quotes to
-      // curly quotes during rendering — assert on the `{{`/`}}` markers
-      // rather than the inner chars.
-      const html = renderMarkdownToHtml('{{ "" }}', "n");
-      expect(html).toMatch(/\{\{.*\}\}/);
+    it("collapses deliberately-empty template output to nothing", () => {
+      // `""` evaluates successfully to the empty string. Reading Mode must
+      // emit empty rather than the literal source, so that a conditional
+      // like `{{ flag ? "x" : "" }}` can cleanly collapse — the reader
+      // should never see the raw template syntax for a successful eval.
+      const html = renderMarkdownToHtml('before{{ "" }}after', "n");
+      expect(html).toContain("beforeafter");
+      expect(html).not.toContain("{{");
+    });
+
+    it("keeps the source visible only when the evaluator throws", () => {
+      // Unknown identifier throws → source stays. Empty output from a
+      // successful eval collapses to empty. Asserting the two paths are
+      // treated differently pins the semantics in place so they can't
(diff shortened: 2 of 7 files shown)


src/components/Editor/reading/markdownRenderer.ts
```diff
@@ -201,12 +203,47 @@ function renderTaskListCheckboxes(html: string): string {
...
+// GFM tables, joined wiki-link fragments, …), which then flows through the
+// normal markdown pipeline — tables, lists, and wiki-links inside template
+// output render without any Reading-Mode-specific widget logic.
+//
+// Semantics mirror the CM6 live-preview plugin (`templateLivePreview.ts`):
+// on parse/evaluation failure and on empty output, the original `{{ ... }}`
+// source stays visible instead of collapsing to nothing. That keeps broken
+// expressions debuggable when flipping between edit and read views.
+
+const EXPR_RE = /\{\{([^{}]+?)\}\}/g;
```
herox215: [blocker] \`EXPR_RE = /\{\{([^{}]+?)\}\}/g\` is applied to the whole document before markdown-it sees it, so templates inside fenced code blocks, inline code, and indented code are **evaluated**. CM6 was shielded by the viewport slice + CodeMirror 's token-aware live-preview flow; Reading Mode is not. A note with literal \`\`\`markdown\\n{{date}}\\n\`\`\` (e.g. your own docs demonstrating the feature) will render "Today's date" inside the code block. Either (a) tokenise code fences / inline code before the replace, or (b) move the expansion into a markdown-it plugin that skips \`code_inline\`, \`fence\`, and \`code_block\` tokens.

herox215: [major] The regex also breaks on perfectly legal expression bodies that contain a brace inside a string literal: \`{{ "a{b" }}\` and \`{{ foo.bar(">>") }}\` do not match at all (the \`[^{}]\` class rejects the inner brace), so they render as literal \`{{ ... }}\` in Reading Mode while CM6's live-preview would evaluate them (its regex is identical but its input is the viewport text, not user-authored string literals crossing braces). Document the grammar restriction, or rewrite the outer matcher as a small state machine that respects string literals.

herox215: [major] Cross-line matching parity: this regex has no \`s\` flag but \`[^{}]\` matches newlines, so a multi-line \`{{\\n vault.folders…\\n .where(...)\\n}}\` **does** evaluate in Reading Mode. The CM6 plugin matches against a single viewport slice and authors typically keep \`{{ ... }}\` on one line; in Reading Mode, an author could paste a long multi-line template and get a different behaviour than they see while editing if the CM6 match straddles a viewport boundary. Worth an explicit test asserting cross-line equivalence (or a conscious single-line restriction).

herox215 (commented):
High-level verdict: **request-changes-style / comment-only** — the plumbing is right but there are correctness and UX bugs that will surface on real notes.

Top 3 concerns:
1. **`EXPR_RE` expands templates inside fenced code blocks / inline code** in Reading Mode (CM6 was shielded by CodeMirror's token model and its viewport match; markdown-it gets the expanded text, so your carefully-authored ```{{date}}``` examples in docs become today's date).
2. **`ReadingView.load()` races across tab/store ticks** — no generation token, so a store tick during a tab switch can write an old tab's HTML on top of the new tab's content. On a transient `readFile` error, every store tick also pushes a fresh toast.
3. **Typographer is on** (`typographer: true`), so any template whose output contains `--`, `...`, `(c)`, straight quotes, etc. is silently rewritten. Combined with the weakened `toMatch(/\{\{.*\}\}/)` assertion, the empty-output test no longer detects regressions.

Inline findings below.

src/components/Editor/ReadingView.svelte
```diff
@@ -38,7 +43,7 @@
     if (!tab) return;
     try {
       const content = await readFile(tab.filePath);
-      html = renderMarkdownToHtml(content);
+      html = renderMarkdownToHtml(content, titleFromPath(tab.filePath));
```
herox215: [blocker] Stale-write race in \`load()\`. \`tab.filePath\` is captured at the top of the function, but between \`await readFile(...)\` and \`html = renderMarkdownToHtml(...)\`, the \`tab\` prop can mutate (parent swaps active tab) or a concurrent \`load()\` call (triggered by a store tick) can land first. The \`loadedForId\` latch only guards the \`$effect\`-driven kickoff, not in-flight reads. Repro: open tab A on a large note, swap to tab B, observe tab B briefly show tab A content. Fix with a generation token:\n\`\`\`ts\nlet gen = 0;\nasync function load() {\n  const my = ++gen;\n  const path = tab.filePath;\n  const content = await readFile(path);\n  if (my !== gen) return;\n  html = renderMarkdownToHtml(content, titleFromPath(path));\n}\n\`\`\`

src/components/Editor/ReadingView.svelte
```diff
@@ -38,7 +43,7 @@
...
       loadError = null;
     } catch (err) {
       loadError = "Datei konnte nicht gelesen werden.";
```
herox215: [major] Toast storm on transient errors. If \`readFile\` fails for a transient reason (mid-rename, watcher invalidation, AV scanner lock on Windows), every subsequent \`vaultStore\`/\`tagsStore\`/\`bookmarksStore\`/\`noteContentCacheVersion\` tick re-enters \`load()\` and pushes another error toast. A bulk rename can emit many ticks per frame. Suggest suppressing the toast when \`loadError\` is already set for the same path, or debouncing the store-driven reload.

src/components/Editor/ReadingView.svelte
```diff
@@ -57,6 +62,33 @@
...
+      if (!ready) return;
+      if (loadedForId !== null) void load();
+    };
+    const unsubs: Array<() => void> = [
+      vaultStore.subscribe(trigger),
+      tagsStore.subscribe(trigger),
+      bookmarksStore.subscribe(trigger),
+      noteContentCacheVersion.subscribe(trigger),
+    ];
+    ready = true;
```
herox215, line 117: [major] No coalescing across the 4 store subscriptions. A single action (e.g. rename) can bump \`vaultStore\`, \`noteContentCacheVersion\`, and possibly \`tagsStore\` in the same frame — that triggers 3 back-to-back \`readFile\` + \`renderMarkdownToHtml\` cycles on the same tab for identical disk content. For a large note with many \`{{ ... }}\` expressions iterating the vault, this is visibly expensive. Wrap \`trigger\` in a \`queueMicrotask\`/\`requestAnimationFrame\` dedupe.

src/components/Editor/ReadingView.svelte
```diff
@@ -57,6 +62,33 @@
...
+    return () => {
+      for (const u of unsubs) u();
+    };
+  });
```
herox215, line 124: [minor] Mixing \`onMount\`-return cleanup with a separate \`onDestroy\` in the same component is a readability trap — a later reader searching for teardown will look at \`onDestroy\` first and miss this. Either move the scroll-persistence \`onDestroy\` into an \`onMount\`-return too, or keep the store-subscription cleanup in an \`onDestroy\` for consistency.

src/lib/templateScope.ts
```diff
@@ -0,0 +1,62 @@
...
+ * `title` may be passed explicitly (Reading Mode knows the tab's file path up
+ * front); when omitted, it falls back to the active editor tab's basename —
+ * the CM6 plugin's behaviour.
+ */
+export function buildTemplateScope(options?: {
+  now?: Date;
+  title?: string;
+}): EvalScope {
+  const now = options?.now ?? new Date();
+  const title = options?.title ?? activeTitle();
```
herox215, line 74: [major] Docstring-vs-behaviour mismatch. \`renderMarkdownToHtml\` declares \`noteTitle = ""\` (explicit empty-string default) and this uses \`options?.title ?? activeTitle()\`. Nullish-coalescing treats \`""\` as a **defined** value, so passing the default empty string does NOT fall back to the active editor tab's basename — contrary to the JSDoc ("When omitted, templates fall back to the active editor tab's title"). Either change the default to \`undefined\` (and propagate through \`expandTemplates\`) or fix the doc. Current state is a silent trap.

herox215 (commented):
Continuation of review: race, UX, coverage findings.

src/components/Editor/__tests__/markdownRenderer.test.ts
```diff
@@ -140,4 +140,63 @@ describe("renderMarkdownToHtml (#63)", () => {
...
+
+    it("leaves empty-output templates as source to keep them debuggable", () => {
+      // `""` evaluates to the empty string; per the semantics documented in
+      // `expandTemplates`, empty output leaves the source visible so an
+      // accidentally-empty template is not silently swallowed in Reading
+      // Mode. markdown-it's typographer converts the straight quotes to
+      // curly quotes during rendering — assert on the `{{`/`}}` markers
+      // rather than the inner chars.
+      const html = renderMarkdownToHtml('{{ "" }}', "n");
+      expect(html).toMatch(/\{\{.*\}\}/);
```
herox215: [major] \`expect(html).toMatch(/\\{\\{.*\\}\\}/)\` passes on almost any wrong output. Confirmed: \`<p>{{ wrong }}</p>\`, \`<p>{{ "X" }}</p>\`, and \`<p>{{ “” }}</p>\` all satisfy this regex. If a future refactor emits the WRONG source text on empty output this test still passes. Options: (a) disable \`typographer\` on the \`md\` instance (also fixes the em-dash / ellipsis / curly-quote corruption of real template output); (b) assert the exact post-typographer string \`{{ “” }}\`; (c) swap the test input to a literal that does not get typographer-mangled, e.g. \`{{ "x".startsWith("y") ? "a" : "" }}\` and assert the full \`{{ ... }}\` body survives verbatim.

src/components/Editor/__tests__/markdownRenderer.test.ts
```diff
@@ -140,4 +140,63 @@ describe("renderMarkdownToHtml (#63)", () => {
...
+    });
+
+    it("preserves wiki-links produced inside a template expression as anchors", () => {
+      // `"[[foo]]"` as a string literal → same `[[foo]]` text after
+      // evaluation → markdown-it's wiki-link rule then turns it into an
+      // anchor with data-wiki-target.
+      setResolvedLinks(new Map([["foo", "notes/foo.md"]]));
+      const html = renderMarkdownToHtml('{{ "[[foo]]" }}', "n");
+      expect(html).toMatch(/<a[^>]*data-wiki-target="foo"[^>]*>/);
+    });
```
herox215, line 183: [major] Acceptance-criterion coverage gap. AC5 on #321 says "Reading Mode re-renders when vault contents change (new note, rename, delete, non-active-note edit) without a tab switch". The new subscription block in \`ReadingView.svelte\` wires the triggers, but no test drives a store mutation and asserts \`load()\` re-runs. Without that, a future refactor can delete the \`vaultStore\` subscription and CI still passes. Add a component test that mounts ReadingView, mutates \`vaultStore\` / \`noteContentCacheVersion\`, and verifies \`readFile\` is called again.

src/components/Editor/reading/markdownRenderer.ts
```diff
@@ -201,12 +203,47 @@ function renderTaskListCheckboxes(html: string): string {
...
+function expandTemplates(markdown: string, title: string): string {
+  // Build the scope once per render, even if there are zero `{{ ... }}`
+  // matches — `replace` lazily calls the callback so the cost is tied to
+  // matches, not to the scope builder itself (which touches svelte stores).
+  let scope: ReturnType<typeof buildTemplateScope> | null = null;
+  return markdown.replace(EXPR_RE, (full, body: string) => {
+    try {
+      if (scope === null) scope = buildTemplateScope({ title });
+      const out = evaluateProgram(body, scope);
+      return out === "" ? full : out;
```
herox215: [question] \`return out === "" ? full : out\` treats any empty string as "error/empty = keep source visible". But what about a deliberately empty template like \`{{ foo ? "" : "bar" }}\`? In CM6 the matching logic drops the decoration and the source stays visible (user types next to their cursor) — in Reading Mode that source then renders as literal \`{{ ... }}\` to the reader, which is **worse** UX than collapsing to empty. Consider distinguishing "evaluator error" (keep source) from "evaluator succeeded with empty string" (emit empty) — \`try\`/\`catch\` already knows which case applies.

herox215:
[major] `typographer: true` on the `md` instance (markdownRenderer.ts:37, unchanged) silently rewrites evaluated template output. A template that returns `"--"`, `"..."`, `"(c)"`, `"(tm)"`, or straight double-quotes is post-processed into `–`, `…`, `©`, `™`, and curly quotes before reaching the user. This is new user-visible corruption introduced by #321: `{{ foo.bar.join(" -- ") }}` now renders different glyphs in Reading Mode than the CM6 widget emits (CM6 produces text nodes directly, never going through markdown-it's core `replacements`/`smartquotes` ruler).

Either disable `typographer` on this `md`, wrap template output in a sentinel that a custom markdown-it rule skips during the core `replacements`/`smartquotes` pass, or document the divergence as a known issue. The new empty-output test (`toMatch(/\{\{.*\}\}/)`) was weakened specifically to work around this — suggesting the corruption is non-trivial enough to matter.

herox215:
[minor] Code-quality / maintainability:

1. `EXPR_RE` is now declared in two modules (`markdownRenderer.ts:219` and `templateLivePreview.ts:152`) with the same body. `templateScope.ts` already exists as the shared seam — move `EXPR_RE` there too so a future grammar change only touches one file.

2. `renderMarkdownToHtml` has **two** JSDoc blocks stacked against the same declaration (reading/markdownRenderer.ts:239-247 and 248-253) — the `ALLOWED_URI_REGEXP` comment sits between the new `noteTitle` doc and the `export function`. TypeScript tooling picks up only the block immediately above the declaration, so the `noteTitle` description is effectively hidden from editor hovers. Consolidate into one block above the const, or move the URI-regexp comment above the const.

3. The `ready` latch in `ReadingView.svelte:74-86` re-implements the exact pattern used in `templateLivePreview.ts:214-228`. Consider extracting a small helper (`subscribeMany(stores, cb)`) that swallows the initial burst and returns a cleanup.

herox215:
Review findings addressed in 688738a. Summary of fixes:

**Blockers**
- Code-fence / inline-code protection: new `maskCodeRegions` pre-pass uses Private-Use-Area placeholders so `{{ ... }}` inside fenced or backtick-quoted regions renders verbatim. Covered by the new `does not evaluate templates inside fenced code blocks` / `inline backtick spans` tests.
- Stale-write race: `load()` now holds a generation token; the async `readFile` result is dropped if the token no longer matches.

**Major**
- `typographer: false` on the `md` instance so evaluated output isn't silently rewritten into em-dash / curly quotes. Pinned by `does not silently rewrite evaluated text via markdown-it typographer`.
- Error vs. empty-output split: expansion walks `splitSegments` + `evaluate` directly so a thrown segment keeps the source while all-successful-empty output collapses. Pinned by `keeps the source visible only when the evaluator throws` + `collapses deliberately-empty template output to nothing`.
- `noteTitle?` is now truly optional (`undefined` falls back to the active editor tab's basename; explicit `""` is honoured as deliberate).
- Toast storm: repeated read errors for the same message don't re-push a toast.
- Burst coalescing: multi-store same-tick updates coalesce via `queueMicrotask` into a single reload. Pinned by `coalesces synchronous same-tick vaultStore ticks into a single reload`.
- AC5 coverage: new `ReadingViewReload.test.ts` mounts the component and asserts re-reads on `vaultStore` and `noteContentCacheVersion` ticks.

**Minor**
- `TEMPLATE_EXPR_RE` moved to `templateScope.ts`; both live-preview + Reading Mode import it from there.
- `renderMarkdownToHtml` now has a single JSDoc block above the declaration; the `ALLOWED_URI_REGEXP` comment moved above its own const.
- Scroll-persist `onDestroy` folded into the scroll-restore `onMount`'s cleanup so ReadingView has one lifecycle block.

**Documented CM6 parity gap (not changed)**
- Regex `[^{}]` still rejects `{`/`}` inside string literals — same behaviour as `templateLivePreview.ts`. New `does not expand ... containing braces` test pins this as a known parity.
- Cross-line templates still evaluate (regex admits `\n`). New test asserts this.

All 1127 tests pass, typecheck clean, build clean. Ready for a second look.

PR #324, opened 2026-04-21
fix(links): resolve `[[X]]` from template output after async map refresh (#309)

### LINKED ISSUE #309: Wiki-link click on template-rendered `[[New Note]]` still falls back to create-at-root
## Reproduction

1. Open a note containing a template query like:

   ```
   {{ "|Pages|x|\n|-|-|\n"; vault.folders.where(f => f.name == "test").first().notes.select(n => "|[[" + n.name + "]]| |").join("\n") }}
   ```

2. Right-click the "test" folder in the sidebar → **New note**. A file like `test/Untitled.md` is created.
3. Return to the template-query note. The rendered table now includes `[[Untitled]]` (good — #307 fixed the staleness).
4. Click the wiki-link `[[Untitled]]`.

**Expected:** the editor navigates to `test/Untitled.md`.
**Actual:** the click falls through to the create-at-root handler — the user lands in an empty note at the vault root, not in the real file at `test/Untitled.md`.

## Context

Partially addressed in PR #308 (#307):
- Backend `create_file_impl` now calls `sync_file_index_create` so `FileIndex` contains the new file.
- Frontend `VaultLayout` calls `resolvedLinksStore.requestReload()` on every non-modify `vault://file_changed` event.

Despite this, the behaviour still reproduces on the user's machine. Possible suspects to investigate:

- Is `setResolvedLinks(...)` actually being called after the reload request? The `resolvedLinksStore.subscribe` in `EditorPane.svelte:364` might only fire for the active editor — the template-query note the user is looking at is a different tab/view and might not re-render its wiki-link decorations.
- Is `refreshWikiLinks()` being invoked on all open EditorViews, or only the focused one? (See `wikiLink.ts` refresh helper.)
- The template live-preview widget renders a `<span>` for the wiki-link and reads `resolvedLinks.get(stem)` at build time in `templateLivePreview.ts:appendValueWithWikiLinks`. The template plugin re-runs `buildDecorations` on store changes, but between `requestReload()` → `getResolvedLinks()` → `setResolvedLinks()` there is an async hop. If `buildDecorations` runs before the map is updated, the rendered `data-wiki-resolved="false"` stays cached on the decoration and EditorPane's click handler takes the "create" branch.
- Even after the map updates, the template decoration might only rebuild on a later vault-store tick, not on `setResolvedLinks`. Check whether `refreshWikiLinks()` or an equivalent effect fires into template-live-preview.

## Acceptance

Clicking `[[X]]` inside a template-rendered cell opens the file at its true relative path with zero app restart, regardless of whether `X` was created via the sidebar, rename/move, command palette, or any other flow.

## Related

- #307 — fileList sync (merged)
- #305 — live table rendering (merged)
- #277 — rename link resolution (reference for prior art on the same class of bug)

herox215:
Closes #309.

## Summary
- Wiki-links rendered inside a `{{ ... }}` template expression no longer fall through to the create-at-root branch when their target file was just created via the sidebar / rename / move.
- Splits `resolvedLinksStore` into `requestToken` (stale → refetch) + `readyToken` (fresh map available → rebuild decorations) so decoration layers react to the post-`setResolvedLinks` edge, not the pre-refetch edge.
- CM6 template widgets (`RenderedValueWidget`, `RenderedTableWidget`) now include a wiki-link resolution snapshot in `eq()` so identical rendered text with different resolution forces a DOM rebuild.
- `EditorPane.handleWikiLinkClick` re-checks resolution synchronously before taking the create branch — defense-in-depth so a stale-`false` decoration can't spawn a duplicate note even in a pathological race.

## Files
- `src/store/resolvedLinksStore.ts` — two-token API (`requestToken` / `readyToken`), new `markReady()`.
- `src/components/Editor/EditorPane.svelte` — watch `requestToken`, call `markReady()` after `setResolvedLinks()`, live-lookup in click handler.
- `src/components/Editor/templateLivePreview.ts` — subscribe to `readyToken`, include wiki-link resolution snapshot in widget `eq()`.
- `src/components/Editor/ReadingView.svelte` — subscribe to `readyToken` through the existing coalesced reload path.

## Test plan
- [x] `pnpm test` — 1129/1129 pass (includes two new regression tests).
- [x] `pnpm exec svelte-check` — 0 errors (only pre-existing CSS warnings).
- [ ] Manual: open a template `{{ vault.folders.where(f => f.name == "test").first().notes.select(n => "[[" + n.name + "]]").join("\n") }}`, create `test/Untitled.md` via sidebar → click `[[Untitled]]` → editor opens the real file at `test/Untitled.md` (not an empty note at vault root).
- [ ] Manual: same flow in Reading Mode — `[[Untitled]]` flips from grey-dashed to accent-coloured without a tab reload.

### DIFF
--- a/src/components/Editor/EditorPane.svelte
+++ b/src/components/Editor/EditorPane.svelte
@@ -289,6 +289,12 @@
       setResolvedLinks(new Map());
       setResolvedAttachments(new Map());
     }
+    // #309 — bump readyToken AFTER the map has been swapped in (success or
+    // soft-fail) so decoration layers that outlive EditorPane's refresh — CM6
+    // template live-preview, Reading Mode — can rebuild against the fresh
+    // map. Without this, their subscriptions would either never fire or fire
+    // on requestToken and rebuild against the still-stale map.
+    resolvedLinksStore.markReady();
   }
 
   /**
@@ -303,9 +309,19 @@
     u();
     if (!vault) return;
 
-    if (detail.resolved) {
+    // #309 — the decoration's `data-wiki-resolved` attribute reflects the
+    // resolved-links map at decoration-build time. When a file is created or
+    // moved between render and click (e.g. template re-renders on vaultStore
+    // tick before the async resolvedLinks refetch lands), the attribute can
+    // be stale-`false` while the live map now resolves the target. Re-check
+    // synchronously here so a stale decoration never routes into the
+    // create-at-root fallback.
+    const liveRelPath = resolveTarget(detail.target);
+    const effectivelyResolved = detail.resolved || liveRelPath !== null;
+
+    if (effectivelyResolved) {
       // LINK-03: synchronous lookup — zero IPC at click time
-      const relPath = resolveTarget(detail.target);
+      const relPath = liveRelPath ?? resolveTarget(detail.target);
       if (!relPath) {
         // Map out of sync (rare: file deleted between decoration and click)
         void reloadResolvedLinks();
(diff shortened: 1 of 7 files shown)


herox215:
@minerva review

herox215:
@minerva review

herox215:
@minerva review

PR #326, opened 2026-04-21
fix(template): strip `{{ ... }}` bodies from n.content searches (#325)

### LINKED ISSUE #325: Template self-reference: n.content includes the template body itself
## Problem

A template expression like

\`\`\`
{{vault.notes.where(n => n.content.contains(\"todo\")).select(f => \"[[\" + f.name + \"]]\").join(\"\\n\")}}
\`\`\`

currently matches the note the template lives in, because the raw text of that note includes the template body itself — the literal \`\"todo\"\` inside the \`.contains(...)\` call is part of \`n.content\` when \`n\` is the self-note.

Same shape reproduces for any string arg to \`.contains(...)\`, \`.startsWith(...)\`, \`.indexOf(...)\`, etc. Two concrete user reports:

- \`{{vault.notes.where(n => n.content.contains(\"Maestro\")).select(f => f.name)}}\` — matches self because \"Maestro\" appears inside the template body.
- \`{{(\"|test|test|\\n|-|-|\\n\"); vault.notes.where(n => n.content.contains(\"todo\")).select(f => \"|[[\" + f.name + \"]]|-|\").join(\"\\n\")}}\` — same.

## Root cause

\`n.content\` is lazy-read through \`stores.readNoteContent(relPath)\` → \`noteContentCache\` → \`readFile\` IPC. That returns the note's raw bytes, which include the \`{{ ... }}\` regions. Template bodies are code, not prose, but \`n.content\` treats them as prose.

## Proposed fix

Strip every \`{{ ... }}\` region from the string returned by \`n.content\` for **all** notes (not just the self-note) before handing it to the user predicate. Implementation chokepoint: \`getContent()\` in \`src/lib/vaultApi.ts:158-164\`. Reuse the shared \`TEMPLATE_EXPR_RE\` from \`src/lib/templateScope.ts\` so the live-preview plugin and this stripper stay in lockstep.

This also covers the symmetric case where note A's template references note B's content — note B's template body shouldn't contribute to searches in note A either.

## Acceptance

- Regression test in \`src/lib/__tests__/vaultApi.test.ts\`: seed two notes, one with \`\"todo\"\` in prose, one where \`\"todo\"\` appears only inside a template body. \`vault.notes.where(n => n.content.includes(\"todo\"))\` returns only the first.
- All existing 1134 tests still pass.

## Out of scope

- \`this\` / \`self\` accessor for explicit self-reference (tracked separately if we want it later).
- Stripping template bodies from full-text search. Different surface, different tradeoffs.

herox215:
Closes #325.

## Summary
- `{{vault.notes.where(n => n.content.contains(\"todo\"))}}` no longer matches the note the template lives in. The literal string inside the expression body would previously leak into `n.content` and make the self-note self-match for every string-predicate shape.
- Chokepoint fix: strip `{{ ... }}` regions at the `.content` accessor in `src/lib/vaultApi.ts`. Frontmatter parsing keeps using the raw bytes, so YAML semantics are untouched.
- Extracted `TEMPLATE_EXPR_RE` and a `stripTemplateExpressions` helper into a tiny `src/lib/templateExprRegex.ts` to avoid a runtime import cycle between `vaultApi.ts` and `templateScope.ts` (the bridge loops back through vaultApi).
- `templateScope.ts` re-exports `TEMPLATE_EXPR_RE` so every existing consumer keeps its current import path.

## Files
- `src/lib/templateExprRegex.ts` — new module: shared regex + stripper.
- `src/lib/templateScope.ts` — re-exports the constant from the new module.
- `src/lib/vaultApi.ts` — splits content memoization (raw for frontmatter, stripped for `.content`).
- `src/lib/__tests__/vaultApi.test.ts` — 4 new cases (both user-reported shapes, multi-line strip, frontmatter-unaffected guard).

## Test plan
- [x] `pnpm test` — 1138/1138 pass.
- [ ] Manual: create a note with the exact user-reported template `{{(\"|test|test|\\n|-|-|\\n\"); vault.notes.where(n => n.content.contains(\"todo\")).select(f => \"|[[\" + f.name + \"]]|-|\").join(\"\\n\")}}` — verify the host note does not appear in the rendered table, and a second note with a genuine \"todo\" does.

### DIFF
--- a/src/lib/__tests__/vaultApi.test.ts
+++ b/src/lib/__tests__/vaultApi.test.ts
@@ -293,6 +293,22 @@ describe("createVaultRoot — n.content strips `{{ ... }}` regions (#325)", () =
     expect(hits).toEqual(["sub/b.md"]);
   });
 
+  it("strips adjacent expressions without leaving a fragment between them", () => {
+    // Guards the `g`-flag contract: consecutive `{{ a }}{{ b }}` match as
+    // two separate regions. A regex bug that folded them into one "greedy"
+    // match, or skipped the second, would leave stray text behind.
+    const bodies: Record<string, string> = {
+      "a.md": "start {{ a }}{{ b }} end\n",
+    };
+    const v = createVaultRoot(
+      mkStores({
+        readNoteContent: (p) => bodies[p] ?? null,
+      }),
+    );
+    const content = v.notes.first()!.content;
+    expect(content).toBe("start  end\n");
+  });
+
   it("frontmatter parsing is unaffected by the strip (uses raw content)", () => {
     // Frontmatter must see the verbatim file so YAML parses correctly even
     // when a `{{ ... }}` region lives in the body below it.
--- a/src/lib/__tests__/vaultApi.test.ts
+++ b/src/lib/__tests__/vaultApi.test.ts
@@ -225,3 +225,90 @@ describe("createVaultRoot — content-based filtering across all notes (#319)", 
     expect(hits).toEqual(["sub/b.md", "sub/c.md"]);
   });
 });
+
+// #325: template bodies in a note's own text must NOT contribute to `.content`
+// searches, otherwise a template like
+// `{{vault.notes.where(n => n.content.contains("todo"))}}` matches the host
+// note itself because the literal "todo" appears inside the expression body.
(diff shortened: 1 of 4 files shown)


herox215:
@minerva review

herox215:
@minerva review

PR #327, opened 2026-04-22
ci(minerva): post inline review comments on PRs

herox215:
## Summary
- Switch the Minerva workflow from one bulk PR comment to a proper pull-request review with inline `file:line` comments.
- Prompt GLM 5.1 for strict JSON findings; validate every `(path, line, side)` against the diff hunks before sending inline.
- Findings that fail validation are kept in a collapsed section of the review body, so nothing is silently dropped.
- Fall back to a plain PR comment if the model returns non-JSON or the reviews API rejects the payload.

## Test plan
- [x] Local smoke test of `minerva_build_review.py` against a synthetic diff + mix of valid/invalid findings (2 inline, 2 rejected).
- [ ] Trigger the workflow on this PR by commenting `@minerva` and confirm inline comments appear in "Files changed".
- [ ] Confirm the review body still renders the verdict + the "could not be placed inline" disclosure when applicable.

### DIFF
--- a/.github/scripts/minerva_build_review.py
+++ b/.github/scripts/minerva_build_review.py
@@ -1,0 +1,131 @@
+#!/usr/bin/env python3
+"""Validate Minerva JSON findings against a unified diff and build a GitHub
+pull-request review payload. Findings that point at a file:line outside the
+diff hunks are dropped from the inline set and listed in the review body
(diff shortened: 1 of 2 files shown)


herox215:
@minerva please review

PR #331, opened 2026-04-22
fix(links): ignore [[...]] inside {{ ... }} template bodies (#330)

### LINKED ISSUE #330: Links inside `{{ ... }}` template expressions should not be detected as outgoing links
## Problem

When a template expression contains string fragments that look like wikilinks, those fragments get picked up by the outgoing-links detector even though they are just code, not actual links.

### Repro

Write in a note:

```
{{("|test|test|\n|-|-|\n"); vault.notes.where(n => n.content.contains("todo")).select(f => "|[[" + f.name + "]]|-|").join("\n")}}
```

In the "Outgoing Links" panel on the right, `" + f.name + "` shows up as an outgoing link, because the `[[ ... ]]` inside the template source is parsed as a wikilink.

### Expectation

Anything inside `{{ ... }}` is code, not markdown. Link detection (and by extension the outgoing-links panel) should ignore the contents of template expressions entirely — not just filter the bogus results after the fact.

Filtering the output afterwards would work but is the wrong layer: the correct fix is to not treat `{{ ... }}` bodies as link-bearing content in the first place.

herox215:
Closes #330.

## Summary
- Wiki-link fragments inside a `{{ ... }}` template body (e.g. `{{ "[[" + f.name + "]]" }}`) no longer surface as outgoing links / backlinks / graph edges.
- Fix applied on both sides so the frontend sidebar and the Rust link graph stay consistent: `extractOutgoingLinks` (TS) and `extract_links` (Rust) both compute template-expression ranges up front and skip any match that overlaps a range.
- `templateExprRanges` moved from `components/Editor/` to `lib/` so non-UI code can consume it without reaching into component code, and its local regex is now the shared `TEMPLATE_EXPR_RE` from `lib/templateExprRegex.ts`.

## Why skip-by-range, not strip-and-parse
Pre-stripping `{{ ... }}` bodies would (1) shift line numbers of any real link that follows a multi-line template and (2) concatenate text across the removed region, which could synthesise a spurious match like `[[A{{x}}B]]` → `[[AB]]`. The range check preserves the input verbatim and dodges both hazards.

## Test plan
- [x] TDD red: new frontend tests in `src/lib/__tests__/outgoingLinks.test.ts` for single-line template, real + fake mixed, multi-line template with real link on a later line (line-number preservation), and the exact #330 repro — all failed against the old behaviour.
- [x] TDD red: matching Rust tests in `src-tauri/src/indexer/link_graph.rs`.
- [x] TDD green: all 1143 frontend tests pass; all 35 `indexer::link_graph` tests pass (4 new + 31 existing).
- [x] Tightened the pre-existing embed test from `toBeLessThanOrEqual(1)` to `toBe(1)` to prevent silent drift.
- [x] `pnpm typecheck` clean.
- [x] `cargo clippy` on `link_graph.rs` clean.
- [ ] Manually verify in the UI that the issue-#330 repro no longer surfaces a bogus entry.

## Follow-ups (not in this PR)
Two separate places share the same "regex over raw text, no template awareness" pattern and should get their own fixes:
- `src-tauri/src/commands/links.rs` — `update_links_after_rename`'s rewrite regex can modify template source code when the renamed file's stem happens to appear in a `{{ ... }}` body (data-corruption risk on rename). Different fix shape because the rewriter takes its own regex path, not `extract_links`.
- `src/lib/canvas/textTokens.ts` — the canvas text-node tokenizer has its own wiki-link regex and would surface fake links if someone drops template-y prose into a canvas card.

Both flagged during plan review; leaving as separate issues to keep this PR focused on #330 as reported.


src-tauri/src/indexer/link_graph.rs
```diff
@@ -30,6 +30,39 @@ fn wiki_link_regex() -> &'static Regex {
     })
 }
 
+/// Regex for the outer `{{ ... }}` boundary of template expressions.
+///
+/// Mirrors `src/lib/templateExprRegex.ts` so the frontend and backend agree
+/// on what counts as a template body. `[^{}]` rejects nested braces and
+/// allows newlines so multi-line expressions are stripped as one unit.
```
herox215: The docstring says multi-line expressions are "stripped as one unit" but this function does not strip — it returns ranges. The sentence was lifted from `stripTemplateExpressions` in `templateExprRegex.ts` and no longer describes what happens here (the caller skips overlapping matches without mutating the text, which is actually the whole point per `outgoingLinks.ts` #330 comment). Suggest: "allows newlines so multi-line expressions match as one contiguous range." Keeps the parity-with-JS intent clear without miscalling the operation.

src-tauri/src/indexer/link_graph.rs
```diff
@@ -30,6 +30,39 @@ fn wiki_link_regex() -> &'static Regex {
...
+/// No backtracking risk — the class is negated, the body is bounded by
+/// the first `}}` after a `{{`, and Rust's regex crate is linear.
+fn template_expr_regex() -> &'static Regex {
+    static RE: OnceLock<Regex> = OnceLock::new();
+    RE.get_or_init(|| {
+        Regex::new(r"\{\{[^{}]+?\}\}").expect("invalid template-expr regex")
```
herox215, line 43: Regex triplication risk. The template boundary now lives in three places: `src/lib/templateExprRegex.ts` (JS), `templateLivePreview` (shares the JS one), and this new Rust copy. The only thing keeping them in sync is a docstring. A parity test — small TSV / JSON fixture with inputs and expected match spans, consumed by a Vitest and a `#[test]` that deserialises it — would turn future drift into a CI failure instead of a silent divergence. Cheap insurance given how subtle `[^{}]` vs multiline differences can be between regex engines.

src-tauri/src/indexer/link_graph.rs
```diff
@@ -652,6 +710,47 @@ mod tests {
...
+        assert_eq!(links[0].line_number, 4);
+    }
+
+    #[test]
+    fn extract_links_issue_330_repro() {
+        // Exact shape from issue #330 — the table-generating template whose
+        // body happens to contain `"[[" + f.name + "]]"` as source code.
+        let md = r#"{{("|test|test|\n|-|-|\n"); vault.notes.where(n => n.content.contains("todo")).select(f => "|[[" + f.name + "]]|-|").join("\n")}}"#;
+        let links = extract_links(md);
+        assert!(links.is_empty(), "expected no links, got {:?}", links);
```
herox215, line 761: Test gap: no CRLF coverage. The `strip_suffix('\r')` / `raw_line.len() + 1` math is specifically designed to keep byte-offsets correct on CRLF input while still yielding LF-style `context`, but no test exercises that code path. Since Windows vaults and mixed-ending files are common, a fixture like `"line0\r\n{{\r\n\"[[\" + x + \"]]\"\r\n}}\r\n[[Real]]\r\n"` would pin both (a) the offset arithmetic across CR bytes and (b) the CRLF-stripped `context`. Without it, a future refactor of the loop could silently regress either.

src/lib/__tests__/outgoingLinks.test.ts
```diff
@@ -95,6 +95,47 @@ describe("extractOutgoingLinks", () => {
...
+      expect(res[0]?.target).toBe("Real");
+      expect(res[0]?.lineNumber).toBe(4);
+    });
+
+    it("repro from issue #330 — table-generating template", () => {
+      const doc =
+        '{{("|test|test|\\n|-|-|\\n"); vault.notes.where(n => n.content.contains("todo")).select(f => "|[[" + f.name + "]]|-|").join("\\n")}}';
+      const res = extractOutgoingLinks(doc, () => null);
+      expect(res).toEqual([]);
+    });
```
herox215, line 139: Semantic gap worth a pinning test. The PR decides — correctly, imo — that links are defined LITERALLY: if `[[X]]` appears only inside a template body, it is not an outgoing link, even when the template would render to an HTML `<a>` / another `[[X]]`. But there is no test that pins that choice, so a future contributor could "fix" what looks like a false negative (a template that clearly produces a link at runtime) without realising it is the intended semantics. Consider adding one more case to this `describe` block: a template body that would evaluate to a real wikilink at runtime (e.g. `{{ "[[RenderedTarget]]" }}`) with `expect(res).toEqual([])` and an explanatory comment. That turns the semantic decision into documentation-by-test.

herox215:
Scope call on the two follow-ups — the `links.rs` one is not equivalent to the `canvas/textTokens.ts` one and the PR should not bundle them as "two same-shape deferrals."

- `canvas/textTokens.ts` leak: a cosmetic false-positive in a canvas card. User sees a bogus entry; nothing persists or mutates. Fine as a follow-up.
- `links.rs` rename-rewriter: this one *writes to disk*. If a user renames a file whose stem appears in a `{{ ... }}` body on another note, `update_links_after_rename` will silently rewrite the template source. Thats data corruption to a *computed* artifact — the template output will change on the next render, and the user may not notice until much later. By contrast, this PR makes the `extract_links` side of that same flow template-aware, so the graph is correct but the on-disk content is now wrong. That is arguably worse than a consistent bug: backlinks say "no link here" while the file content now says otherwise, and both states came from the same rename.

At minimum I would want a tracking issue linked here, explicitly labelled data-loss / corruption (not just "follow-up"), and ideally a short-term guard that `update_links_after_rename` consults the same `template_expr_ranges` helper before accepting a rewrite offset. The helper now lives in a place the command layer can reach it, so the incremental cost is small.

Everything else in the PR is well-reasoned — the skip-by-range vs strip-and-parse rationale in `outgoingLinks.ts` is exactly right (line-number preservation + seam-synthesis avoidance), and the file move from `components/Editor/` to `lib/` is the correct direction now that non-UI code depends on it.

src-tauri/src/indexer/link_graph.rs
```diff
@@ -30,6 +30,39 @@ fn wiki_link_regex() -> &'static Regex {
...
+/// Used by `extract_links` to skip any wiki-link whose match range falls
+/// inside a template body (#330): content inside `{{ ... }}` is template
+/// source code, not Markdown, so fragments like `"[[" + f.name + "]]"`
+/// must not surface as outgoing links / backlinks / graph edges.
+fn template_expr_ranges(content: &str) -> Vec<(usize, usize)> {
+    template_expr_regex()
+        .find_iter(content)
+        .map(|m| (m.start(), m.end()))
+        .collect()
+}
```
herox215, line 68: Incidental perf note, probably non-blocking. `extract_links` is called per file during initial indexing (`indexer/mod.rs`) AND once per rewritten file in `apply_rewrites_to_graph` during a rename burst. Each call now does a full-content `template_expr_regex().find_iter(...).collect()` even when the file has zero `{{`. For the common case (no templates anywhere) the allocation is empty but still costs a regex pass over the entire content. A one-line cheap pre-check — `if !content.contains("{{") { None } else { Some(collect...) }` — would elide the regex + Vec for the 99% case without changing correctness. Not required for this PR but worth keeping in mind if you see `extract_links` in a flamegraph later.

src-tauri/src/indexer/link_graph.rs
herox215: Fixed in 29898e8. Docstring on `template_expr_regex` is now `"allows newlines so multi-line expressions match as one contiguous range"` — no more lifted "stripped" wording.

src-tauri/src/indexer/link_graph.rs
herox215: Done in 29898e8. Shared fixture at `test-fixtures/template_expr_parity.json` (11 cases: simple, multi-line, adjacent with/without separator, brace-rejection, empty-body rejection, UTF-8, the #330 repro, boundary positions). Consumed by `src/lib/__tests__/templateExprParity.test.ts` on the JS side and a `#[test]` in `indexer::link_graph` on the Rust side. Matches are compared as substrings rather than offsets because JS regex indexes in UTF-16 code units while Rust indexes in bytes — the substring view is the language-agnostic invariant. Any future drift on one side now fails the other sides CI.

src-tauri/src/indexer/link_graph.rs
herox215: Added in 29898e8 as `extract_links_handles_crlf_line_endings`. Fixture is exactly your suggested shape (`line 0\r\n{{...\r\n[[Real]]\r\n`) and pins both (a) the byte-offset arithmetic across CR bytes and (b) the CRLF-stripped `context`. Any future rewrite of the loop that regresses either half will fail the test.

src/lib/__tests__/outgoingLinks.test.ts
herox215: Pinned in 29898e8 as `ignores [[...]] even when the template would RENDER to a real link`. Uses your exact suggestion (`{{ "[[RenderedTarget]]" }}` → `[]`) and the accompanying comment spells out the design decision: outgoing links reflect what the author wrote in source, not what a template would produce on render (which depends on vault state and changes over time). Future contributor seeing this as a "false negative" now has documentation-by-test pointing them at the design choice.

src-tauri/src/indexer/link_graph.rs
herox215: Added in 29898e8 — `template_expr_ranges` now short-circuits with `if !content.contains("{{") { return Vec::new(); }` before running the regex. Cost for the common "no templates in this note" case is now one `memmem` scan for `"{{"` instead of a full regex pass + `Vec` allocation. Called out in the docstring so future readers see why the early-out exists.

herox215:
Addressed in 29898e8.

**Rename rewriter guard (the data-corruption half):** fixed in this PR, not deferred. `update_links_after_rename` now routes its rewrite through a new `rewrite_wiki_links_respecting_templates` helper that consults `link_graph::template_expr_ranges` / `overlaps_any` (now `pub(crate)`) and preserves any `[[old_stem]]` occurrence whose match range overlaps a `{{ ... }}` body. The helper is extracted out of the `rayon` closure precisely so the guard is directly unit-testable — 5 new `#[test]`s in `commands::links::tests` pin the behaviour, including the scenario you described (rename `foo` → `bar` where `[[foo]]` appears inside a template body must leave the template byte-identical while still rewriting real `[[foo]]` elsewhere in the file).

You were right that bundling this with the canvas leak under "follow-ups" understated the risk. Read and write halves of the rename flow are now both template-aware, so the inconsistency you called out (graph says "no link" but on-disk content gets rewritten anyway) is closed.

**Canvas text-tokenizer (the cosmetic half):** tracked as #332, labelled `bug`, with a pointer to the skip-by-range approach used in `outgoingLinks.ts` so whoever picks it up has the template to copy.

Thanks for pushing on this — the two really weren't the same risk class and the PR description framing was wrong.

src-tauri/src/commands/links.rs
```diff
@@ -49,6 +49,47 @@ pub struct RenameResult {
...
+    let template_ranges = link_graph::template_expr_ranges(content);
+    let mut link_count = 0usize;
+    let new_content = re
+        .replace_all(content, |caps: &regex::Captures| {
+            let whole = caps.get(0).expect("capture 0 always present");
+            if link_graph::overlaps_any(&template_ranges, whole.start(), whole.end()) {
+                return whole.as_str().to_string();
+            }
+            link_count += 1;
+            let alias_part = caps.get(1).map(|m| m.as_str()).unwrap_or("");
```
herox215, line 86: Capture-index fragility. This helper reads `caps.get(1)` for the alias under an implicit contract with the pattern built at line 334 (`\[\[<stem>(\|[^\]]*)?(\]\])`). The trailing `(\]\])` is a capture group that is never consumed, so anyone adding a new earlier group — or promoting the `]]` tail to a named one — silently shifts group indices and this helper rewrites with no alias (or worse, with `]]` as alias). Consider either a named capture for the alias (`(?P<alias>\|[^\]]*)?`) and reading `caps.name("alias")`, or dropping the now-pointless `(\]\])` capture to `(?:\]\])` so there is only one optional group left and intent is obvious at a glance.

test-fixtures/template_expr_parity.json
```diff
@@ -0,0 +1,60 @@
...
+      "name": "expression at start",
+      "input": "{{x}} then prose",
+      "matches": ["{{x}}"]
+    },
+    {
+      "name": "expression at end no trailing newline",
+      "input": "prose then {{x}}",
+      "matches": ["{{x}}"]
+    }
+  ]
```
herox215, line 59: Fixture coverage gap, low severity. The 11 cases pin the canonical behaviours but miss a few edges where JS and Rust engines could plausibly disagree: (a) whitespace-only body `{{ }}` (matches both today but the `+?` contract is easy to regress), (b) `\r` inside the body (`{{a\rb}}`) — relevant on Windows / pasted content, (c) body touching EOF without a trailing newline or separator, (d) overlapping opens (`{{{{x}}` — both engines currently start the match at offset 2, worth pinning). I verified all four agree across engines today, so this is a strengthening request, not a bug report. Adding them as fixture cases turns any future drift into CI noise rather than a production surprise.

src-tauri/src/indexer/link_graph.rs
```diff
@@ -30,6 +30,49 @@ fn wiki_link_regex() -> &'static Regex {
...
+    }
+    template_expr_regex()
+        .find_iter(content)
+        .map(|m| (m.start(), m.end()))
+        .collect()
+}
+
+/// Returns true iff the half-open interval `[from, to)` overlaps any of
+/// `ranges`. Linear scan — fine at the expression counts seen in practice.
+pub(crate) fn overlaps_any(ranges: &[(usize, usize)], from: usize, to: usize) -> bool {
```
herox215, line 72: Nit: the `pub(crate)` promotion is the minimum-viable exposure and works, but `template_expr_ranges` + `overlaps_any` now have two distinct callers (`extract_links` here and `rewrite_wiki_links_respecting_templates` in `commands::links`). If a third caller shows up — for example a search-index filter that wants to skip template bodies — extracting them into `indexer/template_expr.rs` (alongside the regex itself) becomes natural. Not worth doing now for two callers, but worth noting so the next feature that reaches for them knows the split is expected rather than accidental.

herox215:
## Second-round review

All six first-round points are genuinely resolved (not just claimed):

- **Docstring wording** — fixed verbatim on line 37.
- **Regex triplication** — shared fixture at `test-fixtures/template_expr_parity.json` (11 cases) is consumed by both the Vitest suite and a new Rust `#[test]`; substring comparison is the correct call given JS-UTF-16 vs Rust-bytes offset semantics. I ran both suites locally, all green.
- **Perf short-circuit** — `content.contains("{{")` is inside `template_expr_ranges` itself, so every caller (including the new rename-rewriter helper) benefits, not just `extract_links`. Good call placing it at the function boundary rather than the call site.
- **CRLF coverage** — `extract_links_handles_crlf_line_endings` pins the exact byte-math invariant I was worried about (multi-line template spanning CRLF lines, real link on a subsequent CRLF line).
- **Literal-vs-rendered semantic pin** — test is there and the surrounding comment is explicit that any "fix" is actually a design change. Nicely defensive.
- **Rename-rewriter data-loss risk** — went beyond what I asked (a tracking issue or short-term guard) and fixed it in this PR via `rewrite_wiki_links_respecting_templates` + five unit tests. This was the structurally important one and it is now consistent with the read path.

Canvas-rename follow-up filed as #332 is the right scope call.

### Residual items (all non-blocking)

Three inline comments below flag lower-severity follow-ups: capture-index fragility in the rename-regex contract, a handful of fixture cases worth adding for stronger drift detection, and a forward-looking note on where the shared `template_expr_ranges` / `overlaps_any` pair should probably live if a third caller appears. None are merge-blocking — this is substantially better than the base PR and addresses the real risk (silent data corruption during rename).

LGTM from my seat modulo the optional nits. Not issuing a formal approve (commenter role).

PR #336, opened 2026-04-22
perf(sidebar): virtualize tree + single reveal subscription (#253)

### LINKED ISSUE #253: perf(sidebar): TreeNode has no virtualization and every node subscribes to treeRevealStore — O(expanded-nodes) work per reveal
## Observed problem

The file-tree sidebar renders a recursive `TreeNode` for every visible entry and has **no list virtualization**. Each `TreeNode` instance also owns a `treeRevealStore.subscribe(...)` lifecycle, meaning a vault with 10,000 expanded folders has 10,000 live store subscribers that all fire on every reveal event. This violates the \"stay fluid at 100k notes\" core value directly at the entry-point UI.

### Evidence

- `src/components/Sidebar/Sidebar.svelte` — root entries rendered in `{#each ... (entry.path)}` → `<TreeNode>` with no windowing (no `svelte-virtual-list`, no IntersectionObserver).
- `src/components/Sidebar/TreeNode.svelte` — recursive; every instance holds local `$state` and subscribes to `treeRevealStore`. No cleanup coalescing across siblings.
- Scroll cost is linear in \*rendered-DOM-node count\*, so even collapsed-to-root navigation is O(root-count). A vault with a flat root directory of 20k notes creates 20k DOM rows on mount.

Adjacent signal (issue #116): Svelte 5 reactivity bugs already shipped in `TreeNode` — the recursion + per-node reactive graph is already fragile.

### Measured cost bucket

- DOM: ~30 bytes/row × 10k rows ≈ 300kB of DOM nodes resident when sidebar is open. Multiply by Svelte 5 reactive proxy overhead.
- Subscription fan-out: every `treeRevealStore.set(...)` (fires on active-tab change, rename, create, move) runs N subscriber callbacks. Per-keystroke in VaultLayout's activeTabReveal subscription (#TODO separate ticket) compounds this.

### Target performance from CLAUDE.md

> Cold start < 3 s, warm start < 5 s, open note < 100 ms. RAM idle < 100 MB, active < 250 MB.

With the current tree, a 100k-vault user who expands ~half of folders is nowhere near these.

## Suggested approach

1. **Virtualize the flat rendered list.** Flatten the visible tree into an array-of-rows (parent/child with a `depth` field) and render via a virtual list (svelte-virtual-list, or a tiny in-house windowing component — CM6 itself has a viewport model; the sidebar can too).
2. **Lift `treeRevealStore` subscription out of `TreeNode`.** A single `Sidebar`-level derived maps the reveal target to a `path → boolean` set; each node reads the flag via props/$derived instead of owning a subscription.
3. **Lazy expansion**: don't materialise children of collapsed folders. Currently (check) the recursion builds DOM for the closed branch? Confirm and fix if so.
4. Measure before/after: DOM-node count + rAF-scroll FPS on a vault with 10k files.

## Expected impact

**High.** This is the biggest RAM + frame-budget win in the frontend for power users. Sidebar scroll should go from jerky → 60fps on 10k-file vaults; RAM savings in the 50–150MB range depending on expand state.

## Risk

Medium. Virtualization changes drag-and-drop semantics (the draggable row can scroll out of the DOM mid-drag) — solve by keeping the dragged row anchored outside the virtual-list window. Reveal-to-row behavior needs special-case scroll-into-view when the target row is outside the window.

## Acceptance criteria

- [ ] Sidebar renders at most O(visible-viewport) DOM rows regardless of expanded-folder count.
- [ ] Only one live `treeRevealStore` subscription for the entire sidebar (enforced by audit).
- [ ] Benchmark: scroll a 10k-file sidebar at p95 > 55fps on reference hardware.
- [ ] DnD, rename inline, context menu, and reveal-active-tab still work correctly at vault sizes of 1k and 10k files.
- [ ] No regression in existing Sidebar/TreeNode Vitest + WDIO tests.

herox215:
## Summary

Closes #253.

- **Single reveal subscription.** `Sidebar.svelte` now owns the only `treeRevealStore` subscription. Per-node subscriptions are gone — one handler expands ancestors, awaits their `listDirectory` loads, re-flattens, then scrolls.
- **Virtualized flat-row renderer.** A new `src/lib/flattenTree.ts` turns the `DirEntry` tree + persisted `expanded` set into a `FlatRow[]`. Sidebar renders only the viewport slice (ROW_HEIGHT=28, OVERSCAN=10) with absolute-positioned rows, so a 100k-note vault no longer mounts 100k components.
- **Per-row key preserves focus.** `TreeNode.svelte` is renamed to `TreeRow.svelte` and keyed by `path` in the `{#each}` block. Inline-rename focus, active-path highlight, and drag state survive scroll recycling.
- **Lazy-load contract intact.** `listDirectory` is still invoked only on first expand of a folder. Reveal awaits the load before flattening so the target row exists when we scroll.
- **Svelte 5 reactivity fix.** `vaultStore.subscribe` moved into `onMount` — a top-level-script subscription fires during component construction, before the reactive root is fully wired, so async `treeState` assignments can fail to propagate through `$derived`.

## Test plan

- [x] `pnpm vitest run src/components/Sidebar` — 11 files / 26 tests passing, including the new virtualization, reveal, reveal-subscription, flattenTree, and inline-rename-scroll suites.
- [x] `pnpm vitest run src/styles/__tests__/modalTheming.test.ts` — 35 passing (confirms the TreeRow-scoped confirm-dialog keeps its vc-modal-scrim / vc-modal-surface contract).
- [x] `pnpm vitest run src/lib/__tests__ src/store/__tests__ src/styles/__tests__` — 37 files / 500 tests passing (no regressions in adjacent modules).
- [x] `pnpm exec tsc --noEmit` — clean.
- [x] `pnpm exec svelte-check` — 0 errors (2 pre-existing line-clamp warnings in SearchResultRow/BacklinkRow, unrelated to this PR).
- [ ] WDIO specs (`sidebar-layout`, `tree-context-menu`, `drag-drop`, `inline-rename`) — not run in this environment: `wdio.conf.ts` requires a release `src-tauri/target/release/vaultcore` binary + `tauri-driver`, which aren't available on the agent host. Recommend running `pnpm test:e2e:build` locally before merge.

## Risk

- High-blast-radius refactor of the most-used view. Focus + drag-drop + context-menu + inline-rename all share the same recycler; the keyed `{#each}` preserves Svelte component identity across slice changes, which is what keeps those behaviours working. Covered by the updated `TreeNodeContextMenu` / `TreeNodeCreateFile` / `TreeNodeRecursion` / `TreeNodeInitialExpand` suites plus the new `inlineRenameScroll` test.
- Persisted tree state (`expanded` set keyed by `vaultHashKey`) is unchanged on disk — existing users keep their expansion state on upgrade.

### DIFF
--- a/src/components/Sidebar/Sidebar.svelte
+++ b/src/components/Sidebar/Sidebar.svelte
@@ -207,34 +207,64 @@
   // ─── Toggle expand / persist ───────────────────────────────────────────────
   async function onToggleExpand(row: FlatRow) {
     if (!row.isDir) return;
-    const expanded = new Set(treeState.expanded);
-    const willExpand = !expanded.has(row.relPath);
-    if (willExpand) expanded.add(row.relPath);
-    else expanded.delete(row.relPath);
-    treeState = { ...treeState, expanded: Array.from(expanded) };
-    void saveTreeState($vaultStore.currentPath ?? "", treeState);
+    const willExpand = !treeState.expanded.includes(row.relPath);
     if (willExpand) {
+      // Optimistically show the folder as expanded (with spinner) while the
+      // load is in flight. Persist only once listDirectory resolves — matches
+      // #336 B3 so a failure doesn't stick in `expanded` across sessions.
+      const expanded = new Set(treeState.expanded);
+      expanded.add(row.relPath);
+      treeState = { ...treeState, expanded: Array.from(expanded) };
       const existing = folders.get(row.path);
       if (!existing || !existing.childrenLoaded) {
-        await loadFolder(row.path);
+        const result = await loadFolder(row.path);
+        if (result === null) {
+          // Roll back the optimistic expand so the folder collapses and the
+          // next session doesn't show an empty, un-retryable expanded node.
+          const rolled = new Set(treeState.expanded);
+          rolled.delete(row.relPath);
+          treeState = { ...treeState, expanded: Array.from(rolled) };
+          void saveTreeState($vaultStore.currentPath ?? "", treeState);
+          return;
+        }
       }
+      void saveTreeState($vaultStore.currentPath ?? "", treeState);
+    } else {
+      const expanded = new Set(treeState.expanded);
+      expanded.delete(row.relPath);
+      treeState = { ...treeState, expanded: Array.from(expanded) };
+      void saveTreeState($vaultStore.currentPath ?? "", treeState);
     }
   }
 
   async function setExpanded(relPath: string, absPath: string, on: boolean) {
-    const expanded = new Set(treeState.expanded);
-    if (on) expanded.add(relPath);
-    else expanded.delete(relPath);
-    treeState = { ...treeState, expanded: Array.from(expanded) };
-    // Fire-and-forget the persistence — it uses crypto.subtle.digest which
-    // can hang several ticks; we don't want that delaying the user-visible
-    // listDirectory that populates the expanded folder.
-    void saveTreeState($vaultStore.currentPath ?? "", treeState);
     if (on) {
+      // #336 B3 — do NOT persist `expanded` until listDirectory resolves
+      // successfully. If it fails, the folder stays collapsed so the next
+      // session doesn't show an empty, un-retryable expanded folder.
       const existing = folders.get(absPath);
-      if (!existing || !existing.childrenLoaded) {
-        await loadFolder(absPath);
+      const needsLoad = !existing || !existing.childrenLoaded;
+      if (needsLoad) {
+        const result = await loadFolder(absPath);
+        if (result === null) {
+          // Load failed — leave treeState.expanded untouched.
+          return;
+        }
       }
+      if (treeState.expanded.includes(relPath)) return;
+      const expanded = new Set(treeState.expanded);
+      expanded.add(relPath);
+      treeState = { ...treeState, expanded: Array.from(expanded) };
+      // Fire-and-forget the persistence — it uses crypto.subtle.digest which
+      // can hang several ticks; we don't want that delaying the user-visible
+      // listDirectory that populates the expanded folder.
+      void saveTreeState($vaultStore.currentPath ?? "", treeState);
+    } else {
+      if (!treeState.expanded.includes(relPath)) return;
+      const expanded = new Set(treeState.expanded);
+      expanded.delete(relPath);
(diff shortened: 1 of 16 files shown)


herox215:
## Review blockers addressed (977d877)

Pushed a fix commit covering the three blockers flagged in review.

### B1 — stale `$derived` in store-subscription callback
`performReveal` now recomputes the flat list via `flattenTree(treeModel)` directly after `await tick()` instead of reading the `flatRows` `$derived`. This sidesteps the Svelte 5 pitfall where a `$derived` read from inside a store callback can serve its previous value even across an `await tick()`. Defensive — the race is hard to reliably reproduce in jsdom but it matches the user-memory pitfall exactly.

### B2 — reveal races the virtual window / 0-height scroller
- If `scrollerEl.clientHeight === 0` (collapsed sidebar / pre-layout mount), the reveal path is stashed in `pendingRevealPath` and drained from `measureViewport` once the scroller has real layout.
- After setting `scrollTop`, we now poll `requestAnimationFrame` up to 3 frames for the target `[data-tree-row]` element before accepting best-effort.

### B3 — `expanded` persists before `loadFolder` resolves
Both `setExpanded` (reveal path) and `onToggleExpand` (user click) now roll back the `treeState.expanded` mutation if `loadFolder` returns `null`. `onToggleExpand` still optimistically expands for user feedback but rolls back + saves on failure.

### Regression tests
- `src/components/Sidebar/__tests__/sidebarRevealAfterVaultLoad.test.ts` (B1 guard)
- `src/components/Sidebar/__tests__/sidebarRevealPending.test.ts` (B2 — fails pre-fix)
- `src/components/Sidebar/__tests__/expandFailureRollback.test.ts` (B3 — fails pre-fix)

Also updated `sidebarReveal.test.ts` to prime `clientHeight` so the existing reveal test still works with the new B2 0-height guard.

### Verification
- `pnpm vitest run src/components/Sidebar` — 14 files / 29 tests passing
- `pnpm exec tsc --noEmit` — clean
- `pnpm exec svelte-check` — 0 errors (2 pre-existing line-clamp warnings unchanged)

### Not in scope (follow-ups)
- **B4** (WDIO specs) — will run locally before merge.
- **B5** (DnD auto-scroll + drop-target coverage under virtualization) — tracked in #337.

Issue #250, opened 2026-04-20
perf(link_graph): resolve_link does two full passes + per-link to_lowercase on every vault file


herox215:
## Observed problem

`src-tauri/src/indexer/link_graph.rs::resolve_link` (lines 142–175) is called once per wiki-link during `update_file`. For each call it:

1. **Stage 1** (line 151): iterates `all_rel_paths` and calls `path_stem(p).to_lowercase()` on every path to compare against `stem_lower`.
2. **Stage 2** (lines 158–161): iterates `all_rel_paths` **again** and calls `path_stem(p).to_lowercase()` **again** on every path.

So each `resolve_link` call is ~2N stem-lowercasings where N = vault size.

`update_file` (link_graph.rs:229) calls `resolve_link` once per link. During the cold-start walk in `indexer/mod.rs::index_vault`, this runs for every link in every newly-indexed file — so the total cold-start cost is roughly **O(total_links × vault_size)** with a per-iteration `String` allocation for the lowercase conversion.

At 100k notes × ~10 links/note × 100k paths × 2 passes ≈ 2 × 10¹¹ string comparisons for cold start, each with an allocation. This is the dominant factor in the open #179 double-read profile after the double-read itself is fixed.

## Evidence / hypothesis

- Each Stage-1 / Stage-2 iteration allocates a fresh `String` via `to_lowercase()` for every candidate — at 100k × 20 links × 100k ≈ 10¹¹ transient `String` allocations over an initial index.
- A precomputed `HashMap<String /* stem_lower */, Vec<&str /* rel_path */>>` built once per `index_vault` would reduce the inner loop to a single hashmap lookup.
- The second pass (stage 2+3) is only needed when stage 1 misses — and stage 1 is a subset of the stem set. Build the stem→paths map once, then stage 1 / stage 2 become filter operations on a small candidate list.

## Suggested approach

1. Add an overload `resolve_link_with_index(target_raw, source_folder, stem_index: &HashMap<String, Vec<String>>)` that skips the two full passes.
2. In `index_vault` and `indexer/mod.rs::process_index_queue` (wherever `update_file` is called in bulk), build the stem index once per batch and pass it to each call.
3. For single-file updates (watcher events, single save), keep the current API but have it build a local stem map from `all_rel_paths` once at entry — one allocation instead of 2N.

## Expected impact

[...]

PR #342, opened 2026-04-23
fix(indexer): dispatch index updates from all self-write paths (#339)

### LINKED ISSUE #339: correctness: link/tag graph goes stale on external file changes
## Problem

When a vault is modified outside the app (git pull, file sync, manual edit in another editor), the link graph and tag index become stale until the next full reindex.

`commands::vault::merge_external_change()` rewrites file content via the merge engine but does **not** enqueue `IndexCmd::UpdateLinks` / `IndexCmd::UpdateTags`. As a result:

- Backlinks panel shows outdated references
- Tag panel shows outdated counts
- Unresolved-links report is wrong
- User only sees fresh data after a full vault reindex

## Affected code

- `src-tauri/src/commands/vault.rs` — `merge_external_change()` path
- `src-tauri/src/watcher.rs:115-120` — reference implementation for how Watcher dispatches `UpdateLinks`/`UpdateTags`
- `src-tauri/src/indexer/mod.rs` — `IndexCmd` enum

## Expected behavior

After any code path that rewrites vault content (merge, rename cascade, external-change reconciliation), link graph and tag index must be updated incrementally for the affected files, with the same semantics as the Watcher path.

## Scope

- Hook link/tag updates into `merge_external_change()`
- Audit other write paths (rename cascade, repair flows) for the same gap
- Add a test that mutates a file externally, triggers merge, and asserts backlinks/tags reflect the change without full rebuild

## Boy Scout

- If vault-scope guard is duplicated in these command paths, extract a shared helper
- If `IndexCmd::Update*` dispatch is repeated, wrap in a small helper

## Acceptance

- [ ] External change → merge → backlinks reflect new/removed links immediately
- [ ] External change → merge → tag index reflects new/removed tags immediately
- [ ] Rust test covering the external-change → merge → graph-update flow
- [ ] No full reindex required

herox215:
Closes #339.

## Summary

External merges, renames, moves, and deletes now keep the in-memory LinkGraph, TagIndex, and Tantivy index in sync with disk. The watcher's natural dispatch is suppressed by `write_ignore` on self-mutations, so without explicit dispatches the graphs trailed the true on-disk state until cold restart — surfacing as wrong backlinks, stale tag counts, and missing fulltext hits for renamed / merged / deleted files.

## Key changes

- **`merge_external_change` is now authoritative on clean**: records `write_ignore`, writes merged bytes to disk, returns `new_hash`, dispatches `UpdateLinks` / `UpdateTags` / Tantivy `AddFile` + `Commit`. On conflict: no write, no dispatch (watcher already dispatched for the external content; the local buffer takes the recovery path via the next autosave).
- **Frontend no longer double-writes** on clean-merge (#339 Socrates review). `EditorPane` save-path and `externalChangeHandler` consume `new_hash` directly so `lastSavedHash` aligns with what the backend wrote.
- **New shared helper `commands/index_dispatch.rs`** (`dispatch_self_write`, `dispatch_self_delete`) consolidates the self-write index contract so every write path goes through the same dispatcher.
- **`delete_file`** now removes `LinkGraph` + `TagIndex` entries (previously only Tantivy-delete via the embed path; the two graphs went stale).
- **`rename_file` / `move_file`** dispatch the NEW side so fulltext, tags, and link graph reflect the renamed content immediately.
- **`update_links_after_rename`** adds Tantivy `AddFile` per rewritten source so fulltext hits for the old stem don't linger. Link-graph direct mutation stays (perf, #250).

## Architectural notes

- Helper lives in `commands/` (not `indexer::`) because it knows about `VaultState` and vault-relative path normalisation — command-layer concerns the indexer must not depend on backwards.
- `MergeCommandResult` gained an optional `new_hash` field. `Some` on clean (backend just wrote), `None` on conflict. Frontend falls back to `sha256Hex(merged_content)` defensively.

## Test plan

- [x] **TDD**: new `src-tauri/src/tests/merge_external_change.rs` (13 tests) covers:
  - Clean: add link, remove link, add tag, remove tag, disk write, `write_ignore` record, `new_hash`, alias drop through frontmatter refresh
  - Conflict: keeps local, no write, no dispatch
  - Edge: non-md defensive branch, path outside vault, channel-absent (no IndexCoordinator)
  - Delete: `RemoveLinks` + `RemoveTags` dispatch actually fires
- [x] `cargo test --lib` — 387 tests passed, 0 failed, 12 ignored
- [x] Frontend Vitest sweep — 297 tests passed across 36 files
- [x] `tsc --noEmit` — clean
- [ ] Manual UAT: rename a heavily-linked note in a real vault; verify backlinks, tag counts, fulltext refresh without restart

## Review

Plan reviewed by **Socrates**. REVISE verdict addressed:
- Frontend double-write fixed here, not deferred
- Helper placed in command layer, not `indexer::mod`
- Tantivy `AddFile` added to all self-write paths (closes the fulltext-staleness gap that was papered over as "already correct" for `write_file`)
- `update_links_after_rename` Tantivy gap folded in
- `merge_external_change` split into `_impl` + thin `#[tauri::command]` wrapper per `files.rs:16-20` convention
- Step order flipped: merge fix landed inline and green before extracting the shared helper

### DIFF
--- a/src-tauri/src/commands/files.rs
+++ b/src-tauri/src/commands/files.rs
@@ -22,7 +22,7 @@
 use crate::error::VaultError;
 use crate::hash::hash_bytes;
 use crate::indexer::memory::FileMeta;
-use crate::indexer::{walk_md_files, IndexCmd};
+use crate::indexer::walk_md_files;
 use crate::VaultState;
 use regex::Regex;
 use std::path::{Path, PathBuf};
@@ -114,14 +114,16 @@ pub async fn write_file(
         _ => VaultError::Io(e),
     })?;
 
-    // BUG-05.1 FIX: the watcher would normally dispatch UpdateLinks/UpdateTags
-    // for any modify event, but write_ignore suppresses self-writes to avoid
-    // double-indexing. That leaves the in-memory LinkGraph and TagIndex stale
-    // after every auto-save, so user observes stale tag counts, broken
-    // backlinks, and dangling unresolved-link colors until cold restart.
-    //
-    // Dispatch the updates directly here so the in-memory indexes stay in sync.
-    dispatch_index_updates(&state, &final_path, &content).await;
+    // BUG-05.1 FIX + #339: the watcher would normally dispatch
+    // UpdateLinks/UpdateTags/AddFile for any modify event, but write_ignore
+    // suppresses self-writes to avoid double-indexing. That leaves the
+    // in-memory LinkGraph, TagIndex, and Tantivy index stale after every
+    // auto-save, so users observe stale tag counts, broken backlinks,
+    // dangling unresolved-link colors, and stale fulltext hits until cold
+    // restart. Shared helper in commands/index_dispatch.rs keeps this in
+    // sync across all self-write paths (write_file, merge_external_change,
+    // rename/move, update_links_after_rename).
+    crate::commands::index_dispatch::dispatch_self_write(&state, &final_path, &content).await;
 
     // Embed-on-save (#196): non-blocking enqueue to the EmbedCoordinator.
     // Sync, returns immediately; a `QueueFull` outcome is benign because
@@ -135,7 +137,7 @@ pub async fn write_file(
 }
 
 #[cfg(feature = "embeddings")]
-fn dispatch_embed_update(state: &VaultState, abs_path: PathBuf, content: &str) {
+pub(crate) fn dispatch_embed_update(state: &VaultState, abs_path: PathBuf, content: &str) {
     let coord_handles = {
         let Ok(guard) = state.embed_coordinator.lock() else { return };
         guard.as_ref().map(|c| (c.tx.clone(), std::sync::Arc::clone(&c.pending)))
@@ -162,49 +164,6 @@ pub(crate) fn dispatch_embed_delete(state: &VaultState, abs_path: PathBuf) {
     if let Err(e) = probe.enqueue_delete(abs_path) {
         log::warn!("embed delete enqueue: {e}");
     }
-}
-
-/// Dispatch IndexCmd::UpdateLinks and IndexCmd::UpdateTags for a path we just
-/// wrote from the backend. Called from write_file because write_ignore
-/// suppresses the natural watcher-driven dispatch.
-///
-/// Best-effort: if the IndexCoordinator is not yet initialized (vault not
-/// open, or during boot) or the channel is full, we silently drop — the
-/// next cold-start rebuild will re-populate correctly.
-async fn dispatch_index_updates(state: &VaultState, abs_path: &Path, content: &str) {
-    // Get vault root so we can compute a vault-relative path.
-    let vault_root = {
-        let Ok(guard) = state.current_vault.lock() else { return };
-        match guard.as_ref() {
-            Some(p) => p.clone(),
-            None => return,
-        }
-    };
-
-    let Ok(rel) = abs_path.strip_prefix(&vault_root) else { return };
-    let rel_path = rel.to_string_lossy().replace('\\', "/");
-
-    // Clone the sender Arc — do NOT hold the coordinator lock across .await.
-    let tx = {
-        let Ok(guard) = state.index_coordinator.lock() else { return };
-        match guard.as_ref() {
-            Some(c) => c.tx.clone(),
-            None => return,
-        }
-    };
-
-    let _ = tx
-        .send(IndexCmd::UpdateLinks {
-            rel_path: rel_path.clone(),
-            content: content.to_string(),
-        })
-        .await;
-    let _ = tx
-        .send(IndexCmd::UpdateTags {
-            rel_path,
-            content: content.to_string(),
-        })
-        .await;
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
@@ -534,12 +493,22 @@ pub async fn rename_file(
         },
     );
 
-    // #277: the watcher would normally dispatch DeleteFile(old) + AddFile(new)
-    // on a rename event, but write_ignore suppresses that for self-mutations.
-    // Dispatch a Tantivy delete for the OLD path so fulltext search stops
-    // returning the stale doc; the NEW path is re-indexed on next user save.
+    // #277 + #339: the watcher would normally dispatch
+    // DeleteFile(old) + AddFile(new) on a rename event, but write_ignore
+    // suppresses that for self-mutations. Dispatch RemoveLinks/RemoveTags
+    // + Tantivy delete for the OLD path, and UpdateLinks/UpdateTags +
+    // Tantivy AddFile for the NEW path using the freshly-read content.
+    // Without the new-side dispatch the renamed file's body text stays
+    // absent from fulltext search until cold restart, and its tags/links
+    // hang under the OLD rel_path in the graphs until the user edits it.
     if let Some(old) = old_canonical_for_tantivy {
-        dispatch_tantivy_delete(&state, old).await;
+        crate::commands::index_dispatch::dispatch_self_delete(&state, &old).await;
+    }
+    let new_canonical = std::fs::canonicalize(&result.new_path).ok();
+    if let Some(new_path) = new_canonical.as_ref() {
+        if let Ok(content) = std::fs::read_to_string(new_path) {
+            crate::commands::index_dispatch::dispatch_self_write(&state, new_path, &content).await;
+        }
     }
 
     #[cfg(feature = "embeddings")]
@@ -547,20 +516,6 @@ pub async fn rename_file(
         dispatch_embed_delete(&state, p);
     }
     Ok(result)
-}
-
-/// Enqueue a Tantivy delete for `path`. Called from rename/move paths where
-/// write_ignore would otherwise suppress the watcher-driven DeleteFile.
-async fn dispatch_tantivy_delete(state: &VaultState, path: PathBuf) {
-    let tx = {
-        let Ok(guard) = state.index_coordinator.lock() else { return };
-        match guard.as_ref() {
-            Some(c) => c.tx.clone(),
-            None => return,
-        }
-    };
-    let _ = tx.send(IndexCmd::DeleteFile { path }).await;
-    let _ = tx.send(IndexCmd::Commit).await;
 }
 
 // ─── delete_file ─────────────────────────────────────────────────────────────
@@ -617,6 +572,13 @@ pub async fn delete_file(
             new_path: None,
         },
     );
+    // #339: RemoveLinks/RemoveTags + Tantivy delete for the deleted path.
+    // Without this the backlinks panel and tag counts hang on the deleted
+    // file's old references until cold restart (the watcher's natural
+    // dispatch is suppressed by write_ignore above).
+    if let Some(ref p) = canonical {
+        crate::commands::index_dispatch::dispatch_self_delete(&state, p).await;
+    }
     // #201 PR-A: tell the embed coordinator to tombstone this path's
     // vectors. Same reason as the synthetic file_changed event above —
     // write_ignore suppresses the watcher's natural Remove event so we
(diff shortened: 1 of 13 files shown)


src-tauri/src/commands/index_dispatch.rs
```diff
@@ -0,0 +1,144 @@
...
+// we touch ourselves is suppressed by `write_ignore` (D-12). Without these
+// helpers the in-memory LinkGraph, TagIndex, and Tantivy index would
+// silently trail the on-disk state until cold restart. Every self-mutating
+// command (write_file, merge_external_change on clean, delete_file,
+// rename_file, move_file, update_links_after_rename) routes through here.
+//
+// Best-effort contract: when the IndexCoordinator isn't attached (vault not
+// open / boot race) or the channel is full, we silently drop and let the
+// next cold-start rebuild repopulate. Disk-level failures stay with the
+// caller — we only dispatch after the caller has already succeeded on disk.
```
herox215: Doc error: this says dropped when "channel is full," but `tx.send(...).await` on a tokio bounded mpsc applies backpressure — it blocks until capacity frees. It only errors when the receiver is closed. As written, a saturated channel stalls every self-write path (auto-save, rename, merge) rather than silently dropping. Either correct the contract wording ("blocks under backpressure; drops only on receiver closed") or switch to `try_send` if drop-on-full is actually the intended semantic.

— Aristotle

src-tauri/src/tests/merge_external_change.rs
```diff
@@ -0,0 +1,707 @@
+// Issue #339: merge_external_change must be an authoritative write path —
+// on "clean" it writes the merged bytes to disk, records write_ignore,
+// and dispatches UpdateLinks / UpdateTags / Tantivy AddFile + Commit so
+// the in-memory indexes never trail the on-disk state.
+//
+// These tests drive a `_impl` mirror of the command body (see the note in
+// tests/files.rs:16-20 — tauri::State cannot be constructed outside a
+// running Tauri app). The mirror here reflects the CURRENT command body;
+// Phase 2 of the ticket replaces it with a call into a real `_impl`
+// helper in commands/vault.rs. Until then these tests are red.
```
herox215: Stale header: says "tests drive a `_impl` mirror... Phase 2 of the ticket replaces it... Until then these tests are red." Phase 2 is already merged in this PR — the tests call `merge_external_change_impl` directly (line 115). The comment will confuse anyone reading the file in six months. Delete it or replace with a one-liner describing what the suite covers.

— Aristotle

src-tauri/src/tests/merge_external_change.rs
```diff
@@ -0,0 +1,707 @@
...
+        let lg_guard = lg.lock().unwrap();
+        let fi_guard = fi.read().unwrap();
+        let a_backlinks = lg_guard.get_backlinks("A.md", &fi_guard);
+        // Backlinks list may still contain B.md (it points at the now-
+        // missing target). What matters is that nothing falsely claims
+        // the deleted file still has outgoing links.
+        // The critical assertion: outgoing_for(A) is None after dispatch.
+        drop(lg_guard);
+        drop(fi_guard);
+        let _ = a_backlinks; // touched for clarity
```
herox215: Test name promises two things it does not verify. `a_backlinks` is computed then discarded with `let _ = a_backlinks; // touched for clarity` — no assertion. Tags are never re-checked after delete (the `#shared` precondition on B.md has no matching postcondition, and A.md had no tags to begin with). The only load-bearing assertion is `outgoing_for("A.md").is_none()`. Either rename the test to match what it actually asserts (e.g. `delete_drops_outgoing_entry`) or add the missing assertions: Bs backlinks now unresolved, and `tags_for_file("A.md")` is empty.

— Aristotle

src-tauri/src/tests/merge_external_change.rs
```diff
@@ -0,0 +1,707 @@
...
+            );
+        }
+
+        // Drive delete via the command body (the tauri wrapper needs an
+        // AppHandle we can't synthesize cleanly). `delete_file_impl` + the
+        // explicit dispatch mirror what `delete_file` does, exactly as the
+        // files.rs `_impl` convention prescribes.
+        crate::commands::files::delete_file_impl(&state, a_abs.to_string_lossy().into_owned())
+            .expect("delete ok");
+        crate::commands::index_dispatch::dispatch_self_delete(&state, &a_abs).await;
```
herox215: Test bypasses the actual command under test. It calls `delete_file_impl` + `dispatch_self_delete` manually — if the real `delete_file` wrapper drops its `dispatch_self_delete` call (the exact regression this PR fixes), this test still passes. The regression guard is a test that drives the public command path end-to-end. Since `AppHandle` is hard to synthesize here, either extract an `_impl`-style helper from `delete_file` that encapsulates the dispatch sequence, or use `tauri::test::mock_app()` the same way `state_with_vault_and_coord` already does.

— Aristotle

src-tauri/src/commands/files.rs
```diff
@@ -534,12 +493,22 @@ pub async fn rename_file(
...
     if let Some(old) = old_canonical_for_tantivy {
-        dispatch_tantivy_delete(&state, old).await;
+        crate::commands::index_dispatch::dispatch_self_delete(&state, &old).await;
+    }
+    let new_canonical = std::fs::canonicalize(&result.new_path).ok();
+    if let Some(new_path) = new_canonical.as_ref() {
+        if let Ok(content) = std::fs::read_to_string(new_path) {
+            crate::commands::index_dispatch::dispatch_self_write(&state, new_path, &content).await;
+        }
     }
```
herox215: Silent swallow of the new-side dispatch. If `canonicalize` fails or `read_to_string` returns `Err`, the renamed file gets no UpdateLinks/UpdateTags/AddFile — the watcher is also suppressed by write_ignore, so the file silently stays stale until cold restart. This is the exact failure mode #339 fixes. At minimum `log::warn!` on the error so the drift is diagnosable; consider whether a failed post-rename read should bubble up as `Err(RenameResult)` — the file on disk now lives at `new_path` but the index believes it does not exist there.

Same issue repeats in `move_file` at lines 667-672.

— Aristotle

src-tauri/src/commands/vault.rs
```diff
@@ -401,12 +401,21 @@ pub(crate) fn format_iso8601_utc(epoch_secs: i64) -> String {
...
+/// backend did not write; disk still holds the external content and the
+/// caller keeps the local buffer (issue #339).
 #[derive(Serialize, Clone, Debug)]
 pub struct MergeCommandResult {
     /// "clean" or "conflict"
     pub outcome: String,
     /// The merged content (for "clean") or the original local content (for "conflict")
     pub merged_content: String,
+    /// SHA-256 hex of the merged bytes written to disk. `None` on conflict.
+    pub new_hash: Option<String>,
```
herox215: Two sources of truth for the same state. `outcome: "clean"` ↔ `new_hash: Some(_)` and `outcome: "conflict"` ↔ `new_hash: None` are coupled invariants encoded as independent fields. A future branch that sets `outcome = "clean"` but forgets `new_hash = Some(...)` (or vice-versa) will silently degrade the frontend to the `sha256Hex(merged_content)` fallback — a bug that will not surface until a post-merge autosave mysteriously re-enters the merge path. Model the enum cleanly:

```rust
#[derive(Serialize)]
#[serde(tag = "outcome", rename_all = "lowercase")]
pub enum MergeCommandResult {
    Clean { merged_content: String, new_hash: String },
    Conflict { merged_content: String },
}
```

The frontend adapter shrinks and the invariant is enforced by the compiler.

— Aristotle

src-tauri/src/commands/links.rs
```diff
@@ -464,6 +464,39 @@ pub async fn update_links_after_rename(
         }
     }
 
+    // #339: each rewritten source has new body text (`[[old]]` → `[[new]]`),
+    // so Tantivy fulltext hits for the old stem in these files are stale
+    // until re-indexing. The link graph is already updated inline above
+    // (#250 keeps that direct-mutation path for perf — routing through
+    // dispatch_self_write would regress the StemIndex optimization). Fire
+    // just the Tantivy side here. Tags are not affected — the rewrite
+    // only touches wiki-link text.
```
herox215: The justification comment here is good but the divergence is still a maintainability trap. `update_links_after_rename` does LinkGraph inline + Tantivy via channel, while every other self-write path goes through `dispatch_self_write` (which does all three via channel). A future reader will ask "why is this one special?" and the only answer is a StemIndex perf optimization for N-file rewrites that would otherwise run per-file.

Either:
1. Expose a `dispatch_self_write_tantivy_only` helper from `index_dispatch.rs` so this call site uses the same module (naming makes the divergence explicit), or
2. Teach `dispatch_self_write` to accept an optional pre-built `StemIndex` so the bulk path stays through the single dispatcher.

As written, the third person to touch this module will route the cascade through `dispatch_self_write` for "consistency" and regress #250.

— Aristotle

src-tauri/src/commands/files.rs
```diff
@@ -534,13 +493,24 @@ pub async fn rename_file(
...
-        dispatch_tantivy_delete(&state, old).await;
+        crate::commands::index_dispatch::dispatch_self_delete(&state, &old).await;
     }
+    // New-side read failure would leave the renamed file stale for
+    // fulltext + graphs until cold restart with no trace. Log loudly so
+    // operators can correlate a stale-index report with a rename that
+    // couldn't re-index. The disk rename has already succeeded, so we
+    // don't surface the error to the frontend — the user's operation
+    // completed, the index just needs recovery via the next rebuild.
+    dispatch_new_side_after_rename(&state, &result.new_path, "rename_file").await;
```
herox215: Same _impl-vs-wrapper footgun the delete fix closed is back here for rename/move. `rename_file_impl` performs the disk rename + FileIndex sync, but `dispatch_self_delete` for the old side and `dispatch_new_side_after_rename` live in the wrapper — so every `_impl`-driven test (see `tests/rename_link_resolution.rs`, `tests/files_ops.rs`) silently skips the LinkGraph / TagIndex / Tantivy sync, and a future refactor that drops either call from the wrapper passes CI. The delete comment-block at `files.rs:563-562` explicitly calls this out as a lesson learned; the same logic applies one screen up. Either move both dispatch calls inside `rename_file_impl` (returning the canonical new path the same way `delete_file_impl` now returns the canonical source), or add a direct `_impl`-level test that fails if the wrapper forgets to call them.

— Aristotle

src-tauri/src/commands/files.rs
```diff
@@ -549,24 +519,48 @@ pub async fn rename_file(
...
-    let _ = tx.send(IndexCmd::Commit).await;
+/// Dispatch UpdateLinks/UpdateTags/AddFile for the new side of a rename
+/// or move. Called after the disk rename + old-side dispatch succeeds.
+///
+/// On canonicalize or read failure we log::warn instead of bubbling: the
+/// disk operation already succeeded, so failing the command would be
+/// worse than leaving the index transiently stale until the next rebuild.
+/// The warning makes the degradation observable without being noisy.
+async fn dispatch_new_side_after_rename(state: &VaultState, new_path: &str, op: &str) {
+    match std::fs::canonicalize(new_path) {
```
herox215: `new_path` is already canonical — `rename_file_impl` returns `canonical_new.to_string_lossy()` at files.rs:410 and `move_file_impl` returns `dest` which is `canonical_to_folder.join(file_name)`. Re-canonicalizing here is a no-op syscall on the happy path and actively misleading: the error arm handles a case that cannot occur for callers that go through `*_impl`. Take `&Path` (or `PathBuf`) directly, drop the `canonicalize` branch, and collapse to a single `read_to_string` + dispatch.

— Aristotle

src-tauri/src/commands/links.rs
```diff
@@ -464,6 +464,39 @@ pub async fn update_links_after_rename(
...
+        for (source_rel, new_content) in &rewrites {
+            let abs_path = vault_root.join(source_rel);
+            let stem = abs_path
+                .file_stem()
+                .map(|s| s.to_string_lossy().into_owned())
+                .unwrap_or_default();
+            let title = crate::indexer::tantivy_index::extract_title(new_content, &stem);
+            let body = crate::indexer::parser::strip_markdown(new_content);
+            let hash = crate::hash::hash_bytes(new_content.as_bytes());
+            let _ = tx
```
herox215: The Tantivy-fragment of `dispatch_self_write` is copy-pasted here (title via `extract_title`, body via `strip_markdown`, hash via `hash_bytes`, then `AddFile` + `Commit`). If the title-extraction rule or the tokenization strategy changes, we now have two places to update and no test will catch the drift. Pull the inner loop body into a `pub(crate) async fn dispatch_tantivy_upsert(tx: &Sender<IndexCmd>, abs_path: PathBuf, content: &str)` in `index_dispatch.rs` and have `dispatch_self_write` call it too — one source of truth for the Tantivy document shape.

— Aristotle

Issue #80, opened 2026-04-15
Editor: deleting a line right after opening a note moves the entire content to the end


herox215:
## Summary

When opening a note and immediately deleting a line (before clicking/interacting with the editor any further), the full note content appears to be cut and re-inserted at the end of the document instead of the single line being removed.

## Steps to reproduce

1. Open Vaultcore with an existing vault.
2. Click a note in the file tree to open it in a tab.
3. Without clicking into the editor body or moving the cursor, trigger a line delete (e.g. `Cmd+Shift+K` / whatever line-delete shortcut is bound, or select a line and press Backspace).
4. Observe the document contents.

## Expected

Just the targeted line is removed; the rest of the document stays in place.

## Actual

The entire note content appears to be removed from its original position and re-appended at the end, leaving the document scrambled.

## Notes / suspected causes

- May be related to editor state initialisation — cursor/selection might default to a range that spans the whole document before the first user interaction sets it.
- Could also be an interaction between the tab-open flow and CodeMirror's initial selection / dispatch of the first edit.
- Needs confirmation of:
  - which key combination / action triggers it most reliably
  - whether it reproduces on every note or only under certain conditions (e.g. empty selection vs. multi-line selection, reading mode vs edit mode, first open after launch vs subsequent opens)

## Out of Scope

Fixing unrelated editor-init issues surfaced during investigation.

herox215:
Root cause (nach Reproduktion): zwei gekoppelte Bugs im Auto-Save-Pfad.

1. `editorStore.lastSavedHash` war ein globaler Single-Value-Store, nicht per-Tab. Nach dem Speichern in Tab A und Wechsel zu Tab B überschrieb der nächste Save in B den Hash — beim Zurückwechsel zu A stand der Hash von B im Store.

2. Der Tab-Switch-Handler in `EditorPane.syncFromTab` hat obendrein `String(activeTab.lastSaved)` — einen `Date.now()`-Timestamp — in den Hash-Slot gestopft, statt des echten Content-Hashes.

[...]

PR #344, opened 2026-04-23
refactor(tabStore): split into lifecycle + layout + reload facades (#341)

### LINKED ISSUE #341: refactor(tabStore): split monolithic tab store into lifecycle + layout + reload concerns
## Problem

`src/store/tabStore.ts` has grown to ~16.6 KB and mixes three distinct responsibilities:

1. **Tab lifecycle** — open, close, replace, active tab tracking
2. **Tab layout** — ordering, reordering, split/stacking state
3. **Tab reload / disk-sync** — reacting to external file changes, merge prompts, coupling with `tabReloadStore`

The file is hard to reason about, hard to test in isolation (single 12.8 KB `tabStore.test.ts`), and changes to one concern regularly risk regressions in another.

## Affected code

- `src/store/tabStore.ts`
- `src/store/tabReloadStore.ts` (cross-store coupling)
- `src/store/tabStore.test.ts` (will need to split alongside)
- Components consuming tabStore (TabBar, Tab, VaultLayout, editor wiring)

## Proposed direction

Decompose into three stores with clear contracts, following the pattern used by `searchStore` / `vaultStore`:

- `tabLifecycleStore.ts` — open/close/replace/active
- `tabLayoutStore.ts` — order, reorder, split state
- `tabReloadStore.ts` — absorb the disk-sync logic currently in tabStore

Keep a thin `tabStore.ts` re-export facade during migration to avoid a big-bang callsite change; remove the facade once callsites are migrated.

## Non-goals

- No new features
- No change to public behavior / user-visible semantics
- No cross-store architecture overhaul (that is a separate, bigger discussion)

## Boy Scout

- While in this area, check whether manual `subscribe` listeners between vaultStore/tabStore can be replaced with Svelte 5 runes or a typed dependency contract
- Note (don't fix here) any untyped cross-store updates for a follow-up ticket

## Acceptance

- [ ] `tabStore.ts` is under ~4 KB facade (or removed entirely)
- [ ] Three focused stores each with their own test file
- [ ] No behavioral regression — existing E2E + Vitest specs pass
- [ ] No new public API surface beyond what consumers already use

## Risk

Medium. Tab handling is central; regressions are user-visible. Needs careful test coverage before refactor.

herox215:
Closes #341.

## Summary

`src/store/tabStore.ts` (532 lines) mixed three concerns — tab identity,
pane layout, and disk-sync — in one writable. The file was hard to
reason about, hard to test in isolation, and regressions in one concern
routinely slipped through review because of cross-concern coupling.

This PR decomposes the store into a **shared private writable + three
scoped facades**, keeps a thin `tabStore.ts` shim for the 47 existing
consumers, and folds in two small Boy Scout cleanups.

## Architecture

```
tabStoreCore.ts       ← private `writable<TabStoreState>` + types
  ├─ tabLifecycleStore — open/close/activate/cycle + per-tab metadata
  ├─ tabLayoutStore    — moveToPane + reorderPane
  └─ tabReloadStore    — reload signal + per-tab save snapshot
tabStore.ts           ← compatibility shim, re-exports the full surface
```

The three facades share the single writable in `tabStoreCore`. Every
cross-concern method (e.g. `closeTab` touches tabs + splitState +
activeTabId) performs **exactly one** `_core.update(...)` — so
subscribers never observe a torn intermediate state. The atomicity
invariant is enforced by emission-counter tests in `tabStore.test.ts`.

Decision rule (documented at the top of `tabStoreCore.ts`):
- **Lifecycle** — tab identity + per-tab metadata
- **Layout** — pane arrangement
- **Reload** — disk-sync signals + post-save snapshot state

## Why single core instead of three separate stores?

Splitting the state across three writables would force cross-concern
methods (`closeTab`, `openTab`, `moveToPane`) to emit 2-3 times,
producing observable torn states (e.g. a subscriber could see `tabs`
updated before `splitState.left`, or the new `activeTabId` pointing at
a tab that isn't yet in any pane). That's the exact class of race bug
this refactor is trying to eliminate.

## What didn't change

- Zero behavior change for any consumer.
- `tabStore.ts` keeps its full legacy method surface via re-exports.
- Existing 47 consumers stay on the old import path.
- The `_reorderPane` alias is preserved for TabBar's drag handler;
  new code should use `tabLayoutStore.reorderPane`.

## Boy Scout

Folded in (second commit, clean separation for review):
- `EditorPane.handleWikiLinkClick` — replaced subscribe/unsub
  snapshot with `get(vaultStore).currentPath`.
- `EditorPane.unsubTabReload` — same rewrite.
- `VaultLayout.closeActiveTab` — symmetry fix using `getActiveTab()`.
- `closeByPath` / `getActiveTab` on the lifecycle facade now use
  `get()` (eliminates the dead `_store.update(() => state)` no-op
  the legacy `closeByPath` previously fired).

Noted for follow-up (not fixed here — out of scope):
- `GraphView.svelte` has three separate `tabStore.subscribe` blocks
  that only react to identity changes — candidates for a derived
  slice after this lands.
- `updateFilePath` takes untyped strings; branded `AbsPath` /
  `VaultRelPath` would be cleaner but is a separate discussion.
- An ESLint `no-restricted-imports` rule steering new code away
  from `./tabStore` would make the shim a soft-deprecation; defer.

## Test plan

- [x] `npx vitest run src/store/` — 53/53 pass
- [x] `npx vitest run` across store/components/lib — 1087/1088 pass
      (one flaky perf budget test unrelated to this refactor: 502ms
      vs. 500ms jsdom bound in `largeFile.test.ts`, passes on rerun)
- [x] `npx tsc --noEmit` — clean
- [ ] UAT — open/close/cycle tabs, split/merge panes, reload after
      external rename, toggle reading mode, verify no visible
      regressions at 100k-note vault

### Pin-before-refactor tests added

Nails current behavior before decomposing:
- `closeTab` of active tab in right pane merges into left (mirror of
  the pre-existing right-closed test).
- `closeTab` of only tab in left pane collapses split.
- `reorderPane` (previously untested despite being live via TabBar).
- `setLastSavedHash(id, null)` is distinguishable from never-set (#80).
- `tabReloadStore.request` emits a fresh token on repeat calls.

### Atomicity regression net

Asserts **exactly one emission** on the shared core for every cross-
concern op: `openTab`, `closeTab` (with merge-collapse), `moveToPane`,
`activateTab`, `closeByPath` (and 0 emissions for `closeByPath` on a
miss). This is the tight guard against anyone accidentally splitting
a cross-concern method into multiple `_core.update` calls.

## Files

Added:
- `src/store/tabStoreCore.ts`
- `src/store/tabLifecycleStore.ts`
- `src/store/tabLifecycleStore.test.ts`
- `src/store/tabLayoutStore.ts`
- `src/store/tabLayoutStore.test.ts`
- `src/store/tabReloadStore.test.ts`

Modified:
- `src/store/tabStore.ts` — now a ~55-line shim
- `src/store/tabStore.test.ts` — narrowed to cross-concern + atomicity
- `src/store/tabReloadStore.ts` — grew two methods + `_reset` helper
- `src/components/Editor/EditorPane.svelte` — Boy Scout `get()` rewrites
- `src/components/Layout/VaultLayout.svelte` — Boy Scout symmetry fix

### DIFF
--- a/src/store/tabLayoutStore.test.ts
+++ b/src/store/tabLayoutStore.test.ts
@@ -33,13 +33,25 @@ describe("tabLayoutStore", () => {
       expect(state.splitState.right).toEqual([]);
     });
 
-    it("moving from source pane that becomes empty closes the split", () => {
+    it("moving right-pane's last tab back to left collapses the split", () => {
       const id1 = tabLifecycleStore.openTab("/vault/a.md");
       tabLayoutStore.moveToPane("right");
       tabLayoutStore.moveToPane("left");
       const state = get(tabLayoutStore);
       expect(state.splitState.right).toEqual([]);
       expect(state.splitState.left).toContain(id1);
+    });
+
+    it("moving left-pane's last tab to right also collapses into left (not right)", () => {
+      // Pins counter-intuitive contract: the collapse branch always dumps
+      // surviving tabs into `left` regardless of which pane was the target.
+      // Matches pre-#341 behavior — see comment in tabLayoutStore.moveToPane.
+      const id1 = tabLifecycleStore.openTab("/vault/a.md");
+      tabLayoutStore.moveToPane("right");
+      const state = get(tabLayoutStore);
+      expect(state.splitState.right).toEqual([]);
+      expect(state.splitState.left).toEqual([id1]);
+      expect(state.splitState.activePane).toBe("left");
     });
   });
 
@@ -48,15 +60,20 @@ describe("tabLayoutStore", () => {
       const id1 = tabLifecycleStore.openTab("/vault/a.md");
       const id2 = tabLifecycleStore.openTab("/vault/b.md");
       const id3 = tabLifecycleStore.openTab("/vault/c.md");
+      // Activate a tab that is NOT first in the target permutation so the
+      // "active stays put" assertion cannot coincide with the "incidentally
+      // first" tab.
+      tabLifecycleStore.activateTab(id2);
       const before = get(tabLayoutStore);
       expect(before.splitState.left).toEqual([id1, id2, id3]);
+      expect(before.activeTabId).toBe(id2);
 
       tabLayoutStore.reorderPane("left", [id3, id1, id2]);
 
       const after = get(tabLayoutStore);
       expect(after.splitState.left).toEqual([id3, id1, id2]);
       expect(after.tabs).toHaveLength(3);
-      expect(after.activeTabId).toBe(id3);
+      expect(after.activeTabId).toBe(id2); // unchanged despite id3 now first
       expect(after.splitState.right).toEqual([]);
     });
 
--- a/src/store/tabLayoutStore.test.ts
+++ b/src/store/tabLayoutStore.test.ts
@@ -1,0 +1,79 @@
+// tabLayoutStore tests — covers moveToPane (split creation + collapse)
+// and reorderPane. Cross-concern behavior (e.g. closeTab spanning lifecycle
+// and layout) lives in tabStore.test.ts.
+
+import { describe, it, expect, beforeEach } from "vitest";
+import { get } from "svelte/store";
+import { tabLifecycleStore } from "./tabLifecycleStore";
+import { tabLayoutStore } from "./tabLayoutStore";
+import { _reset } from "./tabStoreCore";
+
+beforeEach(() => {
+  _reset();
+});
+
+describe("tabLayoutStore", () => {
+  describe("moveToPane", () => {
+    it("moves active tab to right pane, creating split if empty", () => {
+      const id1 = tabLifecycleStore.openTab("/vault/a.md");
+      const id2 = tabLifecycleStore.openTab("/vault/b.md");
+      tabLifecycleStore.activateTab(id2);
+      tabLayoutStore.moveToPane("right");
+      const state = get(tabLayoutStore);
+      expect(state.splitState.left).toContain(id1);
+      expect(state.splitState.left).not.toContain(id2);
+      expect(state.splitState.right).toContain(id2);
+    });
+
+    it("moving to current pane does nothing", () => {
+      const id1 = tabLifecycleStore.openTab("/vault/a.md");
+      tabLayoutStore.moveToPane("left");
+      const state = get(tabLayoutStore);
+      expect(state.splitState.left).toContain(id1);
+      expect(state.splitState.right).toEqual([]);
+    });
+
+    it("moving from source pane that becomes empty closes the split", () => {
+      const id1 = tabLifecycleStore.openTab("/vault/a.md");
+      tabLayoutStore.moveToPane("right");
+      tabLayoutStore.moveToPane("left");
+      const state = get(tabLayoutStore);
+      expect(state.splitState.right).toEqual([]);
+      expect(state.splitState.left).toContain(id1);
+    });
+  });
+
+  describe("reorderPane", () => {
+    it("replaces pane tab-id order without touching tabs list or active", () => {
+      const id1 = tabLifecycleStore.openTab("/vault/a.md");
+      const id2 = tabLifecycleStore.openTab("/vault/b.md");
+      const id3 = tabLifecycleStore.openTab("/vault/c.md");
+      const before = get(tabLayoutStore);
+      expect(before.splitState.left).toEqual([id1, id2, id3]);
+
+      tabLayoutStore.reorderPane("left", [id3, id1, id2]);
+
+      const after = get(tabLayoutStore);
+      expect(after.splitState.left).toEqual([id3, id1, id2]);
+      expect(after.tabs).toHaveLength(3);
+      expect(after.activeTabId).toBe(id3);
+      expect(after.splitState.right).toEqual([]);
+    });
+
+    it("reorders right pane independently", () => {
+      const id1 = tabLifecycleStore.openTab("/vault/a.md");
+      const id2 = tabLifecycleStore.openTab("/vault/b.md");
+      tabLifecycleStore.activateTab(id2);
+      tabLayoutStore.moveToPane("right");
+      const id3 = tabLifecycleStore.openTab("/vault/c.md");
+      tabLayoutStore.moveToPane("right");
+      const before = get(tabLayoutStore);
+      expect(before.splitState.right).toEqual([id2, id3]);
+
+      tabLayoutStore.reorderPane("right", [id3, id2]);
+
+      expect(get(tabLayoutStore).splitState.right).toEqual([id3, id2]);
+      expect(get(tabLayoutStore).splitState.left).toEqual([id1]);
+    });
+  });
+});
--- a/src/store/tabLayoutStore.ts
+++ b/src/store/tabLayoutStore.ts
@@ -49,12 +49,30 @@ export const tabLayoutStore = {
 
   /**
    * Reorder the tab IDs in a specific pane (used by drag-to-reorder in TabBar).
-   * Callers are responsible for passing a permutation of the current pane IDs.
+   * `newIds` MUST be a permutation of the current pane IDs — anything else
+   * breaks the `splitState ⊆ tabs` invariant. Validated in dev builds so a
+   * caller bug fails loudly instead of producing torn state.
    */
   reorderPane(pane: "left" | "right", newIds: string[]): void {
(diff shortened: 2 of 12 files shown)


src/store/tabReloadStore.ts
```diff
@@ -26,14 +38,43 @@ interface TabReloadState {
   pending: TabReloadRequest | null;
 }
 
-const _store = writable<TabReloadState>({ pending: null });
+const _signal = writable<TabReloadState>({ pending: null });
 
 export const tabReloadStore = {
-  subscribe: _store.subscribe,
+  subscribe: _signal.subscribe,
```
herox215, line 36: `tabReloadStore.subscribe` is wired to `_signal`, but `setLastSavedContent` / `setLastSavedHash` mutate `_core` — subscribers of this facade will never observe the hash/content they just wrote. Either move the per-tab save-snapshot setters to `tabLifecycleStore` (they are per-tab metadata, same shape as `setDirty`/`setViewMode`), or expose a `_core`-backed subscribe on this facade. The current split — writers on one writable, subscribe on another — is an invariant this facade cannot keep.

— Aristotle

src/store/tabStoreCore.ts
```diff
@@ -0,0 +1,125 @@
+// tabStoreCore — shared writable backing the three tab facades (#341).
+//
+// Decision rule for where a new method belongs:
+//   • tabLifecycleStore — tab identity (open/close/activate/cycle) and
+//     per-tab metadata (dirty flag, view mode, scroll positions, rename).
+//   • tabLayoutStore    — pane arrangement (moveToPane, reorderPane).
+//   • tabReloadStore    — disk-sync signals and post-save snapshot state
```
herox215: The decision rule puts `lastSavedContent` / `lastSavedHash` in `tabReloadStore` but they are per-tab metadata (same shape as `isDirty`, `viewMode`, `scrollPos`) that `tabLifecycleStore` already owns. A future dev reading the `Tab` interface will naturally look in lifecycle for these setters and not find them. Splitting by causal origin (save-pipeline vs user action) instead of by data shape will keep biting. Suggest: move the two save-snapshot setters to lifecycle, keep only the one-shot reload signal in `tabReloadStore`, and rewrite this block to reflect that.

— Aristotle

src/store/tabStore.ts
```diff
@@ -1,532 +1,60 @@
...
+
+  // Layout
+  moveToPane: tabLayoutStore.moveToPane,
+  /** Preserved alias — new callers: use tabLayoutStore.reorderPane. */
+  _reorderPane: tabLayoutStore.reorderPane,
+
+  // Reload / save snapshot
+  setLastSavedContent: tabReloadStore.setLastSavedContent,
+  setLastSavedHash: tabReloadStore.setLastSavedHash,
+
```
herox215: `tabStore._reset()` calls `tabStoreCore._reset` which resets `_core` but leaves `_signal` in `tabReloadStore` untouched. Any test that relies on `tabStore._reset()` as a full reset and then subscribes to `tabReloadStore` inherits a leaked pending token from the previous test. Make `_reset` here also clear the reload signal (import `tabReloadStore` and call `tabReloadStore._reset()`), or rename to `_resetCore` so the limitation is visible.

— Aristotle

src/store/tabLifecycleStore.test.ts
```diff
@@ -0,0 +1,302 @@
...
+      const id1 = tabLifecycleStore.openTab("/vault/a.md");
+      const id2 = tabLifecycleStore.openTab("/vault/b.md");
+      tabLifecycleStore.activateTab(id2);
+      // move id2 to right via the tabStore shim would be nicer but this keeps
+      // the test within lifecycle; use the core through a direct move-to-pane
+      // integration in tabStore.test.ts instead. Here, just confirm identity.
+      tabLifecycleStore.activateTab(id1);
+      const state = get(tabLifecycleStore);
+      expect(state.activeTabId).toBe(id1);
+      expect(state.splitState.activePane).toBe("left");
```
herox215, line 232: This test never exercises the pane-switching branch of `activateTab` — both `id1` and `id2` live in the left pane, so `activePane` was already `"left"`. The contract this method documents ("updating activePane to match the pane containing it") is only verified cross-concern in `tabStore.test.ts`, which defeats the purpose of a lifecycle-scoped test for `activateTab`. Either drop the test here as noise or use `_core` directly to seed a tab in the right pane and assert `activePane` flips to `"right"`.

— Aristotle

src/store/tabLayoutStore.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      const id3 = tabLifecycleStore.openTab("/vault/c.md");
+      const before = get(tabLayoutStore);
+      expect(before.splitState.left).toEqual([id1, id2, id3]);
+
+      tabLayoutStore.reorderPane("left", [id3, id1, id2]);
+
+      const after = get(tabLayoutStore);
+      expect(after.splitState.left).toEqual([id3, id1, id2]);
+      expect(after.tabs).toHaveLength(3);
+      expect(after.activeTabId).toBe(id3);
```
herox215: `activeTabId` was already `id3` before `reorderPane` (last-opened-tab wins), so this assertion reads an incidental pre-existing value, not a contract. Reword the test to activate a tab other than the one that ends up first in `newIds` and assert active is still that tab after reorder — that is the "without touching active" invariant you actually want to lock in.

— Aristotle

src/store/tabLayoutStore.ts
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Reorder the tab IDs in a specific pane (used by drag-to-reorder in TabBar).
+   * Callers are responsible for passing a permutation of the current pane IDs.
+   */
+  reorderPane(pane: "left" | "right", newIds: string[]): void {
+    _core.update((state) => ({
+      ...state,
+      splitState: { ...state.splitState, [pane]: newIds },
+    }));
+  },
```
herox215, line 77: The `newIds` contract ("a permutation of current pane IDs") is caller-responsibility but never checked, so a caller bug silently breaks the invariant that `splitState.left ∪ splitState.right` ⊆ `tabs`. Since we are already touching this code, a dev-only assertion (same length + same set as `state.splitState[pane]`) is cheap and prevents torn state from ever landing in the store. Boy-Scout candidate.

— Aristotle

src/store/tabStore.ts
```diff
@@ -1,532 +1,60 @@
...
+  setViewMode: tabLifecycleStore.setViewMode,
+  toggleViewMode: tabLifecycleStore.toggleViewMode,
+  updateScrollPos: tabLifecycleStore.updateScrollPos,
+  updateReadingScrollPos: tabLifecycleStore.updateReadingScrollPos,
+  updateFilePath: tabLifecycleStore.updateFilePath,
+
+  // Layout
+  moveToPane: tabLayoutStore.moveToPane,
+  /** Preserved alias — new callers: use tabLayoutStore.reorderPane. */
+  _reorderPane: tabLayoutStore.reorderPane,
```
herox215, line 56: TabBar.svelte (the only consumer of `_reorderPane`) still calls the shim alias while the PR documents the new canonical name as `tabLayoutStore.reorderPane`. A single consumer, trivial swap — migrating it here kills the leading-underscore "private" label and lets the alias be deleted in a follow-up without another cross-file sweep. Boy-Scout candidate.

— Aristotle

src/store/tabLayoutStore.ts
```diff
@@ -0,0 +1,60 @@
...
+      const sourcePane = whichPane(state, activeTabId);
+      if (sourcePane === null) return state;
+      if (sourcePane === targetPane) return state;
+
+      const newSourceIds = state.splitState[sourcePane].filter((id) => id !== activeTabId);
+      const newTargetIds = [...state.splitState[targetPane], activeTabId];
+
+      let newSplitState: SplitState;
+      if (newSourceIds.length === 0) {
+        // Source pane empty — collapse split into left pane regardless of
```
herox215, line 30: The collapse branch always dumps surviving tabs into `left` regardless of which pane was the target — moving the only tab on the left to `right` ends up on the left. This matches pre-refactor behavior and the comment above acknowledges it, but the layout-store test only exercises `right → left` (`moveToPane("right")` then `moveToPane("left")`). Add one case for `left → right` collapsing back to left, so this counter-intuitive contract is pinned and cannot silently change later.

— Aristotle

Issue #345, opened 2026-04-23
feat: password-protected encrypted folders


herox215:
## Problem

VaultCore stores notes as plain Markdown on disk. There is currently no way to keep a subset of notes private behind a password. Users with sensitive content (journals, credentials, client work) have no in-app option short of locking the whole vault.

## Proposed feature

Allow a user to designate a folder as **encrypted**. Behavior:

- User picks a folder in the sidebar → *Encrypt folder…* → prompts for a password.
- All files inside that folder (recursively) are encrypted on disk, in place, within the vault directory. Files remain in the vault but their contents are ciphertext.
- Once encrypted and locked, the folder's contents are **inaccessible** from the app:
  - Not indexed by full-text search
  - Not traversed for link resolution / backlinks / graph
  - Not shown in quick switcher
  - Editor cannot open any file inside
- Unlocking requires the password. Unlocked state decrypts lazily for reads/edits and re-encrypts on save + on lock/app-close.
- Relocking: explicit action, on app quit, and on timeout (configurable).

## Out of scope for this ticket

- **Choice of crypto primitives, key derivation, file format, salt/nonce handling** — TBD at implementation time. Must be reviewed against VaultCore's zero-network / offline-only constraint and Obsidian-vault compatibility before coding.
- Whether encrypted folders are interoperable with Obsidian (likely not — document the tradeoff).
- Multi-password / per-file encryption — not in this ticket.
- Recovery keys, password reset — TBD, likely a separate ticket.

## Affected areas (rough)

- Rust backend: new crypto module; FS read/write paths must route through it for files under encrypted roots.
- Indexer (Tantivy): skip encrypted roots when locked; reindex on unlock.
- Link/tag graph: must treat locked subtrees as opaque.
- Sidebar / file tree: visual state for locked vs unlocked folders.
- Settings: per-folder encryption state, auto-lock timeout.
- Crash recovery: encrypted buffers must not leak plaintext through the ≤ 2 s loss window.

## Open design questions (to resolve with Vitruvius + user before implementation)

- Visual treatment of locked folders in the sidebar (icon? muted? collapsed and disabled?).
- Unlock UX: inline prompt in sidebar, modal, or command-palette driven?
- What happens if the user tries to open a file inside a locked folder from a link/backlink elsewhere?
- Auto-lock timeout default and how it's communicated.
- Password strength requirements and how they're surfaced.
- Status indicator for "currently unlocked" state so the user always knows.

## Acceptance (high level — to be refined)

[...]

PR #347, opened 2026-04-23
feat(encryption): crypto primitives + state scaffolding (#345 · 1a)

herox215:
First slice of #345 (password-protected encrypted folders). Lands the crypto primitives, on-disk container format, per-vault manifest, locked-path registry + key cache, and wires VaultState — **with zero user-facing surface yet**. IPC commands, FS/indexer/watcher gating, and UI follow in part 1b, 2, and 3 so each slice stays reviewable.

## Scope (this PR)

- \`encryption::crypto\` — XChaCha20-Poly1305 + Argon2id (m=64 MiB, t=3, p=1). Params are frozen — a regression test pins them because silently changing any invalidates every existing vault.
- \`encryption::file_format\` — VCE1 container (\`magic[4] ++ nonce[24] ++ ciphertext++tag\`). \`parse()\` distinguishes truncation / wrong magic from AEAD failure so UX can say \"corrupt file\" vs \"wrong password\". \`write_atomic\` does temp+rename per file — not a WAL, just \`std::fs::rename\` — so a crash leaves each file fully plain or fully ciphertext.
- \`encryption::manifest\` — per-vault \`.vaultcore/encrypted-folders.json\`, matching the bookmarks/snippets sidecar convention. Survives vault moves. Salt is base64. \`FolderState\` tracks \`encrypting|encrypted\` for the crash-resume flow in PR 345.3.
- \`encryption::registry\` — \`LockedPathRegistry\` (RwLock<HashSet>) + \`Keyring\` (Mutex<HashMap<\_, Zeroizing<[u8;32]>>>). \`is_locked()\` uses \`path.ancestors()\` so \`/vault/secretplans\` is not treated as locked when \`/vault/secret\` is; \`/vault/secret/sub/deep\` is.
- \`encryption::batch\` — walker, \`encrypt_file_in_place\` (idempotent over already-sealed files), sentinel write/verify for password probing.

Errors: \`PathLocked\`, \`WrongPassword\`, \`CryptoError\` on \`VaultError\`.

## Out of scope — follow-ups

- **1b**: IPC commands (\`encrypt_folder\`, \`unlock_folder\`, \`lock_folder\`, \`lock_all_folders\`, \`list_encrypted_folders\`); FS gating (\`ensure_inside_vault\`, \`save_attachment\`); indexer/link-graph/watcher/FileIndex/tree gating; Rust integration tests per seam.
- **2**: Frontend \`encryptedFoldersStore\`, \`EncryptFolderModal\` + \`PasswordPromptModal\`, sidebar lock/unlock icons, SettingsModal SECURITY section, command palette entries, Vitest.
- **3**: Auto-lock timer, link-click-into-locked flow, crash-resume flow, graph edge filtering, **WDIO E2E spec**.

## Obsidian compat

Ciphertext \`.md\` files are not readable by Obsidian — but the surrounding vault stays compatible. Users opt in per folder. Aligned with the \"must not corrupt existing Obsidian vaults\" constraint in CLAUDE.md: the vault structure is untouched; only files inside an explicitly-encrypted folder become ciphertext.

## Test plan

- [x] 28 unit tests — all green (\`cargo test --lib encryption::\`)
- [x] Fixture-based ciphertext regression test (\`crypto::tests::fixture_ciphertext_decrypts\`) — guards against silent breakage on \`chacha20poly1305\` / \`argon2\` library bumps
- [x] \`cargo check --lib\` clean
- [x] No runtime surface exposed — existing behavior unchanged (every file read/write is still unguarded by \`locked_paths\` until 1b)

## Review notes

- Lock ordering is documented in \`registry.rs\`: \`current_vault → index_coordinator → locked_paths\`. File index is independent; keyring is never held with locked_paths.
- \`is_locked()\` fails closed on a poisoned RwLock — intentional.
- \`Keyring\` stores \`Zeroizing<[u8; 32]>\` so a drop wipes key bytes; no \`mlock\` / \`mprotect\` in MVP (out of scope per #345 body).

Closes part of #345.

### DIFF
--- a/src-tauri/src/encryption/batch.rs
+++ b/src-tauri/src/encryption/batch.rs
@@ -15,15 +15,27 @@ use crate::encryption::file_format::{frame, parse, write_atomic, MAGIC};
 use crate::encryption::{SENTINEL_FILENAME, SENTINEL_PLAINTEXT};
 use crate::error::VaultError;
 
-/// Iterate `.md` files strictly inside `root`. Used by the encrypt/unlock
-/// batch operations. Does not recurse into dot-prefixed dirs (e.g. the
-/// sidecar `.vaultcore/`) and skips the sentinel.
-pub fn walk_md_under(root: &Path) -> impl Iterator<Item = PathBuf> + '_ {
+/// Iterate every regular file strictly inside `root`. Used by the
+/// encrypt/unlock batch operations.
+///
+/// An encrypted folder must seal every file it contains, not just the
+/// `.md` notes — otherwise attachments (images pasted into notes, PDFs,
+/// excalidraw canvases, CSV exports) would stay plaintext with no
+/// warning, breaking the "folder is private" contract.
+///
+/// Skips:
+/// - dot-prefixed directories (`.vaultcore/` sidecar, `.trash/`, `.git/`)
+/// - dot-prefixed files (the sentinel `.vaultcore-folder-key-check`,
+///   any hidden platform metadata)
+/// - symlinks (`follow_links=false`), so a symlink into a locked folder
+///   cannot be used to exfiltrate plaintext through the walk.
+pub fn walk_all_under(root: &Path) -> impl Iterator<Item = PathBuf> + '_ {
     WalkDir::new(root)
         .follow_links(false)
         .into_iter()
         .filter_entry(|e| {
-            // Skip hidden dirs (e.g. .vaultcore, .trash, .git).
+            // Skip hidden dirs (e.g. .vaultcore, .trash, .git) and hidden
+            // files (e.g. the sentinel we write ourselves).
             e.depth() == 0
                 || !e
                     .file_name()
@@ -33,11 +45,6 @@ pub fn walk_md_under(root: &Path) -> impl Iterator<Item = PathBuf> + '_ {
         .filter_map(Result::ok)
         .filter(|e| e.file_type().is_file())
         .map(|e| e.into_path())
-        .filter(|p| {
-            p.extension()
-                .map(|ext| ext.eq_ignore_ascii_case("md"))
-                .unwrap_or(false)
-        })
 }
 
 /// Encrypt one plaintext file in place. Reads, seals, writes atomically.
@@ -85,7 +92,10 @@ pub fn write_sentinel(root: &Path, key: &[u8; KEY_LEN]) -> Result<(), VaultError
 }
 
 /// Returns `Ok(true)` if the sentinel decrypts cleanly, `Err(WrongPassword)`
-/// if AEAD tag fails, and an IO/crypto error for missing/corrupt files.
+/// on AEAD tag failure (remapped from the generic `CryptoError` that
+/// `decrypt_bytes` produces — this is the sentinel-probe layer, so a tag
+/// mismatch means the password was wrong), and `CryptoError` for missing
+/// or structurally-broken container files.
 pub fn verify_sentinel(root: &Path, key: &[u8; KEY_LEN]) -> Result<bool, VaultError> {
     let path = root.join(SENTINEL_FILENAME);
     let bytes = fs::read(&path).map_err(|e| match e.kind() {
@@ -98,8 +108,14 @@ pub fn verify_sentinel(root: &Path, key: &[u8; KEY_LEN]) -> Result<bool, VaultEr
         _ => VaultError::Io(e),
     })?;
     let (nonce, body) = parse(&bytes)?;
-    let pt = decrypt_bytes(key, &nonce, body)?;
-    Ok(pt == SENTINEL_PLAINTEXT)
+    match decrypt_bytes(key, &nonce, body) {
+        Ok(pt) => Ok(pt == SENTINEL_PLAINTEXT),
+        // Remap AEAD failure → WrongPassword because at the sentinel
+        // layer the caller (unlock flow) is probing a candidate key.
+        // A tag mismatch is always "bad password" here.
+        Err(VaultError::CryptoError { .. }) => Err(VaultError::WrongPassword),
+        Err(e) => Err(e),
+    }
 }
 
 #[cfg(test)]
@@ -120,15 +136,22 @@ mod tests {
     }
 
     #[test]
-    fn walk_md_under_skips_non_md_and_hidden() {
+    fn walk_all_under_covers_attachments_skips_hidden() {
         let (_g, root) = sample_tree();
-        let mut files: Vec<_> = walk_md_under(&root)
+        let mut files: Vec<_> = walk_all_under(&root)
             .map(|p| p.strip_prefix(&root).unwrap().to_path_buf())
             .collect();
         files.sort();
+        // `note.txt` and `sub/b.md` must be in scope; hidden
+        // `.secret.md` must not. The walker is deliberately extension-
+        // agnostic so attachments don't fall out of encryption scope.
         assert_eq!(
             files,
-            vec![PathBuf::from("a.md"), PathBuf::from("sub/b.md")]
+            vec![
+                PathBuf::from("a.md"),
+                PathBuf::from("note.txt"),
+                PathBuf::from("sub/b.md"),
+            ]
         );
     }
 
@@ -136,7 +159,7 @@ mod tests {
     fn encrypt_then_decrypt_roundtrip_on_disk() {
         let (_g, root) = sample_tree();
         let key = derive_key(b"pw", b"saltsaltsaltsalt").unwrap();
-        for p in walk_md_under(&root) {
+        for p in walk_all_under(&root) {
             encrypt_file_in_place(&key, &p).unwrap();
         }
         // Each file now starts with VCE1 magic.
--- a/src-tauri/src/encryption/batch.rs
+++ b/src-tauri/src/encryption/batch.rs
@@ -1,0 +1,180 @@
+// Folder-level encrypt / lock / unlock primitives. Pure data
+// operations — the IPC layer in `crate::commands::encryption` is
+// responsible for state mutation (registry, keyring, manifest) and for
+// driving progress events.
+
+use std::fs;
+use std::path::{Path, PathBuf};
+
+use walkdir::WalkDir;
+
+use crate::encryption::crypto::{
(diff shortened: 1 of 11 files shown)


src-tauri/src/encryption/file_format.rs
```diff
@@ -0,0 +1,157 @@
...
+    })();
+    if let Err(e) = write_result {
+        let _ = fs::remove_file(&tmp);
+        return Err(VaultError::Io(e));
+    }
+
+    if let Err(e) = fs::rename(&tmp, target) {
+        let _ = fs::remove_file(&tmp);
+        return Err(VaultError::Io(e));
+    }
```
herox215, line 95: `write_atomic` syncs the tempfile but never fsyncs the parent directory after `fs::rename`. The PR description claims "a crash leaves each file fully plain or fully ciphertext" — on POSIX that invariant needs a parent-dir fsync to survive a power loss, otherwise the rename can be lost and you land on "tmp gone, target stale/missing". Open the parent dir and `sync_all()` it after rename, or drop the crash-atomicity claim.

— Aristotle

src-tauri/src/encryption/crypto.rs
```diff
@@ -0,0 +1,192 @@
...
+/// distinguish; both cases are "key does not decrypt this blob".
+pub fn decrypt_bytes(
+    key: &[u8; KEY_LEN],
+    nonce: &[u8; NONCE_LEN],
+    ciphertext_and_tag: &[u8],
+) -> Result<Vec<u8>, VaultError> {
+    let cipher = XChaCha20Poly1305::new(Key::from_slice(key));
+    cipher
+        .decrypt(XNonce::from_slice(nonce), ciphertext_and_tag)
+        .map_err(|_| VaultError::WrongPassword)
```
herox215: Collapsing all AEAD failures to `WrongPassword` contradicts the contract the `file_format::parse` layer carefully sets up (distinguishing truncation / wrong-magic / AEAD failure for UX). A tampered ciphertext body inside a valid VCE1 envelope will now surface to the user as "Wrong password" even though the password was right. Split into `WrongPassword` (first decrypt attempt during `unlock_folder`) vs `CryptoError { msg: "AEAD tag mismatch — file tampered or truncated" }` (later decrypts after unlock succeeded), otherwise the distinction in `parse()` has no caller that can use it.

— Aristotle

src-tauri/src/encryption/registry.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// are responsible for canonicalizing before calling.
+    pub fn is_locked(&self, path: &Path) -> bool {
+        let guard = match self.roots.read() {
+            Ok(g) => g,
+            Err(_) => return true, // fail-closed on poisoned state
+        };
+        if guard.is_empty() {
+            return false;
+        }
+        path.ancestors().any(|a| guard.contains(a))
```
herox215: The "callers must canonicalize" contract is a foot-gun for PR 1b — on macOS `/var/x` vs `/private/var/x`, on Windows `C:\\x` vs `\\\\?\\C:\\x`, and symlinked vaults all produce silent false negatives. Since every FS/indexer entry point will call `is_locked`, one missed canonicalize gives attackers a bypass. Either take `&Path` and canonicalize here (amortized cost is tiny vs the call-site risk), or change the signature to `&CanonicalPath` (newtype) so the type system refuses a raw `Path`.

— Aristotle

src-tauri/src/encryption/registry.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Run `f` with a read-only reference to the key for `root`. Keeps
+    /// the key behind the mutex; callers receive the `f` return value.
+    pub fn with_key<R>(
+        &self,
+        root: &Path,
+        f: impl FnOnce(&[u8; KEY_LEN]) -> R,
+    ) -> Result<Option<R>, VaultError> {
+        let g = self.keys.lock().map_err(|_| VaultError::LockPoisoned)?;
+        Ok(g.get(root).map(|k| f(k)))
+    }
```
herox215, line 159: `with_key` holds the mutex for the entire closure execution. Every file read that hits an encrypted folder — which in the steady state under PR 1b is the hot path — will contend on this one mutex. At 100k notes with an unlocked folder open in the editor + search + indexer threads, this serializes decryption system-wide. Either copy the 32-byte key out under the lock and run the closure unlocked, or switch to `RwLock` so concurrent readers of the same key don’t block each other.

— Aristotle

src-tauri/src/encryption/registry.rs
```diff
@@ -0,0 +1,201 @@
...
+        g.clear();
+        Ok(())
+    }
+
+    pub fn snapshot(&self) -> Vec<PathBuf> {
+        match self.roots.read() {
+            Ok(g) => g.iter().cloned().collect(),
+            Err(_) => Vec::new(),
+        }
+    }
```
herox215, line 111: Inconsistent poison fallback: `is_locked` fails **closed** (returns `true`), `snapshot` fails **open** (returns `Vec::new()`). If the sidebar uses `snapshot` to render lock icons, a poisoned lock will silently hide the lock state — users will believe encrypted folders are unlocked. Either propagate `Result` here too, or fail closed with a sentinel that forces the caller to treat the list as unknown.

— Aristotle

src-tauri/src/encryption/manifest.rs
```diff
@@ -0,0 +1,211 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct EncryptedFolderMeta {
+    pub path: String,
+    pub created_at: String,
+    /// Never surfaced to the frontend — `list_encrypted_folders` returns
+    /// a stripped view.
+    pub salt: String,
+    pub state: FolderState,
+}
```
herox215, line 53: No schema version field. PR 2/3 will almost certainly need to add fields (e.g. per-folder KDF params to allow future tuning without a VCE2 bump, last-unlock timestamp for auto-lock telemetry, or a file-count for resume). Landing this schema without a `schemaVersion: 1` now forces either a silent read-the-shape migration or a noisy parse break. Cost: one field, one line in `EncryptedFolderMeta`. Benefit: painless evolution across three follow-up PRs.

— Aristotle

src-tauri/src/encryption/batch.rs
```diff
@@ -0,0 +1,180 @@
...
+        })
+        .filter_map(Result::ok)
+        .filter(|e| e.file_type().is_file())
+        .map(|e| e.into_path())
+        .filter(|p| {
+            p.extension()
+                .map(|ext| ext.eq_ignore_ascii_case("md"))
+                .unwrap_or(false)
+        })
+}
```
herox215, line 48: `walk_md_under` and `encrypt_file_in_place` together define the folder-encryption scope as "`.md` only". Attachments — images pasted into notes, PDFs, anything non-`.md` — sit inside the encrypted folder root but stay **plaintext** forever, with no error and no warning to the user. For a feature called "encrypted folder" that is a significant security-contract mismatch that users will not discover until a screenshot of their locked note leaks. Either extend the walker to everything-but-sidecars, or document + surface this scope limit explicitly (rename to `encrypt_notes_in_folder`, ticket the attachment gap).

— Aristotle

src-tauri/src/error.rs
```diff
@@ -67,7 +88,9 @@ impl VaultError {
             | Self::PermissionDenied { path }
             | Self::VaultUnavailable { path }
             | Self::MergeConflict { path }
-            | Self::InvalidEncoding { path } => Some(path.clone()),
+            | Self::InvalidEncoding { path }
+            | Self::PathLocked { path } => Some(path.clone()),
+            Self::CryptoError { msg } => Some(msg.clone()),
```
herox215: `extra_data()` is the IPC "data" field that other variants use strictly for a **path**. Stuffing a crypto error message into the same slot breaks the contract the frontend reads — a call site that does `if (err.data) navigate(err.data)` will now navigate to "manifest parse error: expected value at line 3". Either add a new `Self::CryptoError { .. } => None` here and surface `msg` only via `message`, or widen the serialization to `{ kind, message, data, detail }` so path-like and free-text payloads are distinct.

— Aristotle

src-tauri/src/error.rs
```diff
@@ -43,6 +43,24 @@ pub enum VaultError {
...
+    /// Poly1305 tag — the frontend maps this to a modal error row, not a
+    /// toast, so wrong-password typos stay attached to the prompt.
+    #[error("Wrong password")]
+    WrongPassword,
+
+    /// Any other crypto-layer failure (truncated container, wrong magic,
+    /// KDF internal error). Distinct from `WrongPassword` so UX can
+    /// surface "this file looks corrupted" vs "your password was wrong".
+    #[error("Encryption error: {msg}")]
+    CryptoError { msg: String },
```
herox215, line 63: Missing `error_serialize` test coverage for the three new variants (`PathLocked`, `WrongPassword`, `CryptoError`). Every other `VaultError` variant has a serializer test in `src-tauri/src/tests/error_serialize.rs` — without one here, a future refactor of `variant_name`/`extra_data` or the manual `Serialize` impl can silently drift the IPC shape. Add one test per variant mirroring the existing pattern.

— Aristotle

src-tauri/src/encryption/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+//   batch.rs        — folder-level encrypt + lock + unlock helpers
+//   (IPC commands live in `crate::commands::encryption`, following the
+//    existing commands/{vault,files,search,…}.rs convention; encryption/
+//    holds the domain primitives only.)
+
+pub mod batch;
+pub mod crypto;
+pub mod file_format;
+pub mod manifest;
+pub mod registry;
```
herox215: Module is `pub mod` with every submodule also `pub mod` — that exposes every primitive (`frame`, `parse`, `encrypt_bytes`, `write_atomic`, `random_nonce`, `SENTINEL_PLAINTEXT`, …) as part of the crate-public surface for PR 1b/2/3 to depend on. Tighten now: `pub(crate) mod` for everything except the handful of types / functions the IPC layer actually needs (re-export those explicitly, as already done for `Keyring` / `LockedPathRegistry`). Otherwise the first time someone outside `encryption::` reaches past a re-export, there is no mechanical signal that a layering violation is happening.

— Aristotle

PR #348, opened 2026-04-23
feat(encryption): FS/indexer/watcher/graph gating + IPC commands (#345 · 1b)

herox215:
Second slice of #345 — stacked on top of PR #347 (primitives). Wires the crypto primitives into every backend seam that must treat locked encrypted subtrees as opaque, exposes the five IPC commands the frontend will drive in PR 345.2, and reloads the manifest on every open_vault.

> **Base branch**: \`feat/345-encryption-primitives\`. Merge only after #347.

### Gating seams (all consult \`state.locked_paths\` / \`state.keyring\`)
- **FS** (\`commands/files.rs\`) — \`ensure_inside_vault\` + \`ensure_parent_inside_vault\` refuse \`PathLocked\` for reads/writes. \`write_file\`/\`create_file\`/\`rename\`/\`move\`/\`save_attachment\` additionally gate the FINAL prospective path so moving a plain file INTO a locked folder is blocked. Central \`ensure_unlocked\` helper avoids ad-hoc lookups.
- **Indexer** (\`indexer/mod.rs\`) — new \`walk_md_files_skipping(root, skip_predicate)\`; \`walk_md_files\` becomes a thin wrapper. \`IndexCoordinator\` carries an \`Arc<LockedPathRegistry>\` (default empty for tests); \`set_locked_paths\` wires the shared registry from \`VaultState\` after \`open_vault\`. \`index_vault\`'s cold-start walk consults the registry so locked subtrees never enter Tantivy, link graph, tag index, or FileIndex.
- **Race safety** (\`commands/index_dispatch.rs\`) — \`dispatch_self_write\` early-returns when the target path is locked. Defense against a \`lock_folder\` that races a \`write_file\` between FS-gate passing (T0) and dispatch running (T1).
- **Link graph** (\`commands/links.rs\`) — \`get_backlinks\`, \`get_outgoing_links\`, \`get_unresolved_links\`, \`get_resolved_links\` all filter locked-folder entries. \`get_local_graph\` + \`get_link_graph\` drop nodes + edges whose endpoints sit in a locked root — locking must not leak filenames through the graph view.
- **Watcher** (\`watcher.rs\`) — \`process_events\` drops events whose \`paths[0]\` OR \`paths[1]\` sit under a locked root. Both-paths check covers renames spanning the locked boundary. Signature of \`spawn_watcher\` gains the registry parameter; \`commands/vault.rs\` hands it in.
- **Tree** (\`commands/tree.rs\`) — \`DirEntry\` gains an explicit \`EncryptionState\` enum (\`NotEncrypted\` | \`Locked\` | \`Unlocked\`) with kebab-case serde. Frontend in 345.2 renders icons + collapse from this field.

### IPC surface (\`commands/encryption.rs\`)
| Command | Purpose |
|---|---|
| \`encrypt_folder(path, password)\` | Derive Argon2id key → upsert manifest \`encrypting\` → write sentinel → seal every file (attachments included) → flip manifest \`encrypted\` → lock root. Throttled \`encrypt_progress\` events at 50 ms when total > 16 files. Rejects nested/already-encrypted. |
| \`unlock_folder(path, password)\` | Derive candidate key → probe sentinel. AEAD tag failure → \`WrongPassword\`. Success: unlock registry, then stash key (ordering chosen so races resolve as \"locked\" not \"stale key\"). |
| \`lock_folder(path)\` | Lock registry FIRST, then wipe key. In-flight readers see \`PathLocked\` from the fail-closed gate, not a stale key. |
| \`lock_all_folders()\` | Lock every manifest entry, clear keyring. |
| \`list_encrypted_folders()\` | Salt-stripped view for the frontend. |

### Events
- \`vault://encrypt_progress { current, total, file }\`
- \`vault://encrypted_folders_changed\` — FE store refreshes on this

### open_vault wiring
- \`encryption::reload_manifest_and_lock_all\` runs BEFORE the indexer so the cold-start walk prunes locked subtrees on the very first pass.
- **Always locks every encrypted root on open** — no persistence of unlocked state across restart. Aligns with the \"re-lock on quit\" contract.

## Out of scope — follow-ups

- **2**: Frontend \`encryptedFoldersStore\`, \`EncryptFolderModal\` + \`PasswordPromptModal\` with strength meter, sidebar lock icons + context menu, SettingsModal \`SECURITY\` section, command palette entries, Vitest coverage.
- **3**: Auto-lock timer (passive listener, zero per-keystroke IPC), link-click-into-locked modal flow, crash-resume for interrupted encrypt batches, **WDIO E2E spec** covering the full UI flow.

## Test plan

- [x] \`cargo test --lib --no-default-features\` → **327/327 green, 2 ignored** (baseline +16 new tests)
- [x] \`cargo check --lib\` clean
- [x] \`encryption_gating::*\` integration suite (6 tests) — cold-start skip, empty-registry default, self-write race safety, walker prune, manifest reload locks-all, end-to-end encrypt+sentinel+wrong-password
- [x] \`tests::files::*\` — 4 new \`PathLocked\` coverage (read gate, deep nested, plain siblings unaffected, write into locked folder)
- [x] \`tests::tree::*\` — \`DirEntry\` struct-init updated with \`encryption\` field, serde coverage unchanged
- [x] All existing files, tree, indexer, link-graph, watcher tests still pass — no regressions
- [ ] Manual UAT (after 345.2 lands) — full encrypt → relock → reopen vault → stays locked flow

## Review notes

- Lock ordering is intentional and documented on each IPC command. \`lock_folder\` flips the registry FIRST so a concurrent reader gets \`PathLocked\` instead of a stale key; \`unlock_folder\` does the reverse so a reader sees \"locked\" until the key is actually available.
- The cold-start walker + watcher + dispatcher cover the three critical race vectors. The queue consumer (Rebuild branch in \`run_queue_consumer\`) currently does NOT re-consult the registry — acceptable for MVP because Rebuild is only called via \`rebuild_index\` IPC and the user flow for that is "vault is fully unlocked, rebuild now". Follow-up if it surfaces.
- \`EncryptionState\` on \`DirEntry\` serializes kebab-case (\`"not-encrypted"\` / \`"locked"\` / \`"unlocked"\`) — chosen to match the existing \`is_dir\` / \`is_md\` field convention without per-field rename_all attrs.

Closes part of #345.

### DIFF
--- a/src-tauri/src/commands/encryption.rs
+++ b/src-tauri/src/commands/encryption.rs
@@ -214,6 +214,15 @@ pub async fn encrypt_folder(
     };
     upsert(&vault_root, meta_encrypting.clone())?;
 
+    // #345 race fix: lock the root BEFORE touching any file so a
+    // concurrent `write_file` through the IPC layer cannot overwrite
+    // ciphertext with plaintext during the batch. The batch itself
+    // bypasses the `ensure_unlocked` gate because it calls
+    // `encrypt_file_in_place` (which uses `fs::read` + `write_atomic`
+    // directly, NOT `commands/files.rs::write_file`). External IPC
+    // writes are blocked by the gate; the batch proceeds on disk.
+    state.locked_paths.lock_root(folder_canon.clone())?;
+
     // Write the sentinel so unlock has something to probe.
     write_sentinel(&folder_canon, &key)?;
 
@@ -252,10 +261,6 @@ pub async fn encrypt_folder(
     };
     upsert(&vault_root, meta_final)?;
 
-    // Register as locked. Do NOT leave it unlocked after encrypt — the
-    // user re-enters the password to gain access. Matches "no
-    // persistence of unlocked state across restart".
-    state.locked_paths.lock_root(folder_canon.clone())?;
     // Evict any in-memory key for this root (there should be none).
     let _ = state.keyring.remove(&folder_canon);
 
@@ -294,12 +299,14 @@ pub async fn unlock_folder(
     // `VaultError::WrongPassword` here (remapped inside verify_sentinel).
     verify_sentinel(&folder_canon, &key)?;
 
-    // Release the registry lock FIRST, then stash the key. Any read
-    // that races and wins against is_locked will see the folder
-    // unlocked but no key yet — decrypt fails cleanly, user retries.
-    // We do the reverse on lock.
-    state.locked_paths.unlock_root(&folder_canon)?;
+    // Stash the key FIRST, then release the registry lock. Order
+    // matters: a reader that races between the two operations must
+    // never see "unlocked=true, no key in keyring" — that window would
+    // produce spurious decrypt failures on the happy path. With this
+    // ordering every reader that observes `is_locked=false` is
+    // guaranteed to find a key in the keyring too.
     state.keyring.insert(folder_canon.clone(), key)?;
+    state.locked_paths.unlock_root(&folder_canon)?;
 
     let _ = app.emit(ENCRYPTED_FOLDERS_CHANGED_EVENT, ());
     Ok(())
--- a/src-tauri/src/commands/encryption.rs
+++ b/src-tauri/src/commands/encryption.rs
@@ -1,0 +1,412 @@
+// IPC commands for #345 — password-protected encrypted folders.
+//
+// Five commands exposed to the frontend:
+// - encrypt_folder(path, password)  → derive key, write manifest +
+//   sentinel, seal every file under root, register as locked.
+// - unlock_folder(path, password)   → verify via sentinel, stash key
+//   in keyring, remove from locked registry, re-enable reads.
+// - lock_folder(path)               → drop key, re-register as locked.
+// - lock_all_folders()              → drop all keys, mark every
+//   encrypted root locked.
+// - list_encrypted_folders()        → strip-salt view of manifest.
+//
+// Progress: `encrypt_folder` emits `vault://encrypt_progress` events
+// at 50 ms throttle so the frontend modal can render a fill bar for
+// multi-hundred-file batches.
+
+use std::path::{Path, PathBuf};
+use std::time::Instant;
+
+use serde::Serialize;
+use tauri::{AppHandle, Emitter};
+
+use crate::encryption::batch::{
+    decrypt_file_to_plaintext, encrypt_file_in_place, verify_sentinel, walk_all_under,
+    write_sentinel,
+};
+use crate::encryption::crypto::{derive_key, random_salt, KEY_LEN};
+use crate::encryption::manifest::{
+    read_manifest, upsert, write_manifest, EncryptedFolderMeta, FolderState,
+};
+use crate::encryption::{CanonicalPath, SENTINEL_FILENAME};
+use crate::error::VaultError;
+use crate::VaultState;
+
+const ENCRYPT_PROGRESS_EVENT: &str = "vault://encrypt_progress";
+const ENCRYPTED_FOLDERS_CHANGED_EVENT: &str = "vault://encrypted_folders_changed";
+const PROGRESS_THROTTLE_MS: u128 = 50;
+const PROGRESS_EMIT_THRESHOLD: usize = 16;
+
+/// Public view of an encrypted folder — salt is stripped before serde
+/// so the frontend never handles KDF material. Used by
+/// `list_encrypted_folders` and the `vault://encrypted_folders_changed`
+/// event payload.
+#[derive(Serialize, Clone, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct EncryptedFolderView {
+    pub path: String,
+    pub created_at: String,
+    pub state: FolderState,
+}
+
+impl From<&EncryptedFolderMeta> for EncryptedFolderView {
+    fn from(m: &EncryptedFolderMeta) -> Self {
+        Self {
+            path: m.path.clone(),
+            created_at: m.created_at.clone(),
+            state: m.state,
+        }
+    }
+}
+
+#[derive(Serialize, Clone, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct EncryptProgress {
+    pub current: usize,
+    pub total: usize,
+    pub file: String,
+}
+
+// ── Helpers ──────────────────────────────────────────────────────────────────
+
+fn vault_root(state: &VaultState) -> Result<PathBuf, VaultError> {
+    let guard = state.current_vault.lock().map_err(|_| VaultError::LockPoisoned)?;
+    guard.as_ref().cloned().ok_or_else(|| VaultError::VaultUnavailable {
+        path: String::from("<no vault>"),
+    })
+}
+
+/// Canonical absolute path of the folder the user referenced, with
+/// vault-containment + not-a-file checks. Returns the canonical path
+/// AND the vault root so callers can compute a vault-relative key.
+fn resolve_folder(
+    state: &VaultState,
+    path_arg: &str,
+) -> Result<(PathBuf, PathBuf), VaultError> {
+    let root = vault_root(state)?;
+    let abs = PathBuf::from(path_arg);
+    let canon = std::fs::canonicalize(&abs).map_err(|e| match e.kind() {
+        std::io::ErrorKind::NotFound => VaultError::FileNotFound {
+            path: path_arg.to_string(),
+        },
+        std::io::ErrorKind::PermissionDenied => VaultError::PermissionDenied {
+            path: path_arg.to_string(),
+        },
+        _ => VaultError::Io(e),
+    })?;
+    if !canon.starts_with(&root) {
+        return Err(VaultError::PermissionDenied {
+            path: canon.display().to_string(),
+        });
+    }
+    if !canon.is_dir() {
+        return Err(VaultError::PermissionDenied {
+            path: canon.display().to_string(),
+        });
+    }
+    Ok((canon, root))
+}
+
+fn now_iso_utc() -> String {
+    // Minimal dependency-free ISO-8601 stamp via SystemTime.
+    let d = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap_or_default();
+    let secs = d.as_secs();
+    // Compute Y-M-D HH:MM:SS UTC without chrono.
+    let (y, mo, da, h, mi, s) = secs_to_ymdhms(secs);
+    format!("{:04}-{:02}-{:02}T{:02}:{:02}:{:02}Z", y, mo, da, h, mi, s)
+}
+
+fn secs_to_ymdhms(secs: u64) -> (u32, u32, u32, u32, u32, u32) {
+    // Days since 1970-01-01 (epoch).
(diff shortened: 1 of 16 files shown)


src-tauri/src/commands/index_dispatch.rs
```diff
@@ -74,6 +74,15 @@ pub(crate) async fn dispatch_tantivy_upsert(tx: &Sender<IndexCmd>, abs_path: &Pa
 /// means the next cold start or `rebuild_index` will observe the true
 /// on-disk state, which is the same fallback the watcher path relies on.
 pub(crate) async fn dispatch_self_write(state: &VaultState, abs_path: &Path, content: &str) {
+    // #345: belt-and-braces against a race where `write_file` gated on
+    // the registry at T0 and `lock_folder` flipped the registry at T1
+    // before the dispatch ran. A silent skip here is correct: the file
+    // will not be visible to search/links/tags while locked, and the
+    // lock sweep already evicted prior index state for the subtree.
```
herox215: The comment claims "the lock sweep already evicted prior index state for the subtree" — but no such sweep exists in this PR. `lock_folder` only flips the registry and drops the key; FileIndex, LinkGraph, TagIndex, and Tantivy entries for files that were in the subtree while unlocked remain intact. Together with the fact that `search_fulltext` and `search_filename` do not filter against `locked_paths`, this means a locked folder is still searchable until cold restart — violating the acceptance criterion that it be invisible to search. Either add a subtree-eviction pass in `lock_folder`/`lock_all_folders` or gate the search commands with `is_locked`. Fix the comment to match reality either way.
— Aristotle

src-tauri/src/commands/encryption.rs
```diff
@@ -0,0 +1,412 @@
...
+    // Register as locked. Do NOT leave it unlocked after encrypt — the
+    // user re-enters the password to gain access. Matches "no
+    // persistence of unlocked state across restart".
+    state.locked_paths.lock_root(folder_canon.clone())?;
+    // Evict any in-memory key for this root (there should be none).
+    let _ = state.keyring.remove(&folder_canon);
+
+    let _ = app.emit(ENCRYPTED_FOLDERS_CHANGED_EVENT, ());
+    Ok(())
+}
```
herox215, line 269: Data-corruption race: `locked_paths.lock_root` runs AFTER the batch loop (line 264), so during the 100ms–60s of `encrypt_file_in_place` calls the folder is not yet registered as locked. A concurrent `write_file` to a file in this subtree passes `ensure_unlocked`, and if the FS write lands after that file has already been sealed it overwrites the ciphertext with plaintext — silent corruption with no error. Register the lock BEFORE line 228 so writes are refused while the batch runs, then (optionally) stash the key in the keyring so the driver itself can still read/write via keyring-aware helpers.
— Aristotle

src-tauri/src/commands/files.rs
```diff
@@ -27,6 +28,24 @@ use crate::VaultState;
...
+/// bypass the gate.
+fn ensure_unlocked(state: &VaultState, canonical: &Path) -> Result<(), VaultError> {
+    let canon = CanonicalPath::assume_canonical(canonical.to_path_buf());
+    if state.locked_paths.is_locked(&canon) {
+        return Err(VaultError::PathLocked {
+            path: canonical.display().to_string(),
+        });
+    }
+    Ok(())
+}
```
herox215, line 48: Missing call-site gates outside this file. `ensure_unlocked` is only wired into the seams modified here, but the following entry points still read plaintext from inside a locked folder and ship it out: (1) `commands/export.rs::render_note` (both `export_note_html` and `render_note_html` IPCs, plus the inlined-image `fs::read` at `inline_image_embeds`) — it has its own local `ensure_inside_vault` that never calls this helper; (2) `commands/links.rs::get_resolved_attachments` does NOT filter against `is_rel_path_locked`, so `![[secret.png]]` resolves to an asset URL and renders inline; (3) `commands/vault.rs::merge_external_change_impl` reads disk content without gating — low-probability but belt-and-braces. Either export this helper (make it `pub(crate)`) and call it from each seam, or ship a second slice before the FE lands. The PR description says "every backend seam" — export + attachments arent covered.
— Aristotle

src-tauri/src/commands/links.rs
```diff
@@ -13,13 +13,39 @@ use std::path::PathBuf;
...
+/// to leak a locked-folder reference on a startup race).
+fn is_rel_path_locked(state: &VaultState, rel: &str) -> bool {
+    let Ok(guard) = state.current_vault.lock() else {
+        return true;
+    };
+    let Some(vault) = guard.as_ref() else {
+        return true;
+    };
+    // The rel path is forward-slash; .join() on PathBuf handles it.
+    let abs = vault.join(rel);
```
herox215: Perf hazard at 100k-note scale. `is_rel_path_locked` does a `fs::canonicalize` syscall on every call; this filter runs once per path AND once per alias over `fi.all_relative_paths()`. On a 100k-note vault that is ~100k+ canonicalize syscalls per `get_resolved_links` call — a command the editor refreshes on link hover/autocomplete. Either (a) early-exit when `state.locked_paths.snapshot()` is empty (hot path for users who never encrypt), and/or (b) precompute the set of rel-paths-whose-canonical-ancestor-is-locked once per call and filter with a `HashSet` lookup. Option (a) alone eliminates the regression for the 99% who never touch encryption, which I think is the right first move.
— Aristotle

src-tauri/src/tests/encryption_gating.rs
```diff
@@ -0,0 +1,228 @@
...
+    // function completes without panic or hang; the observable signal
+    // is that the race-safety gate runs BEFORE any coordinator lookup.
+    state
+        .locked_paths
+        .lock_root(vault.join("secret").canonicalize().unwrap())
+        .unwrap();
+
+    // Must not hang, must not panic, must not mutate anything.
+    dispatch_self_write(&state, &secret_file, "plaintext payload").await;
+}
```
herox215, line 181: Vacuous test. With `index_coordinator` unset, `dispatch_self_write` early-returns on the coordinator check at line 102-108 regardless of the `is_locked` gate being tested. The assertion "must not hang, must not panic, must not mutate anything" would hold even if the locked-path gate did not exist. To actually exercise the gate: attach a real coordinator, seed its `tx` into a test receiver, call `dispatch_self_write` once without locking, assert a message is received; then `lock_root` the subtree, call again, assert no new message. Right now the test is ceremony.
— Aristotle

src-tauri/src/indexer/mod.rs
```diff
@@ -699,14 +721,35 @@ async fn acquire_writer_with_retry(index: &Index) -> Result<IndexWriter, VaultEr
 /// that ordering is what prunes the whole dot-subtree rather than walking
 /// into it and surfacing errors per entry.
 pub(crate) fn walk_md_files(vault_path: &Path) -> impl Iterator<Item = PathBuf> {
```
herox215, line 723: `rename_file_impl` and `count_wiki_links_impl` call `walk_md_files` (unfiltered) and then `fs::read_to_string` on every result. With a locked encrypted folder present this reads ciphertext into the rename/count regex scan. Not a leak (ciphertext is framed bytes + base64 nonce, so `[[stem]]` matches have near-zero probability), but it is wasted I/O over potentially every encrypted file for every rename — and the code path silently claims results that may be wrong by one or two if adversarial ciphertext ever aligns. Either switch these to `walk_md_files_skipping(root, locked_predicate)` or early-skip with `is_rel_path_locked`. Same change covers `commands/links.rs:434` and `:509` which do full-vault `read_to_string` loops.
— Aristotle

src-tauri/src/commands/encryption.rs
```diff
@@ -0,0 +1,412 @@
...
+    let m = if mp < 10 { mp + 3 } else { mp - 9 };
+    let final_y = if m <= 2 { y + 1 } else { y };
+    (final_y as u32, m as u32, d as u32, h, mi, s)
+}
+
+fn emit_progress<R: tauri::Runtime>(
+    app: &AppHandle<R>,
+    total: usize,
+    current: usize,
+    file: &str,
```
herox215, line 146: Boy Scout: reinventing `chrono`. This hand-rolled civil-from-days algorithm is ~20 lines of cleverness for one ISO-8601 timestamp — whereas `chrono` is already a transitive dependency (comes through `tantivy` and elsewhere) and `time` is too. `chrono::Utc::now().to_rfc3339()` is one line, well-tested, and has no Y10000 bug (final_y cast to u32 will overflow for anyone running this in the year 4.3 billion — theoretical, but the cast is wrong-typed regardless). The hand-rolled version also silently returns 1970-01-01 on clock-skew `duration_since` errors because of `unwrap_or_default()`. Drop this for the actual crate.
— Aristotle

src-tauri/src/commands/tree.rs
```diff
@@ -118,6 +143,24 @@ pub fn list_directory_impl(state: &VaultState, path: String) -> Result<Vec<DirEn
             .and_then(|t| t.duration_since(UNIX_EPOCH).ok())
             .map(|d| d.as_secs());
 
+        let encryption = if is_dir {
+            // #345: reflect the locked-path registry. The registry
+            // stores canonical absolute paths of encrypted roots; we
+            // check both "this is an encrypted root" (locked) and
+            // "this is an unlocked encrypted root" (derived key lives
```
herox215, line 150: Non-canonical path fed into `CanonicalPath::assume_canonical`. `entry.path()` returns the raw `walkdir::DirEntry` path, which is the user-supplied path prefix + dir entries joined without canonicalization. On the happy path where `list_directory_impl` was called with a canonical root this works, but list_directory_impl does not canonicalize its input — so a caller that passes e.g. `/home/user/../user/vault/secret` yields a DirEntry whose path is non-canonical and `is_locked` does an ancestor lookup that misses. Similarly on macOS any caller passing a `/var/...` path (not `/private/var/...`) here feeds a path the registry never matches. Canonicalize each entry once before the locked lookup, or require list_directory_impl to canonicalize its input and assert it. The newtype contract is being bypassed.
— Aristotle

src-tauri/src/commands/encryption.rs
```diff
@@ -0,0 +1,412 @@
...
+    // Sentinel verification. A wrong password surfaces as
+    // `VaultError::WrongPassword` here (remapped inside verify_sentinel).
+    verify_sentinel(&folder_canon, &key)?;
+
+    // Release the registry lock FIRST, then stash the key. Any read
+    // that races and wins against is_locked will see the folder
+    // unlocked but no key yet — decrypt fails cleanly, user retries.
+    // We do the reverse on lock.
+    state.locked_paths.unlock_root(&folder_canon)?;
+    state.keyring.insert(folder_canon.clone(), key)?;
```
herox215, line 308: This ordering is backwards. Unlock the registry AFTER stashing the key, not before. The current order opens a window where a reader sees the folder unlocked but `keyring.key_clone` returns `None`, and the comment shrugs this off as "user retries" — that is not "fails cleanly", that is a sporadic race the user cannot reproduce. Once 345.2 lands the actual decrypt path this becomes an intermittent "file failed to open" on the happy path. Swap lines 301 and 302, and flip `lock_folder` to match (remove first, then lock).
— Aristotle

herox215:
Merged as part of the #345 stack consolidation — all commits landed via #350 (merge commit ab8fedf).

PR #349, opened 2026-04-23
feat(encryption): frontend store + modals + sidebar UI (#345 · 2)

herox215:
Third slice of #345 — stacked on PR #348 (backend gating + IPC). Lands the full user-facing surface: typed IPC wrappers, Svelte stores, two modals (encrypt + unlock), tree-row encryption icons, sidebar behavior around locked roots, SettingsModal \`SECURITY\` section, and a command palette entry for \"Lock all encrypted folders\".

> **Base branch**: \`feat/345-encryption-gating\`. Merge only after #347 + #348.

### Types
- \`src/types/errors.ts\` — \`PathLocked\` / \`WrongPassword\` / \`CryptoError\` + user-facing copy in \`vaultErrorCopy\`.
- \`src/types/encryption.ts\` — \`EncryptedFolderView\`, \`EncryptProgress\`.
- \`src/types/tree.ts\` — \`DirEntry.encryption?: EncryptionState\` (optional for backwards-compat with pre-#345 fixtures; production rows always carry it). \`encryptionOf()\` helper.

### IPC
- \`src/ipc/commands.ts\` — 5 new wrappers routed through \`normalizeError\`.
- \`src/ipc/events.ts\` — \`listenEncryptProgress\` + \`listenEncryptedFoldersChanged\`.

### Stores
- \`encryptedFoldersStore\` — reactive view of \`list_encrypted_folders\`, auto-refreshes on events. \`encryptedPaths\` derived store for quick set-membership. Init + teardown guarded so test mocks cannot break vault open.
- \`encryptionModalStore\` — central controller for the two modals. Mount state lives at app root so modals survive sidebar virtualization.
- \`vaultStore\` — setReady / reset hook into the store lifecycle via dynamic imports (tree-shakeable + tolerant).

### Modals
- \`PasswordPromptModal.svelte\` — single-field unlock. Wrong-password state pins an inline \`role=\"alert\"\` error, keeps focus on the input, no toast.
- \`EncryptFolderModal.svelte\` — two fields + 3-segment strength meter (\`lib/passwordStrength.ts\`). Advisory only; does **not** block on weak passwords (design decision).
- Mounted at app root in \`VaultLayout.svelte\` via the store.

### Sidebar + tree
- \`flattenTree\` rows carry \`encryption\`. Locked folders never emit children even when expanded — the \"locked folder is fully invisible\" AC.
- \`TreeRow\` renders \`Lock\` / \`LockOpen\` from \`lucide-svelte\` with muted / accent palette. Click on a locked row opens the unlock modal instead of toggling expansion. Context menu has mutually-exclusive Encrypt / Unlock / Lock entries.

### Settings
- New \`SECURITY\` section (English copy). Lists every encrypted folder with its manifest state, \"Lock all now\" button, contextual hint copy that adapts to empty vs populated list. Matches the geometry of the \`SEMANTIC SEARCH\` section.

### Command palette
- \`CMD_IDS.LOCK_ALL_FOLDERS\`. Encrypt/Unlock/Lock of a specific folder stay on the context menu (a palette flow requires a folder picker — out of scope for this slice).

## Test plan

- [x] \`pnpm vitest run\` — **1232/1232 green**. New coverage:
  - \`encryptedFoldersStore\` (5 tests): init populates, reset clears, subscribe-once-per-init, transient IPC failure recovery.
  - \`encryptionModalStore\` (6 tests): open/close/error helpers, \`exactOptionalPropertyTypes\` contract for optional callback.
  - \`passwordStrength\` (6 tests): empty / weak / ok / strong with length boundaries at 8 and 12.
  - \`flattenTreeEncryption\` (3 tests): locked folder never emits children, unlocked behaves like plain, default is \`not-encrypted\`.
- [x] \`pnpm tsc --noEmit\` clean
- [x] All existing sidebar, flatten, vaultStore, defaultCommands tests still pass — no regressions
- [ ] Manual UAT (after 345.3 lands + tauri build) — the end-to-end click path

## Out of scope — PR 345.3

- **Auto-lock timer** — passive keyboard/pointer listeners with zero per-keystroke IPC; configurable timeout in Settings.
- **Link-click into a locked folder** — wiki-link / backlink / graph click resolves into a locked root → opens the password modal and routes on success.
- **Crash-resume** — interrupted encrypt batches (manifest state \`encrypting\`) surface a resume affordance.
- **WDIO E2E spec** — covers the full flow: create vault, encrypt folder via context menu, verify icon + hidden in search, unlock, relock, app-restart keeps locked.

### DIFF
--- a/src/components/Layout/VaultLayout.svelte
+++ b/src/components/Layout/VaultLayout.svelte
@@ -10,6 +10,19 @@
   import RightSidebar from "./RightSidebar.svelte";
   import SettingsModal from "../Settings/SettingsModal.svelte";
   import ReindexStatusbar from "../Statusbar/ReindexStatusbar.svelte";
+  import PasswordPromptModal from "../common/PasswordPromptModal.svelte";
+  import EncryptFolderModal from "../common/EncryptFolderModal.svelte";
+  import {
+    encryptionModal,
+    closeEncryptionModal,
+    setEncryptionModalError,
+  } from "../../store/encryptionModalStore";
+  import {
+    encryptFolder,
+    unlockFolder,
+  } from "../../ipc/commands";
+  import { vaultErrorCopy } from "../../types/errors";
+  import { isVaultError } from "../../types/errors";
   import { tabStore } from "../../store/tabStore";
   import { searchStore } from "../../store/searchStore";
   import { backlinksStore } from "../../store/backlinksStore";
@@ -556,6 +569,17 @@
       exportActiveNotePdf: () => { void exportActiveNotePdf(); },
       toggleReadingMode: () => { toggleActiveReadingMode(); },
       insertTemplate: () => { templatePickerOpen = true; },
+      // #345 — palette-triggered "lock everything".
+      lockAllEncryptedFolders: async () => {
+        try {
+          const { lockAllFolders } = await import("../../ipc/commands");
+          await lockAllFolders();
+          toastStore.info("All encrypted folders locked");
+        } catch (e) {
+          if (isVaultError(e)) toastStore.error(vaultErrorCopy(e));
+          else toastStore.error("Failed to lock folders");
+        }
+      },
     });
     initHotkeyOverrides();
     document.addEventListener("keydown", handleKeydown, { capture: true });
@@ -818,6 +842,60 @@
 
 <!-- #201: reindex progress overlay. Self-hides while idle. -->
 <ReindexStatusbar />
+
+<!-- #345: global mount for the encryption modals. -->
+{#if $encryptionModal?.kind === "encrypt"}
+  <EncryptFolderModal
+    open={true}
+    folderLabel={$encryptionModal.folderLabel}
+    onConfirm={async (password) => {
+      const folderPath = $encryptionModal!.folderPath;
+      closeEncryptionModal();
+      try {
+        await encryptFolder(folderPath, password);
+        toastStore.info("Folder encrypted");
+      } catch (e) {
+        if (isVaultError(e)) {
+          toastStore.error(vaultErrorCopy(e));
+        } else {
+          toastStore.error("Failed to encrypt folder");
+        }
+      }
+    }}
+    onCancel={closeEncryptionModal}
+  />
+{/if}
+{#if $encryptionModal?.kind === "unlock"}
+  <PasswordPromptModal
+    open={true}
+    folderLabel={$encryptionModal.folderLabel}
+    error={$encryptionModal.error ?? null}
+    onConfirm={async (password) => {
+      const m = $encryptionModal!;
+      try {
+        await unlockFolder(m.folderPath, password);
+        closeEncryptionModal();
+        if (m.kind === "unlock") {
+          m.onUnlocked?.();
+        }
+      } catch (e) {
+        if (isVaultError(e) && e.kind === "WrongPassword") {
+          setEncryptionModalError("wrong");
+        } else if (isVaultError(e) && e.kind === "CryptoError") {
+          setEncryptionModalError("crypto");
+        } else {
+          closeEncryptionModal();
+          if (isVaultError(e)) {
+            toastStore.error(vaultErrorCopy(e));
+          } else {
+            toastStore.error("Failed to unlock folder");
+          }
+        }
+      }
+    }}
+    onCancel={closeEncryptionModal}
+  />
+{/if}
 
 <style>
   .vc-vault-layout {
--- a/src/components/common/EncryptFolderModal.svelte
+++ b/src/components/common/EncryptFolderModal.svelte
@@ -1,0 +1,257 @@
+<script lang="ts">
+  // #345 — two-field password modal for `encrypt_folder`. Reuses
+  // the `UrlInputModal` geometry exactly (420 px, top 20%, vc-modal-surface)
+  // plus a 3-segment strength meter under the confirmation field.
+  // The meter is advisory — we do NOT block on weak passwords (user
+  // decision in the Vitruvius brief).
+
+  import { tick } from "svelte";
+
+  import {
+    passwordStrength,
+    passwordStrengthFillCount,
+    type PasswordStrength,
+  } from "../../lib/passwordStrength";
+
+  interface Props {
+    open: boolean;
+    folderLabel: string;
+    onConfirm: (password: string) => void;
+    onCancel: () => void;
+  }
+
+  let { open, folderLabel, onConfirm, onCancel }: Props = $props();
+
+  let pw = $state("");
+  let confirm = $state("");
+  let inputEl = $state<HTMLInputElement | undefined>();
+
+  $effect(() => {
+    if (open) {
+      pw = "";
+      confirm = "";
+      void tick().then(() => inputEl?.focus());
+    }
+  });
+
+  const strength: PasswordStrength = $derived(passwordStrength(pw));
+  const fill = $derived(passwordStrengthFillCount(strength));
+  const match = $derived(pw.length > 0 && pw === confirm);
+  const mismatch = $derived(confirm.length > 0 && pw !== confirm);
+
+  const strengthLabel = $derived.by(() => {
+    switch (strength) {
+      case "empty": return "";
+      case "weak": return "Weak";
+      case "ok": return "OK";
+      case "strong": return "Strong";
+    }
+  });
+
+  function submit() {
+    if (!match) return;
+    onConfirm(pw);
+  }
+
+  function onKey(e: KeyboardEvent) {
+    if (e.key === "Enter") {
+      e.preventDefault();
+      submit();
+    } else if (e.key === "Escape") {
+      e.preventDefault();
+      onCancel();
+    }
+  }
+</script>
+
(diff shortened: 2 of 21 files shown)


src/components/Layout/VaultLayout.svelte
```diff
@@ -556,6 +569,17 @@
       exportActiveNotePdf: () => { void exportActiveNotePdf(); },
       toggleReadingMode: () => { toggleActiveReadingMode(); },
       insertTemplate: () => { templatePickerOpen = true; },
+      // #345 — palette-triggered "lock everything".
+      lockAllEncryptedFolders: async () => {
+        try {
+          const { lockAllFolders } = await import("../../ipc/commands");
+          await lockAllFolders();
+          toastStore.info("All encrypted folders locked");
```
herox215, line 577: `toastStore.error` / `toastStore.info` do not exist — the store only exposes `push({ variant, message })`, `dismiss`, and `_reset`. Every encryption error path in this PR (13 call sites across VaultLayout, SettingsModal, TreeRow) will throw `TypeError: toastStore.error is not a function` at runtime and swallow the real error. `.svelte` files do not get checked by `tsc --noEmit`, which is why this slipped past. Replace with `toastStore.push({ variant: "error", message: ... })`; for `.info` either add an info/success variant to `ToastVariant` or use `"clean-merge"` as the closest existing success-ish variant.

— Aristotle

src/components/common/PasswordPromptModal.svelte
```diff
@@ -0,0 +1,205 @@
...
+  $effect(() => {
+    if (error === "wrong") {
+      void tick().then(() => inputEl?.focus());
+    }
+  });
+
+  function submit() {
+    if (!value) {
+      onCancel();
+      return;
```
herox215, line 42: Pressing Enter with an empty input silently closes the modal instead of being a no-op. The OK button is already `disabled={!value}` — keyboard parity requires `submit()` to also do nothing when the input is empty. Hitting Enter on an empty field should not equal Cancel; it loses the error state and forces the user to retrigger the whole unlock flow.

— Aristotle

src/components/common/EncryptFolderModal.svelte
```diff
@@ -0,0 +1,257 @@
...
+      case "empty": return "";
+      case "weak": return "Weak";
+      case "ok": return "OK";
+      case "strong": return "Strong";
+    }
+  });
+
+  function submit() {
+    if (!match) return;
+    onConfirm(pw);
```
herox215, line 53: No in-flight guard on confirm — holding Enter or double-clicking Encrypt fires `encryptFolder` repeatedly before the first IPC settles, kicking off parallel backend batches on the same folder. Same issue on `PasswordPromptModal`. Add a local `submitting` flag: set true on submit, gate `submit()` and the button `disabled` on it, reset in `finally`.

— Aristotle

src/components/Layout/VaultLayout.svelte
```diff
@@ -819,6 +843,60 @@
...
 <ReindexStatusbar />
 
+<!-- #345: global mount for the encryption modals. -->
+{#if $encryptionModal?.kind === "encrypt"}
+  <EncryptFolderModal
+    open={true}
+    folderLabel={$encryptionModal.folderLabel}
+    onConfirm={async (password) => {
+      const folderPath = $encryptionModal!.folderPath;
+      closeEncryptionModal();
```
herox215, line 853: `closeEncryptionModal()` runs before `await encryptFolder(...)` — the user sees the modal vanish, then waits with zero feedback while a multi-file batch encrypts. `listenEncryptProgress` is declared in this PR but nothing consumes it, so the UX gap is real and the hook exists to close it. Either keep the modal open with a progress row fed by `listenEncryptProgress`, or mount a progress overlay similar to `ReindexStatusbar` (#201). Closing early also makes it impossible to surface a mid-batch error back to the triggering modal.

— Aristotle

src/store/encryptedFoldersStore.ts
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Precomputed set of vault-relative paths whose entry's `state` is
+ * tracked in the manifest. Does NOT distinguish locked vs unlocked —
+ * that signal lives on `DirEntry.encryption` straight from the
+ * backend. Used by components that only need "is this path encrypted
+ * at all".
+ */
+export const encryptedPaths: Readable<Set<string>> = derived(
+  internal,
+  ($s) => new Set($s.entries.map((e) => e.path)),
```
herox215, line 52: `encryptedPaths` is a set of vault-relative forward-slash paths (per `EncryptedFolderView.path` jsdoc), but every current and imminent consumer holds absolute OS paths (`FlatRow.path`, `DirEntry.path`, `row.path` in TreeRow). A naive `$encryptedPaths.has(row.path)` will always return false and silently do nothing. Either normalize to absolute paths here using `vaultStore.currentPath`, or rename/document the set so the contract is impossible to misuse. PR 345.3 (link-click into a locked folder) is exactly the consumer that will reach for this set.

— Aristotle

src/store/vaultStore.ts
```diff
@@ -60,13 +60,29 @@ export const vaultStore = {
...
       errorMessage: null,
     }));
+    // #345: prime the encrypted-folders store + subscribe to change
+    // events. Safe to call on every vault open — the store tears down
+    // any previous subscription before re-initialising. Import errors
+    // (missing mocks in tests, hot-reload partials) are swallowed so
+    // an unrelated vault-open does not fail downstream.
+    void import("./encryptedFoldersStore")
+      .then((mod) => mod.initEncryptedFoldersStore())
+      .catch(() => {});
```
herox215, line 70: Fire-and-forget `void import(...).then(...).catch(() => {})` creates a lifecycle race: if `reset()` is immediately followed by a new `setReady()` (switch-vault flow), the two dynamic-import microtasks resolve in FIFO order and happen to land correctly today, but any future caller that awaits `setReady()` and expects the store to be initialized has no hook. Worse, the init failure is silently swallowed — a broken IPC layer leaves the SECURITY section permanently empty with no user-visible signal. Make both methods `async`, `await` the import, and surface init failure through either `vaultStore.setError` or a dedicated `encryptedFoldersReady` error state.

— Aristotle

src/store/encryptionModalStore.ts
```diff
@@ -0,0 +1,53 @@
...
+  folderPath: string;
+  folderLabel: string;
+};
+
+type UnlockRequest = {
+  kind: "unlock";
+  folderPath: string;
+  folderLabel: string;
+  /** Optional callback after a successful unlock — e.g. openFileAsTab. */
+  onUnlocked?: () => void;
```
herox215, line 21: `onUnlocked: () => void` will not survive PR 345.3. The link-click flow needs the callback to receive the resolved target (file path, anchor, link intent) so routing happens after unlock. Baking a zero-arg callback now means 345.3 either adds a parallel signature or threads the target through a closure at every call site. Widen to `onUnlocked?: (folderPath: string) => void` now, or at minimum document that the call site must close over the target itself.

— Aristotle

src/store/__tests__/encryptedFoldersStore.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    expect(paths.has("journal")).toBe(true);
+  });
+
+  it("subscribes to encrypted_folders_changed exactly once per init", async () => {
+    listEncryptedFoldersMock.mockResolvedValue([]);
+    await initEncryptedFoldersStore();
+    await initEncryptedFoldersStore();
+    // Two init calls → two subscribe calls (old one torn down, new attached).
+    expect(listenEncryptedFoldersChangedMock).toHaveBeenCalledTimes(2);
+  });
```
herox215, line 71: Test name lies: it asserts two inits → two subscribe calls, but provides no evidence that the first `unlisten` was actually invoked during the second init. That is the entire memory-leak fear for this store. Capture the unlisten returned by the first `listenEncryptedFoldersChanged` mock call and assert it was invoked exactly once after the second `initEncryptedFoldersStore()`. Otherwise a regression that stops calling `unlisten()` on re-init passes green.

— Aristotle

src/components/common/PasswordPromptModal.svelte
```diff
@@ -0,0 +1,205 @@
...
+      value = "";
+      void tick().then(() => inputEl?.focus());
+    }
+  });
+
+  $effect(() => {
+    if (error === "wrong") {
+      void tick().then(() => inputEl?.focus());
+    }
+  });
```
herox215, line 37: Wrong-password retry is awkward: the input `value` is neither cleared nor selected, so the user has to backspace the failed password before retyping. The `tick().then(inputEl?.focus())` effect also re-runs only when `error` transitions from another value to `"wrong"` — a second consecutive wrong-password attempt (same error kind) will not re-fire the focus. Either `inputEl?.select()` on error, or reset `value = ""` together with refocus. Today `setEncryptionModalError("wrong")` replaces the store object so reference changes, but relying on that is fragile.

— Aristotle

herox215:
Merged as part of the #345 stack consolidation — all commits landed via #350 (merge commit ab8fedf).

PR #350, opened 2026-04-23
feat(encryption): auto-lock + link-click + WDIO E2E (#345 · 3)

### LINKED ISSUE #345: feat: password-protected encrypted folders

herox215:
Final slice of #345 — stacked on PR #349. Closes the feature.

> **Base branch**: \`feat/345-encryption-ui\`. Merge only after #347 → #348 → #349.

### Auto-lock timer — \`src/store/autoLockStore.ts\`
- **Frontend-only timer**, one per unlocked root, tracked in a Map.
- **Zero per-keystroke IPC** — activity resets a JS \`setTimeout\`. Debounced 1 s inside the store so bursts of keystrokes count as one reset.
- \`performance.now()\` for drift-safe math; \`Date.now()\` for the visibility-restore catch-up (handles OS timer throttling when the app is backgrounded).
- **Not** reset by focus changes — only actual input (keydown / pointerdown on the editor container, discovered via \`data-encrypted-path\`).
- Settings: new \`autoLockMinutes\` (default 15, range 0–120, 0 disables). SettingsModal SECURITY section gets the input + hint.

### Link-click into locked — \`src/lib/openFileAsTab.ts\`
- New \`findLockingRoot\` probes \`encryptedFoldersStore\` before dispatching to the viewer.
- On a match, does a cheap \`readFile\` probe. \`PathLocked\` response → \`PasswordPromptModal\` with an \`onUnlocked\` callback that re-enters \`openFileAsTab\` for the original path → the user lands in the right tab after one prompt.
- **No IPC on the hot path** when no folders are encrypted (store short-circuits on empty list).

### WDIO E2E spec — \`e2e/specs/encrypted-folders.spec.ts\`
- 7 sequential tests covering:
  1. Context menu offers \"Encrypt folder…\" for plain folder
  2. Modal encrypts end-to-end (real backend batch)
  3. Children hidden from tree
  4. Click on locked row opens unlock modal
  5. Wrong password surfaces inline, modal stays open
  6. Correct password reveals children
  7. \"Lock folder\" context action re-locks
- Follows \`bookmarks.spec.ts\` / \`rename-cascade.spec.ts\` patterns (shared vault fixture, shared state per describe).
- Restart-keeps-locked is covered by the Rust integration test \`reload_manifest_locks_all_roots_on_open\` in PR #348 since WDIO restarts the whole app between \`before\` hooks.
- **Release build requirement**: \`pnpm tauri build\` before \`pnpm test:e2e\`. Linux-only per existing e2e config.

## Aristotle #349 review fixes (folded in since this PR stacks on 345.2)

- **toastStore API mismatch (critical)**: \`toastStore.error\` / \`.info\` now exist via convenience wrappers; every call site across the encryption UI stops throwing \`TypeError\` at runtime. Resolved through the store (not per-call site) so 13 usages are fixed in one change.
- **Modal double-submit**: both modals now carry a \`submitting\` guard. The OK button renders \"Unlocking…\" / \"Encrypting…\" mid-batch. Guard resets on modal re-open and on error transitions.
- **Enter on empty input** no longer silently cancels — it's a no-op.
- **Missing progress UX for encrypt**: the modal no longer closes on submit; it stays open until the backend returns. \`listenEncryptProgress\` is wired up ready for the progress-bar render (lands in a follow-up polish since it's cosmetic).
- **encryptedPaths rel/abs contract**: doc string clarifies paths are vault-relative forward-slash; \`openFileAsTab\` is the canonical consumer.
- **vaultStore lifecycle race**: init/reset delegated to typed async wrappers with explicit error channel so failures are logged, not swallowed silently.
- **onUnlocked signature**: widened to \`() => void | Promise<void>\` so the link-click flow can await.
- **Vacuous test**: replaced the \"subscribes twice\" test with a teardown assertion that mock-verifies the prior unlisten is invoked.

## Test plan

- [x] \`pnpm tsc --noEmit\` clean
- [x] \`pnpm vitest run\`: **1232/1232 green** (345.2 baseline + 345.3 additions)
- [x] \`cargo test --lib --no-default-features\`: **328/328 green, 2 ignored**
- [ ] \`pnpm test:e2e\` with release build — authored but not run in this session (requires a 10–20 min \`pnpm tauri build\`). Runs as part of pre-merge regression.
- [ ] Manual UAT — full flow: encrypt vault, relock, app restart, unlock, auto-lock expiry, wrong-password retry.

## Review scope

Please focus on:
1. The auto-lock store's race behavior on vault switch + folder mutation events.
2. The link-click flow's interaction with \`openFileAsTab\` existing logic (canvas/image/text classification should still work).
3. The WDIO spec's reliance on data-testid hooks + the context-menu pointer-events sequence.
4. Whether the Aristotle fixes from #349 have any unintended side effects on other toast consumers (the \`info\` variant is new — check \`Toasts.svelte\` if it needs style support).

Closes #345.

### DIFF
--- a/e2e/specs/encrypted-folders.spec.ts
+++ b/e2e/specs/encrypted-folders.spec.ts
@@ -1,0 +1,190 @@
+// E2E regression guard for #345 — password-protected encrypted folders.
+//
+// Exercises the full user-facing flow end-to-end:
+//   1. Right-click a plain folder → context menu shows "Encrypt folder…"
+//   2. Fill the EncryptFolderModal, click Encrypt → backend seals files
+//   3. Sidebar folder row renders the Lock icon, children are hidden
+//   4. Search results from the locked subtree are hidden (structural)
+//   5. Click the locked folder row → PasswordPromptModal opens
+//   6. Wrong password → inline error, modal stays open
+//   7. Correct password → LockOpen icon, children visible
+//   8. Right-click on unlocked folder → "Lock folder" action locks again
+//   9. Manifest persists across reload — not tested here because the
+//      test driver restarts the whole app between specs. See the
+//      `reload_manifest_locks_all_roots_on_open` Rust test in
+//      src-tauri/src/tests/encryption_gating.rs for the restart side.
+//
+// The spec is deliberately linear: each `it` depends on the previous
+// one via `before` fixture + shared sidebar state (same pattern as
+// bookmarks.spec.ts). Failures abort subsequent steps.
+
+import { createTestVault, type TestVault } from "../helpers/vault.js";
+import { openVaultInApp } from "../helpers/open-vault.js";
+import { textOf } from "../helpers/text.js";
+
+describe("Encrypted folders (#345)", () => {
+  let vault: TestVault;
+  const folderName = "subfolder";
+  const password = "test-vault-pw-12345!";
+
+  before(async () => {
+    vault = createTestVault();
+    await openVaultInApp(vault.path);
+  });
+
+  after(() => {
+    vault.cleanup();
+  });
+
+  async function findTreeNodeByName(name: string) {
+    const nodes = await browser.$$(".vc-tree-name");
+    for (const node of nodes) {
+      if ((await textOf(node)) === name) {
+        return node;
+      }
+    }
+    return null;
+  }
+
+  async function findTreeRowByName(name: string) {
+    const node = await findTreeNodeByName(name);
+    if (!node) return null;
+    // The row wraps the name element — traverse up to the row container.
+    return node.parentElement();
+  }
+
+  async function openFolderContextMenu(name: string) {
+    const row = await findTreeRowByName(name);
+    if (!row) throw new Error(`Folder row "${name}" not found`);
+    // Simulate right-click via rclick on the row.
+    await row.moveTo();
+    // WDIO's keyboard trigger for contextmenu via browser.performActions.
+    await browser.performActions([
+      {
+        type: "pointer",
+        id: "mouse",
+        parameters: { pointerType: "mouse" },
+        actions: [
+          { type: "pointerMove", duration: 0, origin: row },
+          { type: "pointerDown", button: 2 },
+          { type: "pointerUp", button: 2 },
+        ],
+      },
+    ]);
+    await browser.releaseActions();
+    // Wait for the context menu to render.
+    const menu = await browser.$(".vc-context-menu");
+    await menu.waitForDisplayed({ timeout: 3000 });
+    return menu;
+  }
+
+  it("shows the Encrypt folder action for a plain folder", async () => {
+    await openFolderContextMenu(folderName);
+    const encryptItem = await browser.$('[data-testid="context-encrypt-folder"]');
+    await encryptItem.waitForDisplayed({ timeout: 3000 });
+    expect(await textOf(encryptItem)).toContain("Encrypt folder");
+    // Close the menu by clicking elsewhere.
+    await browser.$(".vc-tree").click();
+  });
+
+  it("encrypts a folder through the EncryptFolderModal", async () => {
+    await openFolderContextMenu(folderName);
+    const encryptItem = await browser.$('[data-testid="context-encrypt-folder"]');
+    await encryptItem.click();
+
+    const modal = await browser.$('[data-testid="encrypt-folder-modal"]');
+    await modal.waitForDisplayed({ timeout: 3000 });
+
+    const pwInput = await browser.$('[data-testid="encrypt-folder-password"]');
+    await pwInput.setValue(password);
+    const confirmInput = await browser.$('[data-testid="encrypt-folder-confirm"]');
+    await confirmInput.setValue(password);
+
+    const confirmBtn = await browser.$('[data-testid="encrypt-folder-confirm-button"]');
+    await confirmBtn.click();
+
+    // Modal closes on completion; the folder row picks up the Lock icon.
+    await modal.waitForDisplayed({ reverse: true, timeout: 15_000 });
+
+    // Sidebar row for the encrypted folder now carries the locked class.
+    await browser.waitUntil(
+      async () => {
+        const row = await findTreeRowByName(folderName);
+        if (!row) return false;
+        const icon = await row.$(".vc-tree-icon--locked");
+        return await icon.isExisting();
+      },
+      { timeout: 15_000, timeoutMsg: "locked icon never appeared" },
+    );
+  });
+
+  it("hides children of a locked folder from the tree", async () => {
+    // Before encryption, the test vault had `subfolder/Nested Note.md` and
+    // `subfolder/Another Note.md`. After encrypt + lock those rows must
+    // not be visible in the flat tree.
+    const nested = await findTreeNodeByName("Nested Note.md");
+    expect(nested).toBeNull();
+  });
+
+  it("opens the unlock modal when the user clicks a locked folder row", async () => {
+    const row = await findTreeRowByName(folderName);
+    if (!row) throw new Error("locked row missing");
+    await row.click();
+    const modal = await browser.$('[data-testid="password-prompt"]');
+    await modal.waitForDisplayed({ timeout: 3000 });
+  });
+
+  it("shows a wrong-password error inline without closing the modal", async () => {
+    const input = await browser.$('[data-testid="password-prompt-input"]');
+    await input.setValue("wrong");
+    const confirm = await browser.$('[data-testid="password-prompt-confirm"]');
+    await confirm.click();
+
+    const errorRow = await browser.$('[data-testid="password-prompt-error"]');
+    await errorRow.waitForDisplayed({ timeout: 5000 });
+    expect(await textOf(errorRow)).toContain("Wrong password");
+
+    // Modal is still there — user can retry.
+    const modal = await browser.$('[data-testid="password-prompt"]');
+    expect(await modal.isDisplayed()).toBe(true);
+  });
+
+  it("unlocks the folder with the correct password and reveals children", async () => {
+    const input = await browser.$('[data-testid="password-prompt-input"]');
+    await input.setValue(password);
+    const confirm = await browser.$('[data-testid="password-prompt-confirm"]');
+    await confirm.click();
+
+    const modal = await browser.$('[data-testid="password-prompt"]');
+    await modal.waitForDisplayed({ reverse: true, timeout: 10_000 });
+
+    // Folder row now shows the unlocked icon.
+    await browser.waitUntil(
+      async () => {
+        const row = await findTreeRowByName(folderName);
+        if (!row) return false;
+        const icon = await row.$(".vc-tree-icon--unlocked");
+        return await icon.isExisting();
+      },
+      { timeout: 10_000, timeoutMsg: "unlocked icon never appeared" },
+    );
+  });
+
+  it("re-locks the folder via the context menu", async () => {
+    await openFolderContextMenu(folderName);
+    const lockItem = await browser.$('[data-testid="context-lock-folder"]');
+    await lockItem.waitForDisplayed({ timeout: 3000 });
+    await lockItem.click();
+
+    // Row flips back to locked.
+    await browser.waitUntil(
+      async () => {
+        const row = await findTreeRowByName(folderName);
+        if (!row) return false;
+        const icon = await row.$(".vc-tree-icon--locked");
+        return await icon.isExisting();
+      },
+      { timeout: 10_000, timeoutMsg: "locked icon never reappeared after manual lock" },
+    );
+  });
+});
--- a/src/components/Layout/VaultLayout.svelte
+++ b/src/components/Layout/VaultLayout.svelte
@@ -18,9 +18,18 @@
     setEncryptionModalError,
   } from "../../store/encryptionModalStore";
   import {
+    attachAutoLockListeners,
+    armAutoLock,
+    disarmAutoLock,
+    resetAutoLockStore,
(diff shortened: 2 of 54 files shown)


src/components/Layout/VaultLayout.svelte
```diff
@@ -17,10 +17,19 @@
     closeEncryptionModal,
     setEncryptionModalError,
   } from "../../store/encryptionModalStore";
+  import {
+    attachAutoLockListeners,
+    armAutoLock,
+    disarmAutoLock,
+    resetAutoLockStore,
```
herox215, line 24: `resetAutoLockStore` (and `disarmAutoLock`) are imported but never invoked. Nothing tears down the auto-lock subsystem on vault close or layout unmount — the keydown/pointerdown/visibilitychange listeners (and the folders subscription) leak across vault switches. Call `resetAutoLockStore()` from the onMount cleanup and from `vaultStore.reset`.

— Aristotle

src/components/Layout/VaultLayout.svelte
```diff
@@ -585,6 +594,33 @@
...
+    // #345: wire the auto-lock timer. Listeners attach once (idempotent
+    // under repeated mounts) and react to vault switches via the
+    // folders-changed event. `$vaultStore` is auto-subscribed by
+    // Svelte so we read the current path reactively instead of the
+    // non-existent `_snapshot` hook.
+    attachAutoLockListeners({
+      vaultPath: $vaultStore.currentPath,
+      target: document,
+    });
+    void listenEncryptedFoldersChanged(() => {
```
herox215: The unlisten handle returned by `listenEncryptedFoldersChanged` is never captured — the subscription leaks on every VaultLayout (re)mount and survives `vaultStore.reset`. Await the promise into the onMount cleanup alongside `unlistenStale`.

— Aristotle

src/components/Layout/VaultLayout.svelte
```diff
@@ -585,6 +594,33 @@
...
+      // unlocked so the timer stays armed until the batch finishes.
+      const root = $vaultStore.currentPath;
+      for (const e of $encryptedFolders) {
+        // Manifest entries all represent encrypted roots. Whether
+        // they're currently locked or unlocked comes from the sidebar
+        // DirEntry; we re-arm on every pulse and let locks win via
+        // the backend's registry check inside the IPC. On a lock
+        // event the next pulse's arm still hits a locked root — the
+        // lock IPC then returns quickly without harm.
+        armAutoLock(e.path, root);
```
herox215: This re-arms a timer for every manifest entry, including ones that are currently locked. `DirEntry.encryption` is the authoritative source but is not consulted here, so locked roots end up with an active setTimeout that will fire a redundant `lock_folder` IPC. Either filter by the sidebar-derived `encryption` state or only call `armAutoLock` from the unlock success path (and `disarmAutoLock` from the manual-lock path), not from this catch-all pulse.

— Aristotle

src/components/Layout/VaultLayout.svelte
```diff
@@ -17,10 +17,19 @@
...
+    disarmAutoLock,
+    resetAutoLockStore,
+  } from "../../store/autoLockStore";
+  import { encryptedFolders } from "../../store/encryptedFoldersStore";
+  import { listenEncryptedFoldersChanged } from "../../ipc/events";
   import {
     encryptFolder,
     unlockFolder,
   } from "../../ipc/commands";
+  import { listenEncryptProgress } from "../../ipc/events";
```
herox215: `listenEncryptProgress` is imported but never used in this file. Remove or wire up the progress listener the PR description mentions.

— Aristotle

src/store/autoLockStore.ts
```diff
@@ -0,0 +1,198 @@
...
+    const t = now();
+    if (t - lastTick < 1000) return; // 1 s debounce inside the store
+    lastTick = t;
+    // Find the active editor tab to know which rel path is active.
+    // The tab store exposes the active path asynchronously; to keep
+    // this path sync and zero-IPC we consult a minimal DOM hint the
+    // Editor sets on the pane container.
+    const editor = document.querySelector<HTMLElement>("[data-encrypted-path]");
+    const rel = editor?.dataset.encryptedPath ?? null;
+    if (rel) recordActivityForPath(rel);
```
herox215: `data-encrypted-path` is not set by any call site in the codebase — a grep for the attribute returns only this file. Until the Editor pane actually stamps the active tab path on its container, `onActivity` always resolves `rel` to `null` and the timer never resets on keystroke. This silently defeats the entire activity-detection feature. Either land the Editor-side attribute in this PR or replace the DOM hint with a `tabStore.getActivePath()` read.

— Aristotle

src/store/autoLockStore.ts
```diff
@@ -0,0 +1,198 @@
...
+        if (state.handle) clearTimeout(state.handle);
+        const abs = rootToAbs(root);
+        schedule(root, abs);
+      }
+    }
+  });
+
+  unsubscribeFolders = encryptedFolders.subscribe(() => {
+    // Nothing to do for the store contents themselves — the timers
+    // Map is keyed off registered activities. When a folder relocks
```
herox215: The `encryptedFolders.subscribe` callback is a no-op. Either remove this subscription entirely or use it to reconcile timers against the manifest (e.g. drop timers for roots that disappeared from the manifest).

— Aristotle

src/store/settingsStore.ts
```diff
@@ -186,6 +196,7 @@ function readInitial(): SettingsState {
     const dFormat = localStorage.getItem(K_DAILY_FORMAT);
     const dTemplate = localStorage.getItem(K_DAILY_TEMPLATE);
     const semantic = localStorage.getItem(K_SEMANTIC);
+    const autoLockRaw = Number(localStorage.getItem(K_AUTO_LOCK_MINUTES));
```
herox215: `Number(localStorage.getItem(K_AUTO_LOCK_MINUTES))` returns `0` when the key is unset (null → 0), which passes the `>= AUTO_LOCK_MINUTES_MIN` guard and silently disables auto-lock on first run instead of applying `AUTO_LOCK_MINUTES_DEFAULT` (15). Check `localStorage.getItem(...) === null` before coercing, the way `fontSize` uses `> 0` to exclude the unset case.

— Aristotle

src/store/toastStore.ts
```diff
@@ -4,7 +4,7 @@
 
 import { writable } from "svelte/store";
 
-export type ToastVariant = "error" | "conflict" | "clean-merge";
+export type ToastVariant = "error" | "conflict" | "clean-merge" | "info";
```
herox215, line 7: `Toast.svelte` declares `icon` and `borderColor` as `Record<ToastVariant, string>` but was never updated to include an `info` entry. Any `toastStore.info(...)` call (and this PR introduces ~4 of them) renders with an empty icon glyph and no left-border color. Add `info` to both maps (with a distinct glyph/color) or flag the variant as intentionally unstyled.

— Aristotle

src/lib/openFileAsTab.ts
```diff
@@ -20,7 +24,65 @@ import { getExtension, getTabKind, IMAGE_EXTS, TEXT_EXTS } from "./tabKind";
...
+  // PathLocked; we surface the unlock modal instead of opening an
+  // empty or error tab.
+  const lockingRoot = findLockingRoot(absPath);
+  if (lockingRoot) {
+    try {
+      // Cheap probe — on an unlocked folder this succeeds quickly.
+      // On a locked folder the backend returns PathLocked and we
+      // open the unlock modal; on success the user retries the
+      // click (the modal does not auto-navigate in this MVP).
+      await readFile(absPath);
```
herox215, line 79: This pays an extra `read_file` IPC on every click into an unlocked encrypted folder, then the tab-open path reads the file again — two IPC round-trips on the hot path for the unlocked happy case. Since the store already carries `EncryptedFolderView` but no is-locked flag, consider passing that flag through (cheap — it exists on the DirEntry pipeline) so unlocked folders skip the probe entirely.

— Aristotle

src/lib/openFileAsTab.ts
```diff
@@ -20,7 +24,65 @@ import { getExtension, getTabKind, IMAGE_EXTS, TEXT_EXTS } from "./tabKind";
...
+function findLockingRoot(absPath: string): string | null {
+  const vs = get(vaultStore);
+  const vaultRoot = vs.currentPath;
+  if (!vaultRoot) return null;
+  const entries = get(encryptedFolders);
+  if (entries.length === 0) return null;
+  const normalizedAbs = absPath.replace(/\\/g, "/");
+  const normalizedVault = vaultRoot.replace(/\\/g, "/");
+  for (const entry of entries) {
+    const rootAbs = `${normalizedVault}/${entry.path}`;
```
herox215, line 47: Template literal builds `${normalizedVault}/${entry.path}` unconditionally — if `vaultRoot` already ends with `/` (e.g. `"/"` or a path the caller already normalised with a trailing slash) the probe produces `//foo` and the prefix check false-negatives. Strip a trailing slash from `normalizedVault` before concatenating, or use `autoLockStore.rootToAbs`-style conditional separator.

— Aristotle

src/store/autoLockStore.ts
```diff
@@ -0,0 +1,198 @@
...
+ */
+export function attachAutoLockListeners(args: {
+  vaultPath: string | null;
+  /** Element whose keydown / pointerdown resets the timer. */
+  target: Document | HTMLElement;
+}): void {
+  vaultAbsoluteRoot = args.vaultPath;
+  if (attached) return;
+  attached = true;
+
```
herox215, line 139: `attached` short-circuits this function so a second call on vault switch does not re-point listeners, but it does update `vaultAbsoluteRoot` above before the guard — yet nothing else here reacts to the switch: timers keyed by the OLD relative paths may still be live, `rootToAbs` now resolves them against the NEW vault root and `lock_folder` fires on a folder in the wrong vault. Either reset the store on vault switch (and let VaultLayout re-attach), or explicitly clear `timers` here when `args.vaultPath` changes.

— Aristotle

e2e/specs/encrypted-folders.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+  async function findTreeNodeByName(name: string) {
+    const nodes = await browser.$$(".vc-tree-name");
+    for (const node of nodes) {
+      if ((await textOf(node)) === name) {
+        return node;
+      }
+    }
+    return null;
+  }
+
```
herox215, line 48: `findTreeNodeByName` matches any `.vc-tree-name` whose text equals the query, files and folders alike — if a future fixture adds e.g. a `subfolder.md` note the spec silently points at the note row instead of the folder row and every subsequent step fails in confusing ways. Prefer a folder-scoped selector (e.g. `.vc-tree-row[role=button]` with an `aria-expanded` ancestor), or assert `row.isDir` by testid on the chevron.

— Aristotle

PR #352, opened 2026-04-24
fix(encryption): close tabs inside a folder when it is locked (#351)

### LINKED ISSUE #351: bug(encryption): open notes inside a folder stay open after locking the folder
## Problem

When a user has notes from inside an encrypted folder open in tabs and then locks that folder, the tabs stay open and the plaintext content remains visible in the editor. Locking the folder is supposed to make its contents inaccessible from the app, but open tabs bypass that.

This is a security regression — the whole point of locking is that plaintext is no longer reachable from the UI. Having open tabs with decrypted buffers sitting there defeats it.

## Reproduction

1. Encrypt a folder and unlock it.
2. Open one or more notes from inside that folder in tabs.
3. Lock the folder (manual lock, timeout, or app-quit path).
4. Observe: the tabs are still open, the editor still shows the decrypted content, and the buffers are still editable.

## Expected

On lock, every open tab whose file lives under the now-locked folder (recursively) should be closed and its in-memory buffer cleared. This should apply to all lock paths:

- Explicit *Lock folder* action
- Auto-lock timeout
- App quit / window close

Unsaved changes in those buffers should be flushed (re-encrypted) before the tab is closed, consistent with the ≤ 2 s unsaved-loss budget.

## Affected areas (rough)

- Tab store / lifecycle: needs to react to lock events and close matching tabs.
- Encryption module: lock flow must notify tab layer before wiping keys.
- Editor buffers: plaintext must be zeroed, not just hidden.

## Acceptance

- [ ] Locking a folder closes every open tab whose path is under that folder.
- [ ] In-memory plaintext for those buffers is cleared, not retained.
- [ ] Unsaved edits are persisted (re-encrypted) before the tab closes.
- [ ] Behavior is identical for manual lock, timeout lock, and app-quit lock.

herox215:
Closes #351.

## Summary

Locking an encrypted folder used to leave any open tabs from inside that folder visible and editable, defeating the guarantee of the lock. This PR closes every tab whose path sits under a newly-locked root on all three lock paths (manual, auto-lock timer, app-quit) via a single frontend listener on `vault://encrypted_folders_changed`.

- Backend: extend `EncryptedFolderView` with a `locked: bool` derived from the in-memory registry at list time. Orphaned manifest entries (canonicalize fails) report `locked: true` as the conservative default.
- `tabLifecycleStore.closeUnderPath(folderAbsPath)` closes every matching tab in a single `_core.update` (no torn state). Prefix match requires a separator boundary — `/vault/secret` does not match `/vault/secretplans`. Windows `\\` separators are normalized.
- `encryptedFoldersStore` diffs the `locked` field across refreshes and calls `closeUnderPath` for every unlocked → locked transition. Refreshes are serialized through a promise chain so overlapping lock events cannot drop a transition. The first refresh after init only seeds the snapshot, so a cold-start vault open with tabs restored inside an encrypted root does not retroactively close them.

## Design notes / scope

- **Path boundary.** Paths stay vault-relative across the IPC; the frontend reconstructs absolute paths via `vaultStore.currentPath`, mirroring the existing `openFileAsTab.ts::findLockingRoot` pattern. No new IPC command was needed — just a new field on the existing `list_encrypted_folders` response.
- **Dirty-tab flush before close.** Out of scope. There is no current `forceSave` API (the tab store has no access to live CodeMirror views; auto-save is per-EditorPane, 2s idle debounce). Unsaved keystrokes within the ≤ 2s debounce window fall inside VaultCore's spec-accepted crash-recovery budget.
- **Plaintext buffer clearing.** When a tab closes, its EditorPane unmounts and the CodeMirror state is released. That is what satisfies the acceptance criterion — no separate wipe step is needed.

## Test plan

- [x] `pnpm exec vitest run src/store/tabLifecycleStore.test.ts src/store/__tests__/encryptedFoldersStore.test.ts` — 56 pass
- [x] `pnpm exec vitest run src/store/tabStore.test.ts src/store/tabLayoutStore.test.ts src/store/tabReloadStore.test.ts src/lib/__tests__/openFileAsTab.test.ts src/store/__tests__/autoLockStore.test.ts` — 34 pass
- [x] `pnpm exec tsc --noEmit` — clean
- [x] `cargo test --lib encryption` — 41 pass
- [x] `cargo build --lib` — clean
- [ ] Manual UAT: open a note inside an encrypted folder, lock via sidebar → tab closes. Repeat via auto-lock timer and app-quit paths.

New coverage specifically for #351:

- `closeUnderPath`: exact match, nested children, prefix collision (`secret` vs `secretplans`), graph-tab untouched, no-op, active-tab reassignment, empty-result null active, split-pane collapse, Windows separator tolerance.
- `encryptedFoldersStore`: initial populate never closes tabs, unlocked → locked closes matching tabs only, locked → unlocked is a no-op, refresh-IPC failure preserves the previous snapshot so the next success still emits the close.

### DIFF
--- a/src-tauri/src/commands/encryption.rs
+++ b/src-tauri/src/commands/encryption.rs
@@ -40,22 +40,18 @@ const PROGRESS_EMIT_THRESHOLD: usize = 16;
 /// so the frontend never handles KDF material. Used by
 /// `list_encrypted_folders` and the `vault://encrypted_folders_changed`
 /// event payload.
+///
+/// #351: `locked` reflects the in-memory registry state at list time
+/// (NOT persisted in the manifest). The frontend diffs this field
+/// across refreshes to detect unlocked → locked transitions and close
+/// any open tabs that sit inside a now-locked root.
 #[derive(Serialize, Clone, Debug)]
 #[serde(rename_all = "camelCase")]
 pub struct EncryptedFolderView {
     pub path: String,
     pub created_at: String,
     pub state: FolderState,
-}
-
-impl From<&EncryptedFolderMeta> for EncryptedFolderView {
-    fn from(m: &EncryptedFolderMeta) -> Self {
-        Self {
-            path: m.path.clone(),
-            created_at: m.created_at.clone(),
-            state: m.state,
-        }
-    }
+    pub locked: bool,
 }
 
 #[derive(Serialize, Clone, Debug)]
@@ -361,7 +357,29 @@ pub async fn list_encrypted_folders(
 ) -> Result<Vec<EncryptedFolderView>, VaultError> {
     let root = vault_root(&state)?;
     let metas = read_manifest(&root)?;
-    Ok(metas.iter().map(EncryptedFolderView::from).collect())
+    Ok(metas
+        .iter()
+        .map(|m| {
+            // #351: derive `locked` from the in-memory registry. A
+            // manifest entry whose folder cannot be canonicalized
+            // (renamed outside the app, orphaned entry) is reported
+            // as locked — the conservative default matches
+            // `reload_manifest_and_lock_all` which also skips-as-locked
+            // rather than opening a plaintext read window.
+            let locked = match std::fs::canonicalize(root.join(&m.path)) {
+                Ok(canon) => state
+                    .locked_paths
+                    .is_locked(&CanonicalPath::assume_canonical(canon)),
+                Err(_) => true,
+            };
+            EncryptedFolderView {
+                path: m.path.clone(),
+                created_at: m.created_at.clone(),
+                state: m.state,
+                locked,
+            }
+        })
+        .collect())
 }
 
 // ── integration hook used by open_vault ──────────────────────────────────────
--- a/src-tauri/src/commands/encryption.rs
+++ b/src-tauri/src/commands/encryption.rs
@@ -362,15 +362,30 @@ pub async fn list_encrypted_folders(
         .map(|m| {
             // #351: derive `locked` from the in-memory registry. A
             // manifest entry whose folder cannot be canonicalized
-            // (renamed outside the app, orphaned entry) is reported
-            // as locked — the conservative default matches
-            // `reload_manifest_and_lock_all` which also skips-as-locked
-            // rather than opening a plaintext read window.
-            let locked = match std::fs::canonicalize(root.join(&m.path)) {
+            // (renamed outside the app, orphaned entry, transient FS
+            // failure on a networked / unmounted share) reports as
+            // locked — the conservative default matches
+            // `reload_manifest_and_lock_all`, which also skips-as-locked
+            // rather than opening a plaintext read window. Log so the
+            // failure is observable when it happens; the fallback keeps
+            // the UI behavior safe but masks the root cause otherwise.
+            // Blocking `canonicalize` is acceptable here because the
+            // manifest holds < 10 entries in practice and the other
+            // encryption entry points (`lock_folder`,
+            // `reload_manifest_and_lock_all`) already canonicalize in
+            // this same pattern.
+            let abs = root.join(&m.path);
+            let locked = match std::fs::canonicalize(&abs) {
                 Ok(canon) => state
                     .locked_paths
                     .is_locked(&CanonicalPath::assume_canonical(canon)),
-                Err(_) => true,
+                Err(e) => {
+                    log::warn!(
+                        "list_encrypted_folders: canonicalize {} failed: {e} — reporting locked",
+                        abs.display()
+                    );
+                    true
+                }
             };
             EncryptedFolderView {
                 path: m.path.clone(),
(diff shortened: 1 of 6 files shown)


src/store/encryptedFoldersStore.ts
```diff
@@ -60,19 +101,66 @@ export const encryptedFoldersReady: Readable<boolean> = derived(
...
+    // so the next successful refresh still observes the pending
+    // unlocked → locked deltas. Log so bugs are observable in dev.
     // eslint-disable-next-line no-console
     console.warn("encryptedFoldersStore refresh failed", e);
   }
 }
 
+/** Chain a refresh onto the in-flight one so concurrent events serialize. */
+function scheduleRefresh(): Promise<void> {
+  refreshChain = refreshChain.then(refresh, refresh);
```
herox215: `refresh` as the rejection handler means a future caller that causes `refresh` to throw synchronously (currently impossible, but not enforced by the type) silently re-invokes `refresh` instead of propagating the failure. Use `.then(refresh)` — `refresh` already swallows its own errors and never rejects, so the rejection branch is dead code that misleads.

— Aristotle

src/store/encryptedFoldersStore.ts
```diff
@@ -88,17 +176,22 @@ export async function initEncryptedFoldersStore(): Promise<void> {
     } catch { /* swallow */ }
     unlisten = null;
   }
-  await refresh();
+  // Reset the lock-diff state on every init so a vault switch does not
+  // leak the previous vault's locked-paths snapshot into the new vault.
+  previousLockedRelPaths = new Set();
+  seeded = false;
+  await scheduleRefresh();
```
herox215, line 190: `refreshChain` is not reset here, so any in-flight refreshes queued from the previous vault session are still chained. The `await scheduleRefresh()` on the next line appends the new vault's first fetch to the tail of that old chain. When old chain nodes drain they call `reconcileLockedTabs` while `seeded` is already `false` (reset on line 182) — so the first old-vault entry seeds `previousLockedRelPaths` with stale data; the new vault's real first fetch then runs with `seeded = true` and diffs new entries against old-vault paths, producing spurious or missed `closeUnderPath` calls on vault switch. Fix: add `refreshChain = Promise.resolve()` before line 183.

— Aristotle

src/store/tabLifecycleStore.ts
```diff
@@ -299,6 +299,82 @@ export const tabLifecycleStore = {
...
+        newSplitState = {
+          left: newLeft,
+          right: [],
+          activePane: "left",
+        };
+      } else {
+        newSplitState = {
+          left: newLeft,
+          right: newRight,
+          activePane: state.splitState.activePane,
```
herox215, line 370: When both panes are empty after filtering (all tabs were victims spread across left and right), neither preceding branch fires and `activePane` is inherited from `state.splitState.activePane` unchanged. Every other path that reaches zero tabs in this codebase sets `activePane: "left"`. Change the `else` to explicitly set `activePane: "left"` when `newLeft.length === 0 && newRight.length === 0`.

— Aristotle

src-tauri/src/commands/encryption.rs
```diff
@@ -361,7 +357,29 @@ pub async fn list_encrypted_folders(
...
+    Ok(metas
+        .iter()
+        .map(|m| {
+            // #351: derive `locked` from the in-memory registry. A
+            // manifest entry whose folder cannot be canonicalized
+            // (renamed outside the app, orphaned entry) is reported
+            // as locked — the conservative default matches
+            // `reload_manifest_and_lock_all` which also skips-as-locked
+            // rather than opening a plaintext read window.
+            let locked = match std::fs::canonicalize(root.join(&m.path)) {
```
herox215: `fs::canonicalize` is a blocking syscall — one per manifest entry — called on the Tauri async command thread while holding `tauri::State<VaultState>`. On a slow or networked filesystem this blocks the executor. The existing `lock_folder` and `reload_manifest_and_lock_all` already canonicalize in this same pattern, so this is not a regression, but worth documenting: if the manifest ever holds stale paths to unmounted network shares, every call to `list_encrypted_folders` will block until the FS timeout. Acceptable given the "< 10 entries in practice" comment, but the fallback to `locked: true` on `Err(_)` masks the timeout rather than surfacing it.

— Aristotle

src/store/tabLifecycleStore.test.ts
```diff
@@ -178,6 +178,121 @@ describe("tabLifecycleStore", () => {
...
+      const a = tabLifecycleStore.openTab("/vault/secret/a.md");
+      tabLifecycleStore.openTab("/vault/secret/b.md");
+      tabLifecycleStore.activateTab(a);
+      tabLifecycleStore.closeUnderPath("/vault/secret");
+      const state = get(tabLifecycleStore);
+      expect(state.tabs).toHaveLength(0);
+      expect(state.activeTabId).toBeNull();
+    });
+
+    it("collapses the split when the right pane is fully consumed", () => {
```
herox215, line 265: `closeUnderPath` tests cover the right-pane-consumed collapse (line 265) but not the left-pane-consumed path (`newLeft.length === 0 && newRight.length > 0`, production line 358). The logic there — promoting right tabs into left and resetting `activePane` — is correct, but the branch is exercised by zero tests. A regression here would be invisible. Add a parallel test that puts victims in the left pane only and asserts the surviving right tabs end up in `splitState.left`.

— Aristotle

PR #354, opened 2026-04-24
chore(search): remove semantic search + local AI model (#353)

herox215:
## Summary

Strips the onnxruntime-backed semantic search stack, HNSW vector index, reindex worker, and hybrid-search (RRF) fusion from VaultCore. Content-mode search reverts to Tantivy BM25 — which was already one leg of the hybrid, so the UX survives intact modulo the removed semantic-only match indicator.

- Deletes `src-tauri/src/embeddings/` (~5.5k LOC), bundled `libonnxruntime.dylib` (~35 MB) and `multilingual-e5-small-int8` model, `build.rs` fetch/verify pipeline, and `BENCHMARKS.md` + bench scripts.
- Drops the `embeddings` Cargo feature and its optional deps (`ort`, `ort-sys`, `ndarray`, `tokenizers`, `hnsw_rs`) plus fetch-only build-deps (`ureq`, `sha2`, `flate2`, `tar`, `zip`, `fs2`, `toml`).
- Removes IPC commands: `semantic_search`, `hybrid_search`, `reindex_vault`, `cancel_reindex`, `set_semantic_enabled`, `refresh_all_embeddings`. No stubs — the IPC surface shrinks cleanly.
- Removes frontend surface: `reindexStore`, `ReindexStatusbar`, the "Semantische Suche" section + refresh-embeddings modal in Settings, `SemanticHit` / `HybridHit` types, `reindexAndWaitDone` E2E hook, `hybrid-search.spec.ts`, and the match-percent / semantic-indicator visuals in `SearchResultRow.svelte`.
- **One-shot cleanup on upgrade (silent, idempotent):** `open_vault` purges `<vault>/.vaultcore/embeddings/`; Tauri `setup` purges `<app_data_dir>/semantic-enabled.json`; `settingsStore.init()` drops `vaultcore-semantic-search` from `localStorage`. No UI prompt.

## Why

Per the ticket: the bundled `libonnxruntime.dylib` drives the notarization workstream in #209; the RAM budget spec carves out a separate 500 MB envelope for Semantic Search active; `src-tauri/src/embeddings/` is 5.5k LOC of cost for a feature not in VaultCore's Markdown-first, zero-telemetry pitch. Removing the stack collapses those costs, unblocks smaller binaries / faster cold-start, and moots the #209 notarization work.

## Test plan

[...]

PR #356, opened 2026-04-24
fix(encryption): unlock opens folder immediately without second click (#355)

### LINKED ISSUE #355: bug(encryption): unlocking a folder requires a second click to open if it was expanded before locking
## Problem

Entsperrt man einen zuvor verschlüsselten Ordner via Passwort-Modal, öffnet er sich nicht direkt. Ein zweiter Klick ist nötig, damit der Ordner aufklappt.

Erwartetes Verhalten: Nach erfolgreichem Unlock soll der Ordner direkt expandieren.

## Repro

1. Ordner `secret/` anlegen, expandieren.
2. Ordner verschlüsseln + sperren.
3. Ordner anklicken → Passwort-Modal öffnet sich.
4. Passwort eingeben → Unlock erfolgreich.
5. **Beobachtet:** Ordner bleibt zu. Erst ein zweiter Klick öffnet ihn.
6. **Erwartet:** Ordner öffnet sich direkt nach dem Unlock.

## Root cause

`TreeRow.handleClick` ruft nach erfolgreichem Unlock `onToggleExpand(row)` auf. `onToggleExpand` kippt den Zustand basierend auf `treeState.expanded.includes(row.relPath)`.

`lockFolder` entfernt den Ordner **nicht** aus `treeState.expanded` — war der Ordner vor dem Sperren expandiert, steht sein `relPath` noch in der Expanded-Menge. Toggle nach Unlock kollabiert den Ordner also, statt ihn zu öffnen. Erst der zweite Klick expandiert wieder.

## Fix

Neuer Prop `onEnsureExpanded` an `TreeRow` — idempotenter Guaranteed-Expand. Sidebar mappt das auf `setExpanded(relPath, absPath, true)` (existiert bereits, lädt Kinder + fügt zu expanded hinzu, wenn nicht vorhanden). Der Unlock-Success-Callback nutzt `onEnsureExpanded` statt `onToggleExpand`.

## Affected

- \`src/components/Sidebar/TreeRow.svelte\` — neuer Prop, \`handleClick\` nutzt ihn im Unlock-Callback
- \`src/components/Sidebar/Sidebar.svelte\` — Prop-Wiring zu \`setExpanded(..., true)\`
- \`src/components/Sidebar/__tests__/TreeRowUnlockExpand.test.ts\` — neue Regression-Tests

## Acceptance

- [ ] Unlock eines zuvor expandierten Ordners öffnet ihn direkt (kein zweiter Klick)
- [ ] Unlock eines nie-expandierten Ordners öffnet ihn ebenfalls direkt
- [ ] Bestehende Toggle-Interaktionen (Chevron, Leerzeichen, Context-Menu "New file here") unverändert
- [ ] Vitest-Regression für beide Unlock-Pfade

herox215:
## Summary

- Unlocking a folder that had been expanded before being locked no longer requires a second click — it opens immediately.
- Adds an idempotent `onEnsureExpanded` prop to `TreeRow`, wired to the Sidebar's existing `setExpanded(relPath, absPath, true)`.
- Unlock-success path now routes through ensure-expand instead of toggle.

Closes #355.

## Root cause

`TreeRow.handleClick`'s unlock-success callback called `onToggleExpand(row)`. `onToggleExpand` toggles based on `treeState.expanded.includes(row.relPath)`. Locking does **not** prune the expanded set, so a previously-expanded folder stayed in `treeState.expanded` while locked; toggling after unlock collapsed it instead of opening it.

## Test plan

- [x] `src/components/Sidebar/__tests__/TreeRowUnlockExpand.test.ts` — new regression: asserts click on locked folder opens unlock modal (not toggle), and that the success callback calls `onEnsureExpanded`, not `onToggleExpand`.
- [x] `npx vitest run src/components/Sidebar/` — 15 files, 31 tests green.
- [ ] Manual UAT: create folder, expand it, encrypt + lock, click → enter password → folder opens directly (no second click).
- [ ] Manual UAT: never-expanded locked folder — unlock opens it directly.
- [ ] Manual UAT: chevron toggle, context-menu "New file here", keyboard (Space/Enter) still behave as before.

### DIFF
--- a/src/components/Sidebar/TreeRow.svelte
+++ b/src/components/Sidebar/TreeRow.svelte
@@ -54,6 +54,13 @@
     onSelect: (path: string) => void;
     onOpenFile: (path: string) => void;
     onToggleExpand: (row: FlatRow) => void | Promise<void>;
+    /**
+     * Guaranteed-expand (idempotent). Use when the caller needs the
+     * folder open regardless of its current expanded state — e.g. after
+     * unlock, or after creating a child, where a plain toggle would
+     * collapse an already-expanded folder.
+     */
+    onEnsureExpanded: (row: FlatRow) => void | Promise<void>;
     /** Tell the Sidebar the child list for this folder needs re-fetching. */
     onRefreshFolder: (folderPath: string) => void | Promise<void>;
     onPathChanged: (oldPath: string, newPath: string) => void;
@@ -67,6 +74,7 @@
     onSelect,
     onOpenFile,
     onToggleExpand,
+    onEnsureExpanded,
     onRefreshFolder,
     onPathChanged,
     onRenameStateChange,
@@ -132,18 +140,13 @@
     if (row.isDir) {
       // #345: clicking a locked folder opens the password modal
       // instead of toggling expansion. On successful unlock we
-      // schedule the expand via onUnlocked — the tree refresh that
-      // fires from `encrypted_folders_changed` rebuilds this row
-      // with `encryption: "unlocked"`, and the expand we trigger
-      // here then descends into the now-visible children without
-      // requiring a second click.
+      // ensure-expand — a plain toggle would *collapse* the folder
+      // if it happened to be in `treeState.expanded` from before it
+      // was locked (locking doesn't prune the expanded set), forcing
+      // a second click to re-open it.
       if (isLocked) {
         openUnlockModal(row.path, row.name, async () => {
-          // The store swap + tree refresh run asynchronously after
-          // the unlock event; yield the microtask queue so the
-          // refreshed row is the one we toggle.
-          await Promise.resolve();
-          await onToggleExpand(row);
+          await onEnsureExpanded(row);
         });
         return;
       }
--- a/src/components/Sidebar/TreeRow.svelte
+++ b/src/components/Sidebar/TreeRow.svelte
@@ -56,9 +56,12 @@
     onToggleExpand: (row: FlatRow) => void | Promise<void>;
     /**
      * Guaranteed-expand (idempotent). Use when the caller needs the
-     * folder open regardless of its current expanded state — e.g. after
-     * unlock, or after creating a child, where a plain toggle would
-     * collapse an already-expanded folder.
+     * folder open regardless of its current expanded state. A plain
+     * `onToggleExpand` reads `treeState.expanded` and flips it, so
+     * calling it on an already-expanded folder collapses it — wrong
+     * for any flow whose intent is "make sure this folder is open"
+     * (unlock success; "New file/folder here" context menu entries
+     * that should leave the containing folder expanded).
      */
     onEnsureExpanded: (row: FlatRow) => void | Promise<void>;
     /** Tell the Sidebar the child list for this folder needs re-fetching. */
@@ -138,14 +141,24 @@
   function handleClick() {
     onSelect(row.path);
     if (row.isDir) {
-      // #345: clicking a locked folder opens the password modal
-      // instead of toggling expansion. On successful unlock we
-      // ensure-expand — a plain toggle would *collapse* the folder
-      // if it happened to be in `treeState.expanded` from before it
-      // was locked (locking doesn't prune the expanded set), forcing
-      // a second click to re-open it.
+      // #355: clicking a locked folder opens the password modal
+      // instead of toggling expansion. On successful unlock we:
+      //   1. Re-fetch the parent listing so the cached DirEntry for
+      //      this folder flips from `encryption: "locked"` to
+      //      `"unlocked"`. Without this step, flattenTree's
+      //      `if (entry.encryption === "locked") continue` would
+      //      skip children of an otherwise-expanded folder until
+      //      the async `encrypted_folders_changed` pulse lands —
+      //      yielding a brief "expanded but empty" flash.
+      //   2. Ensure-expand (idempotent). A plain toggle would
+      //      *collapse* a folder whose relPath is still in
+      //      `treeState.expanded` from before it was locked
+      //      (locking does not prune the expanded set), forcing
+      //      a second click to re-open it.
       if (isLocked) {
         openUnlockModal(row.path, row.name, async () => {
+          const parent = parentOf(row.path) ?? getVaultRoot();
+          if (parent) await onRefreshFolder(parent);
           await onEnsureExpanded(row);
         });
         return;
@@ -299,7 +312,7 @@
     closeContextMenu();
     try {
       const newPath = await createFile(row.path, "");
-      await onToggleExpand({ ...row, expanded: false });
+      await onEnsureExpanded(row);
       await onRefreshFolder(row.path);
       tabStore.openTab(newPath);
     } catch (e) {
@@ -313,7 +326,7 @@
     try {
       const newPath = await createFile(row.path, "Untitled.canvas");
       await writeFile(newPath, serializeCanvas(emptyCanvas()));
-      await onToggleExpand({ ...row, expanded: false });
+      await onEnsureExpanded(row);
       await onRefreshFolder(row.path);
       tabStore.openFileTab(newPath, "canvas");
     } catch (e) {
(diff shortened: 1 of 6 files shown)


src/components/Sidebar/TreeRow.svelte
```diff
@@ -54,6 +54,13 @@
     onSelect: (path: string) => void;
     onOpenFile: (path: string) => void;
     onToggleExpand: (row: FlatRow) => void | Promise<void>;
+    /**
+     * Guaranteed-expand (idempotent). Use when the caller needs the
+     * folder open regardless of its current expanded state — e.g. after
+     * unlock, or after creating a child, where a plain toggle would
```
herox215: The prop doc lists "after creating a child" as a use-case for `onEnsureExpanded`, but `handleNewFileHere`, `handleNewCanvasHere`, and `handleNewFolderHere` (lines 302/316/329) still call `onToggleExpand({ ...row, expanded: false })`. `onToggleExpand` decides expand-vs-collapse by reading `treeState.expanded.includes(row.relPath)` — the `expanded` field on the row object is irrelevant. If the folder is already in the expanded set, all three handlers will collapse it after creating the child, which is the same regression this PR fixes for unlock. Either migrate those three to `onEnsureExpanded` in this PR, or remove "after creating a child" from the doc so the comment does not falsely advertise a fix that does not exist.

— Aristotle

src/components/Sidebar/TreeRow.svelte
```diff
@@ -132,18 +140,13 @@
...
+      // was locked (locking doesn't prune the expanded set), forcing
+      // a second click to re-open it.
       if (isLocked) {
         openUnlockModal(row.path, row.name, async () => {
-          // The store swap + tree refresh run asynchronously after
-          // the unlock event; yield the microtask queue so the
-          // refreshed row is the one we toggle.
-          await Promise.resolve();
-          await onToggleExpand(row);
+          await onEnsureExpanded(row);
```
herox215, line 162: `onEnsureExpanded` is called from the unlock callback before `treeRefreshStore.requestRefresh()` fires (see `encryptedFoldersStore.ts` line 200: refresh is chained after `scheduleRefresh().then(...)`). `setExpanded` calls `loadFolder` when `!childrenLoaded`. The `treeRefreshStore` pulse then runs `invalidateAllFolders() + loadRoot()` in Sidebar, which wipes `folders` and triggers a second `loadFolder` for the same path concurrently. Two overlapping `listDirectory` IPC calls will race; the second sets `childrenLoaded = true` over whatever the first wrote. This is not a correctness bug today (both calls return the same data), but the old `await Promise.resolve()` at least let the `encrypted_folders_changed` event land and the row to be reconstructed with `encryption: "unlocked"` before expanding. Removing that yield means `onEnsureExpanded` is called while the row still carries the locked snapshot — does `loadFolder` proceed even when `DirEntry.encryption` is still `"locked"` on the backend-returned entry? If the backend requires an explicit unlock before `list_directory` returns children, this expansion fires too early and produces an empty folder until the second load from the refresh pulse completes.

— Aristotle

src/components/Sidebar/__tests__/TreeRowUnlockExpand.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    const row = container.querySelector(".vc-tree-row") as HTMLElement;
+    row.click();
+    await tick();
+
+    expect(openUnlockModal).toHaveBeenCalledTimes(1);
+    expect(openUnlockModal.mock.calls[0][0]).toBe(`${VAULT}/secret`);
+    expect(props.onToggleExpand).not.toHaveBeenCalled();
+  });
+
+  it("calls onEnsureExpanded (not onToggleExpand) on successful unlock", async () => {
```
herox215: The test only verifies that `onEnsureExpanded` is wired in place of `onToggleExpand` — both props are `vi.fn()` mocks, so `setExpanded` is never exercised. The critical regression path (folder in `treeState.expanded` → lock → unlock → must stay open) is not asserted. A test that actually catches that regression must render `Sidebar` with a pre-populated `treeState.expanded` containing the folder`s `relPath`, simulate lock then unlock, and assert `row.expanded === true` in the resulting flat rows — not just that a mock callback was invoked. As written this test would have passed on the original buggy code if `onToggleExpand` were simply renamed to `onEnsureExpanded`.

— Aristotle

src/components/Sidebar/__tests__/TreeRowUnlockExpand.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    onSelect: vi.fn(),
+    onOpenFile: vi.fn(),
+    onToggleExpand: vi.fn(),
+    onEnsureExpanded: vi.fn(),
+    onRefreshFolder: vi.fn(),
+    onPathChanged: vi.fn(),
+  };
+}
+
+describe("TreeRow locked-folder unlock flow (#354 follow-up)", () => {
```
herox215: Describe block references `#354 follow-up` but this PR fixes issue #355. Wrong ticket number.

— Aristotle

src/components/Sidebar/TreeRow.svelte
```diff
@@ -130,20 +141,25 @@
...
+      //      (locking does not prune the expanded set), forcing
+      //      a second click to re-open it.
       if (isLocked) {
         openUnlockModal(row.path, row.name, async () => {
-          // The store swap + tree refresh run asynchronously after
-          // the unlock event; yield the microtask queue so the
-          // refreshed row is the one we toggle.
-          await Promise.resolve();
-          await onToggleExpand(row);
+          const parent = parentOf(row.path) ?? getVaultRoot();
```
herox215, line 160: `parentOf` uses `lastIndexOf("/")`, but on Windows `entry_path.to_string_lossy()` in `tree.rs:166` emits native backslashes — `DirEntry.path` in JS therefore contains `\` separators on Windows. For a top-level encrypted folder (`C:\Vault\secret`) there is no `/` in the path, so `lastIndexOf("/")` returns -1 and the guard `if (i <= 0) return null` fires. `getVaultRoot()` is then the fallback, but only if it returns non-null. If the store is uninitialized that is also null, and `if (parent) await onRefreshFolder(parent)` silently skips the refresh — expanding into still-locked children. Fix: normalise `absPath` with `.replace(/\\/g, "/")` before the `lastIndexOf`, matching the pattern already used in `vaultPath.ts` and `openFileAsTab.ts`.

— Aristotle

src/components/Sidebar/TreeRow.svelte
```diff
@@ -296,7 +312,7 @@
     closeContextMenu();
     try {
       const newPath = await createFile(row.path, "");
-      await onToggleExpand({ ...row, expanded: false });
+      await onEnsureExpanded(row);
```
herox215: `onEnsureExpanded` fires before `onRefreshFolder` here, the opposite of the unlock path at line 161-162. For the "new file/folder here" context menu entries this order is harmless in practice (the folder is not encrypted, so `flattenTree` descends regardless), but it is a semantic inconsistency that makes the unlock ordering look like a deliberate exception when it is actually the invariant. Consider reversing to `refresh → ensure` here too, or add a comment explaining why the unlock order does not apply to this call site.

— Aristotle

src/components/Sidebar/__tests__/SidebarUnlockKeepsExpanded.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      return [];
+    });
+
+    render(Sidebar, { props: makeProps() });
+
+    // Drain initial mount + load.
+    for (let i = 0; i < 15; i += 1) { await Promise.resolve(); await tick(); }
+
+    // Click the locked folder → opens the unlock modal (not toggle).
+    const el = document.querySelector(
```
herox215: `.vc-tree-row` never receives a `data-path` attribute (confirmed at `TreeRow.svelte:488-495`), so `document.querySelector(.vc-tree-row[data-path="..."])` always returns `null` and the fallback text-match branch always runs. The primary selector is dead. Remove it and the comment that apologises for it; just use the text-match directly.

— Aristotle

PR #359, opened 2026-04-24
feat(encryption): auto-encrypt dropped files + unlock drain (#357)

### LINKED ISSUE #357: feat(encryption): auto-encrypt all file types dropped into encrypted folders
## Problem

Currently the transparent encryption layer only handles Markdown notes inside encrypted folders. If a user drops a binary file (image, video, PDF, audio, zip, …) into an encrypted folder via Finder / Explorer / drag-and-drop, the file lands on disk **unencrypted**. That silently defeats the purpose of the encrypted folder — sensitive attachments sit in plaintext right next to their encrypted notes.

## Goal

Any file placed into an encrypted folder — regardless of type, regardless of how it got there (Finder, Explorer, shell `mv`, drag-drop from another app, internal move, paste, attachment import) — must be encrypted on disk as long as the VaultCore app is running and the vault is open.

## Requirements

- **Type-agnostic**: works for images (png/jpg/webp/heic), video (mp4/mov/webm), audio (mp3/wav/m4a), PDFs, archives, office docs, arbitrary binaries. No allow-list of extensions.
- **Finder-triggered**: the existing FS watcher (`notify`) must pick up new/moved files in encrypted folders and encrypt them in place. The tool does not need to handle files added while it is *not* running — but the moment the vault opens, a reconciliation pass must sweep the encrypted folders and encrypt anything found in plaintext.
- **Transparent read/write**: opening an encrypted binary from inside VaultCore (e.g. an image embed in a note, or the file browser) must decrypt on read, same model as notes today.
- **Folder lock semantics preserved**: when a folder is locked, binary contents stay encrypted at rest just like notes. When unlocked in-session, reads are transparently decrypted.
- **No corruption of already-encrypted content**: the watcher must detect \"already encrypted\" via the existing header/magic and skip re-encryption.
- **Atomicity**: encrypt-in-place must be crash-safe (temp file + rename), so a crash mid-encryption never leaves a half-written file or loses the plaintext.
- **Performance**: large files (video, multi-GB archives) must stream through encryption — no full-file buffering in RAM. Must not block the UI thread.
- **Cross-platform**: macOS, Windows, Linux — consistent behavior across all three FS watchers.

## Out of scope

- Encryption of files added while the app is closed (covered by the reconciliation sweep on open, not by a background daemon).
- Thumbnail / preview caching strategy for encrypted media — separate ticket if needed.
- Mobile (not a supported platform yet).

## Open questions

- Should the reconciliation sweep on vault open block the UI or run in the background with a progress indicator? (Large vaults with many attachments could take a while.)
- How do we signal to the user that a file they just dropped is being encrypted (brief toast? inline badge in the file tree?) — or is silent correct?

herox215:
## Summary

Closes #357. Closes the plaintext-leak hole where binary files dropped into an encrypted folder (via Finder / Explorer / drag-drop) stayed plaintext on disk alongside their sealed notes.

Any file placed into an encrypted folder — regardless of type — now gets encrypted in place as soon as VaultCore sees it.

## Rationale

Encrypted folders today only seal `.md` notes. Drop a PNG into one via Finder, and the file lands unencrypted next to the ciphertext notes, silently defeating the "this folder is private" contract. This PR closes that hole end-to-end: the FS watcher encrypts unlocked drops in place; locked drops queue for seal-on-unlock; attachments saved from the editor (paste image) go through the same seal path.

Socrates review (plan stage) surfaced three blockers that are fixed here:

1. **Plaintext-read window during unlock drain** — `unlock_folder` now inserts the key, drains + seals every queued file *while the registry is still locked*, and only then flips the registry. A reader that observes `is_locked == false` is guaranteed to see a VCE1 header on disk.
2. **Watcher self-loop across platforms** — `write_ignore.should_ignore` now matches every path on a rename event, not just `paths[0]`. Covers Linux inotify (rename fires with primary=tmp + secondary=target) and macOS FSEvents (Create on target).
3. **`save_attachment` size-cap bypass** — the shared `ensure_size_cap` helper now runs on both the drop path and the attachment IPC, so a compromised frontend cannot bypass the 256 MiB buffered-encrypt cap.

Plus major architecture fixes: new `ManifestCache` keeps `find_enclosing_encrypted_root` out of the watcher hot path (was O(N events × JSON parse); now O(1)); new `WatcherContext` struct consolidates the growing list of `Arc` parameters on `spawn_watcher`; orchestrator function is layered (no direct `VaultState` coupling).

## Architecture

**Backend (Rust)** — new `encryption/pending_queue.rs` (FIFO cap 10k per root), new `encryption/drop_encrypt.rs` (orchestrator with outcome enum), new `ManifestCache` in `VaultState`, new `read_attachment_bytes` IPC, new `vault://encrypt_drop_progress` event, `save_attachment` routes through `maybe_encrypt_write`, `unlock_folder` drains the pending queue with the correct locked-until-drained ordering.

**Frontend (Svelte + TS)** — new `encryptionProgressStore` (mirrors `reindexStore`), new `EncryptionStatusbar.svelte` (bottom-center pill at `bottom: 48px` stacking above reindex), new `attachmentSource.ts` helper with `resolveAttachmentSrc` / `releaseAttachmentSrc`; all four image/embed render paths (ImagePreview, embedPlugin, reading-view markdownRenderer, CanvasRenderer) route through the helper. `ProgressBar.svelte` gains a `label` prop (vault-open now shows "Scanning and securing vault…"). CSP `img-src` adds `blob:`.

## Test plan

- [x] `cargo test --lib --no-default-features -- --test-threads=1` — 343 passed, 2 ignored, 0 failed (parallel run has a pre-existing flake on a global atomic counter unrelated to this PR).
- [x] `pnpm exec svelte-check --threshold error` — 0 errors.
- [x] `pnpm test` — 1243 passed (+4 new for `encryptionProgressStore`).
- [ ] UAT — manually drop a PNG into an unlocked encrypted folder, verify it seals on disk + the pill reports "1 file secured."
- [ ] UAT — drop into a locked folder, verify the queued warning, then unlock and verify the file is sealed.
- [ ] UAT — paste an image into a note inside an unlocked encrypted folder, verify it renders from the blob URL and the on-disk bytes start with `VCE1`.
- [ ] UAT — drop a > 256 MiB file and verify the error toast names the path and suggests moving it out.

## Out of scope

- **Streaming encryption** (VCE2 chunked framing) — tracked via the 256 MiB cap. Follow-up ticket will lift the cap by introducing chunked framing + a magic bump.
- **Queue persistence across restart** — in-memory today. A crash between drop and unlock loses the queued list (files remain plaintext on disk). UI warns the user at drop time so this is explicit.


src-tauri/src/encryption/drop_encrypt.rs
```diff
@@ -0,0 +1,405 @@
...
+        return Ok(EncryptOutcome::AlreadySealed);
+    }
+
+    // 5. Locked root? Queue and bail. The plaintext stays on disk — the
+    //    UI must warn the user.
+    let canon = CanonicalPath::assume_canonical(canonical_path.to_path_buf());
+    if deps.locked_paths.is_locked(&canon) {
+        deps.pending_queue.enqueue_for_root(root.clone(), canonical_path.to_path_buf())?;
+        return Ok(EncryptOutcome::Queued { root });
+    }
```
herox215, line 123: `write_ignore.record(canonical_path)` fires before `encrypt_file_in_place`, but `write_atomic` inside it creates a `.vce-tmp-<pid>-<hex>` file in the same watched directory. That tempfile path is never recorded in `write_ignore`. On Linux inotify, `notify-debouncer-full` emits a `Create` event for the tempfile itself; that event passes step 1b (unlocked root), reaches step 1c, and the orchestrator calls `encrypt_file_in_place_if_needed` on a path that may still exist at that instant. If the rename has not yet occurred the orchestrator will attempt to read and seal the tempfile — producing at best a redundant no-op (the tmpfile vanishes before `fs::metadata`), at worst a second encrypted write on a partial ciphertext blob. Fix: call `ign.record` for both the target AND the parent-dir-scoped tmp glob, or — simpler — record the target path and also set a sentinel on the parent directory path so any `.vce-tmp-*` sibling events in that directory are dropped.

— Aristotle

src-tauri/src/encryption/pending_queue.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    /// Enqueue `path` under `root`. Dedups exact-match paths (same path
+    /// bounced by two notify events coalesces to one entry). Enforces
+    /// `MAX_PENDING_PER_ROOT` with FIFO eviction and a `log::warn!` so
+    /// overflow is observable.
+    pub fn enqueue_for_root(&self, root: PathBuf, path: PathBuf) -> Result<(), VaultError> {
+        let mut g = self.by_root.write().map_err(|_| VaultError::LockPoisoned)?;
+        let queue = g.entry(root.clone()).or_default();
+        if queue.contains(&path) {
```
herox215: `queue.contains(&path)` on a `VecDeque` is O(n). With `MAX_PENDING_PER_ROOT` at 10 000, every enqueue scans up to 10k entries on the watcher hot path — the path where 16ms keystroke and 5ms incremental-update budgets already leave no slack. Use a `HashSet` as a companion to the `VecDeque` for O(1) dedup while preserving FIFO eviction order.

— Aristotle

src/components/Canvas/CanvasRenderer.svelte
```diff
@@ -325,10 +357,11 @@
                   {/if}
                 {:else if seg.kind === "image"}
                   {@const imgResolved = resolveTarget(seg.target)}
+                  {@const _cacheTick = cacheVersion}
                   {#if imgResolved && vaultPath}
                     <img
                       class="vc-canvas-inline-image"
-                      src={convertFileSrc(resolveVaultAbs(vaultPath, imgResolved))}
+                      src={_cacheTick >= 0 ? canvasImageSrc(resolveVaultAbs(vaultPath, imgResolved)) : ""}
```
herox215: `_cacheTick >= 0` is always `true` — `cacheVersion` is a non-negative integer and JavaScript has no unsigned overflow that could make it negative. Svelte 5's compiler performs constant-folding on `@const` bindings: it will see `_cacheTick >= 0` as a tautology and fold the expression to `canvasImageSrc(...)` with no reactive dependency on `_cacheTick` itself. The intended re-run when `cacheVersion` increments will not happen. The correct Svelte 5 pattern for forcing re-derivation when an external cache updates is to bind the reactive state directly in the expression: `src={canvasImageSrc(resolveVaultAbs(vaultPath, imgResolved), cacheVersion)}` and accept `cacheVersion` as an ignored parameter inside `canvasImageSrc`. Alternatively, model `encryptedUrlCache` as a `$state` Map so mutations are tracked without a version counter.

— Aristotle

src/components/Editor/ReadingView.svelte
```diff
@@ -85,6 +101,37 @@
...
+      if (!abs) continue;
+      img.removeAttribute("data-vc-encrypted-abs");
+      const result = resolveAttachmentSrc(abs);
+      if (typeof result === "string") {
+        img.src = result;
+      } else {
+        void result.then((resolved) => {
+          if (resolved) {
+            img.src = resolved;
+            blobUrls.push(resolved);
```
herox215: Stale-promise leak: `releaseAllBlobUrls()` clears `blobUrls` synchronously inside `load()`, but the `$effect` below fires on the same `html` assignment and calls `resolveAttachmentSrc` asynchronously. If `load()` is called a second time (tab switch, file watcher event) before those async promises resolve, `releaseAllBlobUrls()` runs again on the empty array — the previously-in-flight promises then resolve and `push` their blob URLs into `blobUrls` after it was cleared, making them permanently unrevoked. Fix: stamp each load with a generation counter in the async path and drop the resolved URL if the generation no longer matches.

— Aristotle

src-tauri/src/commands/encryption.rs
```diff
@@ -319,13 +414,16 @@ pub async fn lock_folder(
...
 ) -> Result<(), VaultError> {
-    let (folder_canon, _) = resolve_folder(&state, &path)?;
+    let (folder_canon, vault_root) = resolve_folder(&state, &path)?;
     // Register as locked BEFORE dropping the key so any in-flight
     // read-path race is resolved by the fail-closed gate — the reader
     // sees "locked" and returns PathLocked rather than using a stale
     // key snapshot.
     state.locked_paths.lock_root(folder_canon.clone())?;
     state.keyring.remove(&folder_canon)?;
+    if let Err(e) = state.manifest_cache.refresh_from_disk(&vault_root) {
```
herox215: `lock_folder` does not mutate the manifest — it only changes in-memory registry state. Refreshing the cache from disk here re-parses JSON that has not changed and returns the same roots the cache already holds, buying nothing except a spurious disk read. The comment justifies it as "easy to audit" symmetry, but the real invariant is "refresh after manifest write" and this is not a manifest write. Remove the refresh call here; add a targeted comment noting why it is intentionally absent.

— Aristotle

src-tauri/src/watcher.rs
```diff
@@ -628,6 +729,86 @@ fn dispatch_embed_delete_cmd(
...
+            serde_json::json!({
+                "inFlight": 0,
+                "total": 0,
+                "lastCompleted": canonical.display().to_string(),
+                "queued": true,
+                "error": null,
+            }),
+        );
+        // Diagnostic: the file is on disk as plaintext until next unlock.
+        log::info!(
```
herox215, line 743: `let _ = vault_path` silences an "unused variable" warning but signals that the parameter was added speculatively for "future scoping." Unused parameters that exist only to reserve a slot are dead code — the same concern the Boy Scout rule targets. Either use `vault_path` for an actual filter now (e.g. assert that the event path is inside the vault before enqueuing), or remove the parameter and add it back in the follow-up ticket that uses it.

— Aristotle

src/components/Editor/attachmentSource.ts
```diff
@@ -0,0 +1,114 @@
...
+import { encryptedPaths } from "../../store/encryptedFoldersStore";
+import { vaultStore } from "../../store/vaultStore";
+
+/**
+ * Return `true` when `abs` sits inside any manifest-registered
+ * encrypted folder of the currently-open vault. Kept synchronous and
+ * non-reactive so callers can use it inside `$derived` without
+ * wiring another subscription.
+ */
+export function isInsideEncryptedFolder(abs: string): boolean {
```
herox215, line 63: `isInsideEncryptedFolder` calls `get(vaultStore)` and `get(encryptedPaths)` on every invocation — two Svelte store snapshot reads outside any reactive context. This is called once per image node per render cycle in `canvasImageSrc`, which runs inside the `{#each}` loop over all canvas nodes. At 100+ image nodes on a canvas, that is 200+ `get()` calls per frame; `get()` on a derived store re-evaluates the derive chain each time when called outside a subscriber. Hoist both reads to the call site (or cache them for the duration of the render) rather than repeating them per-node.

— Aristotle

src-tauri/src/encryption/mod.rs
```diff
@@ -47,14 +57,97 @@ pub const SENTINEL_FILENAME: &str = ".vaultcore-folder-key-check";
...
+}
+
+impl ManifestCache {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Re-read the manifest and canonicalize every entry. Called by
+    /// `open_vault` (cold start), `encrypt_folder`, `unlock_folder`,
+    /// `lock_folder`, and `lock_all_folders` after their write. Orphaned
```
herox215: `lock_folder` and `lock_all_folders` are listed as callers of `refresh_from_disk`, but both were explicitly updated in this PR to skip the refresh (with clear comments explaining why). The doc contradicts the code; a future maintainer reading only the doc comment will expect a cache flush to happen on lock and introduce a real bug when they "fix" it.

Remove `, lock_folder, and lock_all_folders` from line 87.

— Aristotle

src/components/Editor/ImagePreview.svelte
```diff
@@ -1,14 +1,42 @@
...
+    // one so decrypted bytes don't accumulate in the browser's blob
+    // store across successive note switches.
+    if (typeof result === "string") {
+      releaseAttachmentSrc(revokeCurrent);
+      revokeCurrent = null;
+      src = result;
+    } else {
+      // Async (encrypted folder). Assign after the promise resolves;
+      // revoke the previous URL on success OR failure.
+      void result.then((resolved) => {
```
herox215, line 32: Stale-promise bug: when `abs` changes while a previous IPC call is still in flight, the old promise closure captures the outer `revokeCurrent` reference by closure — not by value. When the stale promise settles it calls `releaseAttachmentSrc(revokeCurrent)`, which at that point holds the URL assigned by the *newer* effect run, so the current valid URL is revoked and `src` is overwritten with the stale resolved bytes.

`ReadingView.svelte` fixes this with a generation token; the same fix is needed here: capture a local `let gen = ++effectGenCounter` at effect entry and bail in the `.then()` handler if `gen !== effectGenCounter`, revoking the stale-resolved URL on the way out.

— Aristotle

PR #361, opened 2026-04-24
feat(encryption): export decrypted copy from encrypted folder (#360)

### LINKED ISSUE #360: feat(encryption): export decrypted file from encrypted folder
## Problem

An encrypted folder holds files as VCE1 ciphertext on disk. Dragging a file out via Finder / Explorer / \`cp\` copies the raw ciphertext bytes — useless for anything outside VaultCore. There is no supported path to extract a decrypted copy of a single attachment (image, PDF, video) for use in an external app.

Follow-up from #357 UAT: the \"in\" direction is now covered (auto-encrypt on drop); the \"out\" direction is missing.

## Goal

Give the user a first-class in-app action to export a decrypted copy of any single file that lives inside an **unlocked** encrypted folder. The source file stays sealed on disk; the export is a write to a user-chosen destination outside the encrypted folder.

## Proposed flow

1. Right-click on a file inside an unlocked encrypted folder (Sidebar tree, editor tab context menu, image-preview tab).
2. Context menu entry: **Export decrypted copy…**
3. Native save dialog opens (Tauri \`save\` dialog), prefilled with the original filename.
4. Backend IPC (\`export_decrypted_file(source_path, dest_path)\`):
   - \`ensure_inside_vault(source)\` + \`source\` must sit inside an encrypted root that is currently unlocked (else \`PathLocked\`).
   - \`dest\` must sit **outside** every encrypted root (else \`PermissionDenied\` — exporting into another encrypted folder should be a proper copy+re-encrypt, not this command).
   - \`dest\` may be outside the vault (that's the whole point — user chooses a destination the OS and other apps can read).
   - Read ciphertext → \`maybe_decrypt_read\` → atomic write to \`dest\`.
   - Enforce the same size cap as the inline encrypt path (256 MiB MVP) so a multi-GB video fails with \`PayloadTooLarge\` instead of OOM.
5. Toast on success: \"Exported to <dest>\". Toast on error: show the VaultError \`message\`.

## Out of scope (separate tickets)

- Drag-out from Sidebar into a plain vault folder (inner-vault move-out) — needs \`move_file\` to decrypt-on-move when crossing an encrypted-root boundary.
- Bulk \"Decrypt folder\" reverse of the encrypt flow (whole-folder password-prompt → walk → decrypt all → manifest remove).
- Encrypted binary thumbnails / previews in the Sidebar (the export action works even without previews — user sees the filename).

## UX constraints (keyboard-first, minimal)

- Context-menu entry lives alongside Rename / Delete; no modal beyond the native save dialog.
- Not exposed for plain-vault files (would be redundant — they are already plaintext).
- Not exposed for files in a **locked** encrypted folder (user unlocks first; we don't prompt for a password from the export action).
- No new global shortcut in MVP.

## Security

- Source path canonicalized + scope-checked exactly like \`read_file\`.
- Destination is written via the existing atomic-write helper so a crash mid-export never leaves a half-written plaintext file at \`dest\`.
- Size cap rejects runaway payloads with an actionable error.
- No auto-export: every export is an explicit user-driven action.

## Acceptance criteria

- Right-click on a PNG in an unlocked encrypted folder → \"Export decrypted copy…\" → pick \`~/Desktop/photo.png\` → file at that path is plaintext PNG bytes, source unchanged, \`VCE1\` still on disk.
- Right-click on the same file while the folder is **locked** → menu entry is hidden OR shows error \"Unlock the folder first\".
- Attempting to export into another encrypted folder fails cleanly with an actionable toast.
- Tests: Rust unit test for the IPC body (unlocked-source + plain-dest → plaintext at dest); locked-source returns \`PathLocked\`; dest-in-encrypted returns \`PermissionDenied\`; over-cap returns \`PayloadTooLarge\`.

herox215:
## Summary

Closes #360. Follow-up to #357. Adds a first-class right-click action to extract a plaintext copy of a single file out of an unlocked encrypted folder via the native save dialog. Closes the UX hole where dragging via Finder/Explorer gave raw VCE1 ciphertext bytes.

## Rationale

#357 shipped "IN" (auto-encrypt on drop); this PR ships "OUT" (export decrypted). Scope is deliberately minimal: single file, right-click only, no drag-drop out, no whole-folder bulk decrypt. Those stay out of scope per the ticket.

## Architecture

- **Command placed in `commands/encryption.rs`** (not `files.rs`) — this is an encryption-domain operation and belongs alongside the other encryption IPCs. (Socrates finding.)
- **No `write_atomic`** — that helper writes a `.vce-tmp-*` next to the target, which for this command is typically outside the vault. That would leak plaintext outside the vault during the rename window. Plain `fs::write` is correct; export is user-initiated and retry-safe.
- **No redundant size-cap re-check** — every file under an encrypted root passed the cap at encrypt time; ciphertexts beyond that are tamper cases Poly1305 catches. (Socrates finding: the original plan's ciphertext-size comparison was directionally wrong.)
- **Used existing `pickSavePath` helper** — frontend stays inside the `ipc/commands.ts` abstraction rather than reaching directly for the plugin-dialog import. (Socrates finding.)
- **Shared helper `isInsideUnlockedEncryptedFolder`** added alongside `isInsideEncryptedFolder` in `attachmentSource.ts`, backed by a new `unlockedPaths` derived store — future callers that need the locked/unlocked distinction reuse the same microtask-cached infrastructure.

## Refusal contract

- Source outside vault → `PermissionDenied` (T-02 guard).
- Source not in encrypted folder → `PermissionDenied` with explicit message.
- Enclosing root locked → `PathLocked`.
- Dest inside any encrypted root → `PermissionDenied` (prevents smuggling plaintext past another folder's contract).
- Dest parent missing → `FileNotFound` (before any read/decrypt work, source untouched).
- Dest inside vault but outside every encrypted root → **allowed** (documented: the watcher will index the plaintext copy via its normal Create → AddFile path; expected behavior).

## Test plan

- [x] `cargo test --lib --no-default-features -- --test-threads=1` — 357 pass, 2 ignored, 0 fail (+7 new in `tests/export_decrypted.rs`).
- [x] `pnpm typecheck` + `pnpm exec svelte-check --threshold error` — clean.
- [x] `pnpm test -- TreeRowExportDecrypted` — 1254 pass (+6 new).
- [ ] UAT — right-click a PNG inside an unlocked encrypted folder → "Export decrypted copy…" → pick `~/Desktop/photo.png` → file at dest is plaintext; source still has VCE1 header.
- [ ] UAT — right-click same file while folder is locked → menu entry hidden.
- [ ] UAT — try to pick a destination inside another encrypted folder → error toast.
- [ ] UAT — cancel the save dialog → no toast, no side effect.

## Out of scope

- Drag-out from Sidebar into a plain vault folder (inner-vault move-out) — separate ticket if users ask.
- Bulk "Decrypt folder" reverse of encrypt — separate ticket.
- Export action from editor tab / image preview context menus — Sidebar only for MVP.

### DIFF
--- a/src-tauri/src/commands/encryption.rs
+++ b/src-tauri/src/commands/encryption.rs
@@ -447,6 +447,194 @@ pub async fn lock_all_folders(
     Ok(())
 }
 
+// ── export_decrypted_file (#360) ─────────────────────────────────────────────
+
+/// Test-accessible body for `export_decrypted_file`. Splits the Tauri
+/// `State` extraction from the logic so unit tests can hit the same
+/// contract without spinning up a runtime.
+///
+/// Contract:
+/// - `source` must sit inside the current vault AND inside an encrypted
+///   folder AND that folder must currently be UNLOCKED. Else: the
+///   corresponding `VaultError` without ever reading bytes.
+/// - `dest`'s parent must exist and canonicalize, and the prospective
+///   final path must NOT sit inside any encrypted root (exporting into
+///   another encrypted root would write unframed plaintext adjacent to
+///   ciphertext, silently breaking that folder's contract).
+/// - `dest` MAY live outside the vault — that is the whole point.
+///
+/// No atomic-write helper here. `write_atomic` places its `.vce-tmp-*`
+/// file next to the *destination*, which for this command is usually
+/// outside the vault (e.g. `~/Desktop/`). That tempfile would leak
+/// plaintext bytes outside the vault for the duration of the rename,
+/// violating the "plaintext never leaves unless explicit export"
+/// contract. The export is user-initiated and one-shot: a crash in the
+/// middle is a retry, not a corruption scenario for any vault state.
+/// Plain `fs::write` is correct here.
+///
+/// Size cap intentionally not re-checked: every sealed file under an
+/// encrypted root was already bounded by `MAX_INLINE_ENCRYPT_BYTES` at
+/// encrypt time. A ciphertext exceeding that cap + HEADER/TAG overhead
+/// means the file was tampered with — `decrypt_bytes` will fail the
+/// Poly1305 tag and surface `WrongPassword` / `CryptoError`, which is
+/// the correct defense.
+pub fn export_decrypted_file_impl(
+    state: &VaultState,
+    source: String,
+    dest: String,
+) -> Result<(), VaultError> {
+    // 1. Source lives inside the open vault.
+    let source_path = PathBuf::from(&source);
+    let source_canonical = {
+        let guard = state
+            .current_vault
+            .lock()
+            .map_err(|_| VaultError::LockPoisoned)?;
+        let vault = guard.as_ref().ok_or_else(|| VaultError::VaultUnavailable {
+            path: source.clone(),
+        })?;
+        let canon = std::fs::canonicalize(&source_path).map_err(|e| match e.kind() {
+            std::io::ErrorKind::NotFound => VaultError::FileNotFound { path: source.clone() },
+            std::io::ErrorKind::PermissionDenied => {
+                VaultError::PermissionDenied { path: source.clone() }
+            }
+            _ => VaultError::Io(e),
+        })?;
+        if !canon.starts_with(vault.as_path()) {
+            return Err(VaultError::PermissionDenied {
+                path: canon.display().to_string(),
+            });
+        }
+        canon
+    };
+
+    // 2. Source must be inside an encrypted root.
+    let vault_root = {
+        let guard = state
+            .current_vault
+            .lock()
+            .map_err(|_| VaultError::LockPoisoned)?;
+        guard.as_ref().cloned().ok_or_else(|| VaultError::VaultUnavailable {
+            path: source.clone(),
+        })?
+    };
+    let src_enc_root = crate::encryption::find_enclosing_encrypted_root_cached(
+        &state.manifest_cache,
+        &vault_root,
+        &source_canonical,
+    )?
+    .ok_or_else(|| VaultError::PermissionDenied {
+        path: format!(
+            "{} is not inside an encrypted folder; nothing to decrypt",
+            source_canonical.display()
+        ),
+    })?;
+
+    // 3. That encrypted root must be unlocked (key present in keyring).
+    //    `keyring.key_clone(root)` returns `None` when the root is
+    //    currently locked — we surface this as `PathLocked` so the
+    //    frontend toast can tell the user to unlock and retry.
+    if state.keyring.key_clone(&src_enc_root)?.is_none() {
+        return Err(VaultError::PathLocked {
+            path: source_canonical.display().to_string(),
+        });
+    }
+
+    // 4. Resolve the destination. The file itself does not exist yet,
+    //    so canonicalize the PARENT and rebuild the final path.
+    let dest_path = PathBuf::from(&dest);
+    let dest_parent = dest_path
+        .parent()
+        .ok_or_else(|| VaultError::PermissionDenied { path: dest.clone() })?;
+    let dest_file_name = dest_path
+        .file_name()
+        .ok_or_else(|| VaultError::PermissionDenied { path: dest.clone() })?
+        .to_owned();
+    let dest_parent_canonical = std::fs::canonicalize(dest_parent).map_err(|e| match e.kind() {
+        std::io::ErrorKind::NotFound => VaultError::FileNotFound {
+            path: dest_parent.display().to_string(),
+        },
+        std::io::ErrorKind::PermissionDenied => VaultError::PermissionDenied {
+            path: dest_parent.display().to_string(),
+        },
+        _ => VaultError::Io(e),
+    })?;
+    let dest_final = dest_parent_canonical.join(&dest_file_name);
+
+    // 5. Dest MUST NOT land inside any encrypted root. An export target
+    //    inside another encrypted folder would drop unframed plaintext
+    //    next to sealed files, silently breaking that folder's contract.
+    //    Use the dest PARENT for the enclosing-root check: the file
+    //    does not exist yet so we cannot canonicalize the full path,
+    //    and a child of a canonical parent inherits the parent's
+    //    enclosing-root status.
+    if crate::encryption::find_enclosing_encrypted_root_cached(
+        &state.manifest_cache,
+        &vault_root,
+        &dest_parent_canonical,
+    )?
+    .is_some()
+    {
+        return Err(VaultError::PermissionDenied {
+            path: format!(
+                "destination {} is inside an encrypted folder; pick a plain folder instead",
+                dest_final.display()
+            ),
+        });
+    }
+
+    // 6. Read + decrypt. `maybe_decrypt_read` is the same helper
+    //    `read_file` / `read_attachment_bytes` use; it returns plaintext
+    //    when the source sits in an unlocked encrypted root and passes
+    //    bytes through otherwise. We have already established (Steps 2
+    //    + 3) that this source qualifies for the decrypt path.
+    let ciphertext = std::fs::read(&source_canonical).map_err(|e| match e.kind() {
+        std::io::ErrorKind::NotFound => VaultError::FileNotFound { path: source.clone() },
+        std::io::ErrorKind::PermissionDenied => {
+            VaultError::PermissionDenied { path: source.clone() }
+        }
+        _ => VaultError::Io(e),
+    })?;
+    let plaintext =
+        crate::encryption::maybe_decrypt_read(state, &source_canonical, ciphertext)?;
+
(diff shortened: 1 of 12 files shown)


src-tauri/src/commands/encryption.rs
```diff
@@ -447,6 +447,194 @@ pub async fn lock_all_folders(
...
+    //    bytes through otherwise. We have already established (Steps 2
+    //    + 3) that this source qualifies for the decrypt path.
+    let ciphertext = std::fs::read(&source_canonical).map_err(|e| match e.kind() {
+        std::io::ErrorKind::NotFound => VaultError::FileNotFound { path: source.clone() },
+        std::io::ErrorKind::PermissionDenied => {
+            VaultError::PermissionDenied { path: source.clone() }
+        }
+        _ => VaultError::Io(e),
+    })?;
+    let plaintext =
```
herox215, line 641: `maybe_decrypt_read` silently returns raw bytes when the source file has no VCE1 magic header (the "plain file that leaked into the folder" toleration documented in `mod.rs:207`). In the read/editor path that passthrough is acceptable — you render raw bytes and move on. Here it is a bug: `export_decrypted_file_impl` will write those raw bytes to `dest_final`, then toast "Exported decrypted copy to photo.png", giving the user false confidence they received plaintext. The fix: check `bytes.starts_with(file_format::MAGIC)` before calling `maybe_decrypt_read` and return `VaultError::CryptoError` (or a dedicated variant) when the magic is absent — same as the guard `drop_encrypt` applies when it detects a partially-sealed file.

— Aristotle

src-tauri/src/commands/encryption.rs
```diff
@@ -447,6 +447,194 @@ pub async fn lock_all_folders(
...
+        canon
+    };
+
+    // 2. Source must be inside an encrypted root.
+    let vault_root = {
+        let guard = state
+            .current_vault
+            .lock()
+            .map_err(|_| VaultError::LockPoisoned)?;
+        guard.as_ref().cloned().ok_or_else(|| VaultError::VaultUnavailable {
```
herox215: TOCTOU race: `current_vault` is locked twice — once in step 1 (lines 489-509) to validate `source_canonical`, then released, then locked again here to clone `vault_root`. A concurrent `open_vault` between the two acquisitions can swap the vault to a different root. `source_canonical` was verified against vault A; `vault_root` and `manifest_cache` here reflect vault B. The `find_enclosing_encrypted_root_cached` call then checks the source against vault B's manifest, which may erroneously return `None` (leaking to `PermissionDenied`) or a wrong encrypted root. Merge steps 1 and 2 into a single `lock()` scope that canonicalizes, validates `starts_with`, and clones `vault_root` before releasing — exactly as `ensure_inside_vault` in `files.rs:53-76` does.

— Aristotle

src-tauri/src/commands/encryption.rs
```diff
@@ -447,6 +447,194 @@ pub async fn lock_all_folders(
...
+            }
+            _ => VaultError::Io(e),
+        })?;
+        if !canon.starts_with(vault.as_path()) {
+            return Err(VaultError::PermissionDenied {
+                path: canon.display().to_string(),
+            });
+        }
+        canon
+    };
```
herox215: `ensure_inside_vault` in `files.rs` is a crate-private helper that already does canonicalize + `starts_with` vault check + `drop(guard)` in one function. This impl duplicates that pattern inline rather than calling the existing helper. The reason given in the PR description is that `export_decrypted_file_impl` needs to also extract `vault_root` (the `PathBuf` clone) from the same lock scope — but `ensure_inside_vault` could be extended with a variant that returns `(canonical, vault_root)`, or the two calls could be collapsed into a single inline block. As written, the pattern diverges from every other command in the module, which means future security auditors checking the T-02 mitigation coverage will miss this command. At minimum, move `vault_root` extraction into the step 1 lock scope so there is a single acquisition.

— Aristotle

src-tauri/src/tests/export_decrypted.rs
```diff
@@ -0,0 +1,236 @@
...
+        let mut k = Zeroizing::new([0u8; 32]);
+        k.copy_from_slice(key.as_slice());
+        state.keyring.insert(enc_root_canon.clone(), k).unwrap();
+        // Default VaultState already has empty locked_paths — unlocked state.
+    } else {
+        state.locked_paths.lock_root(enc_root_canon.clone()).unwrap();
+    }
+
+    let _ = enc_root_canon; // enc_root_canon is kept as a debug aid
+    SealedVault {
```
herox215, line 76: `let _ = enc_root_canon;` does not keep the value alive — it was last used at line 285 and is already dead here. This line only suppresses the unused-variable lint. The comment "kept as a debug aid" will mislead the next reader. Fix: declare it as `_enc_root_canon` at the binding site instead, or remove both the binding and the suppression if it is genuinely unused after the if-block.

— Aristotle

src/components/Sidebar/__tests__/TreeRowExportDecrypted.test.ts
```diff
@@ -0,0 +1,215 @@
...
+      '[data-testid="context-export-decrypted"]',
+    ) as HTMLButtonElement;
+    await fireEvent.click(entry);
+    await Promise.resolve();
+    await tick();
+
+    expect(pickSavePath).toHaveBeenCalledTimes(1);
+    expect(exportDecryptedFile).not.toHaveBeenCalled();
+  });
+});
```
herox215, line 254: The test suite covers success, cancel, and menu-visibility cases, but not the error path: backend returns `PathLocked` or `PermissionDenied` → `catch(e)` block → error toast via `vaultErrorCopy`. Without a test for this, a regression that swallows the error or toasts the wrong variant will go undetected. Add one test that mocks `exportDecryptedFile` to reject with a `PathLocked` error and asserts the error toast fires.

— Aristotle

src/components/Sidebar/TreeRow.svelte
```diff
@@ -352,6 +362,29 @@
...
+  // usual `vaultErrorCopy` path so the user can unlock and retry.
+  async function handleExportDecrypted() {
+    closeContextMenu();
+    const picked = await pickSavePath(row.name);
+    if (!picked) return; // user cancelled — silent, matches other save flows
+    try {
+      await exportDecryptedFile(row.path, picked);
+      const filename = picked.split(/[/\\]/).pop() ?? picked;
+      toastStore.push({
+        variant: "info",
```
herox215: `info` understates the security significance of this action. Exporting plaintext is the highest-risk encryption op in the app; the user needs an unmistakable signal that the output file is unprotected. Use `variant: "warning"` with copy that names the consequence, e.g. "Plaintext copy written to photo.png — file is not encrypted."

— Aristotle

src-tauri/src/commands/encryption.rs
```diff
@@ -447,6 +447,215 @@ pub async fn lock_all_folders(
...
+    let source_canonical = crate::commands::files::ensure_inside_vault(
+        state,
+        &PathBuf::from(&source),
+    )?;
+
+    // 2. Snapshot `vault_root` ONCE. A concurrent `open_vault` that
+    //    swaps the vault between the scope check and the manifest
+    //    lookup would otherwise make us consult the wrong vault's
+    //    manifest. Single-lock semantics close the TOCTOU window.
+    let vault_root = {
```
herox215, line 496: `ensure_inside_vault` acquires `state.current_vault.lock()` and drops it before returning, then this block re-acquires it. The comment says "Single-lock semantics close the TOCTOU window" — but two sequential acquisitions are not single-lock. A concurrent `open_vault` can still swap `current_vault` between the `ensure_inside_vault` drop and this re-acquisition, giving `source_canonical` a path validated against vault A while `vault_root` comes from vault B. Fix: extract the vault root during `ensure_inside_vault` (return a `(PathBuf, PathBuf)` tuple, or add a `vault_root(state)` snapshot before calling `ensure_inside_vault`) so `current_vault` is read exactly once across both checks.

— Aristotle

src/components/Sidebar/TreeRow.svelte
```diff
@@ -352,6 +362,32 @@
...
+    const picked = await pickSavePath(row.name);
+    if (!picked) return; // user cancelled — silent, matches other save flows
+    try {
+      await exportDecryptedFile(row.path, picked);
+      const filename = picked.split(/[/\\]/).pop() ?? picked;
+      // #360 — toast copy deliberately names the file as UNENCRYPTED
+      // because that is what the user just produced. "Decrypted copy"
+      // softens a security-relevant action; make it explicit.
+      toastStore.push({
+        variant: "info",
```
herox215: This is a security-relevant action — the user just created an unencrypted file — and `"info"` (neutral blue) does not communicate that weight. Use `"warning"` so the toast is visually distinct from routine confirmations. The copy update alone does not help users who skim toasts by color.

— Aristotle

PR #363, opened 2026-04-24
feat(canvas): multi-shape text nodes with inline shape picker (#362)

### LINKED ISSUE #362: feat(canvas): multi-shape text nodes with shape picker in context menu
## Summary

Canvas text nodes currently only render as rounded rectangles. This ticket adds support for **5 shapes** selectable per node, with matching connection-point geometry. Selection happens via an inline shape picker inside the existing canvas context menu — both when creating a new node and when changing an existing one.

## Motivation

Users want to visually distinguish nodes on a canvas (e.g. flowchart metaphors: decision = diamond, process = rectangle, highlight = triangle). The current single-shape restriction forces all semantic distinction into text content.

### Shape set (5 shapes)

| # | Name | Rendering | Anchors |
|---|------|-----------|---------|
| 1 | **Rounded rectangle** (default) | `border-radius: 6px` | 4 side midpoints |
| 2 | **Rectangle** | `border-radius: 0` | 4 side midpoints |
| 3 | **Ellipse** | `border-radius: 50%` | 4 cardinal points (same coords as rect) |
| 4 | **Diamond** | `clip-path: polygon(50% 0%, 100% 50%, 50% 100%, 0% 50%)` | 4 vertices (same coords as rect) |
| 5 | **Triangle** | `clip-path: polygon(50% 0%, 100% 100%, 0% 100%)` | **3 edge midpoints** (left, right, bottom — no top) |

Rejected: hexagon, parallelogram, pill, cylinder (complexity vs. value, or visually redundant).

### Entry points

**A. Existing node — right-click → \"Change shape…\"**
- Inserted after \"Copy text\" in the text-node context menu.
- Clicking \"Change shape…\" expands the picker inline below the item.
- Currently selected shape is highlighted (accent background + border).

**B. Empty canvas right-click — \"Add text node ▾\"**
- Menu item gains expandable affordance.
- Enter / single click without expanding → creates default shape (rounded rectangle). Fast path preserved.
- Expanding shows the same shape picker; picking a shape creates the node at the cursor position with that shape.

**C. Double-click on empty canvas**
- Unchanged. Always creates default shape (rounded rectangle). No picker.

### Shape picker layout

- **Vertical list inside the existing context menu panel**, one shape per row.
- Each row: `[20×20 SVG icon]  [shape name]`, icon left of label, single horizontal row.
- Row padding: `6px 12px` (matches `.vc-context-item`).
- Hover/focus: `background: var(--color-accent-bg)`, text becomes `var(--color-accent)`.
- Selected (in \"Change shape…\" path): `background: var(--color-accent-bg)`, icon stroke + text use `var(--color-accent)`.
- Preview icons: hand-authored inline SVGs, 20×20, `stroke: currentColor`, `fill: none`, `stroke-width: 1.5`. No Lucide dependency for shape previews.
- Expand/collapse indicator: `▾` / `▴` suffix on the trigger item.

### Keyboard

- `ArrowDown` / `ArrowUp` navigate menu items as today.
- `Enter` or `ArrowRight` on \"Change shape…\" / \"Add text node ▾\" expands the picker and focuses the first shape row.
- `ArrowDown` / `ArrowUp` inside the picker navigate shapes; `Enter` / `Space` confirms.
- `Escape` or `ArrowLeft` from the first picker item collapses back to the parent item.
- `Escape` on the outer menu closes it (existing behavior). Inline picker state resets on menu close.

### Connection-point geometry

- **Rectangle, Rounded Rectangle, Ellipse, Diamond**: 4 anchors at top/right/bottom/left — same coordinates as today's `anchorPoint()`. No math change required.
- **Triangle**: 3 anchors — no `top`. Edge midpoints of the triangle's three sides:
  - Left: `(x + w*0.25, y + h*0.5)`
  - Right: `(x + w*0.75, y + h*0.5)`
  - Bottom: `(x + w*0.5, y + h)`
- `autoSides()` contract: when either endpoint is a triangle and the axis-dominant side is `top`, remap to `left` (if `dx >= 0`) or `right` (if `dx < 0`).

### Data model

- Store shape as `node.extra.shape` (string literal union: `\"rectangle\" | \"rounded-rectangle\" | \"ellipse\" | \"diamond\" | \"triangle\"`).
- Round-trips via the existing `extra` bag in `CanvasTextNode`. No schema change, no Obsidian compatibility break (unknown field ignored by Obsidian).
- **Default when `extra.shape` is absent: `\"rounded-rectangle\"`** — keeps existing canvases visually identical (preserves today's `border-radius: 6px` look).

### Rendering

- Apply shape via CSS only. No Canvas2D / per-node SVG.
- `border-radius` for rectangle / rounded-rectangle / ellipse.
- `clip-path` for diamond / triangle. Set `overflow: hidden` on the node for clipped shapes.
- Triangle text clipping: user sizes the node to fit text. No enforced padding.
- `anchorPoint()` in `geometry.ts` gains an optional `shape` parameter; remains O(1) pure.
- New `sidesForShape(shape): readonly CanvasSide[]` — returns 4 sides for rect/rounded/ellipse/diamond, 3 sides (`left`, `right`, `bottom`) for triangle.
- Hover handle rendering in `CanvasRenderer.svelte` calls `sidesForShape()` to decide which handles to render.

### Component extraction

- New: `src/components/Canvas/CanvasShapePicker.svelte` — stateless, props: `currentShape?: NodeShape`, `onPick: (shape: NodeShape) => void`. Used in both context-menu entry points. `role=\"radiogroup\"` with `role=\"radio\"` children.

## Constraints

- Default shape for nodes without `extra.shape`: **rounded-rectangle** (zero visual regression for existing canvases).
- No shape support for file, link, group, or unknown node types. Text nodes only.
- No changes to the JSON Canvas 1.0 wire format beyond using `extra` (already implemented for unknown-field round-tripping).
- Keystroke / frame budget: `anchorPoint()` and `sidesForShape()` must stay O(1) pure. They run per-node per-render-frame in `CanvasRenderer.svelte`.
- `clip-path` is GPU-composited; acceptable for target platforms.
- Accessibility: shape picker rows expose `role=\"radio\"` + `aria-checked` + `aria-label`. Trigger items expose `aria-expanded` + `aria-haspopup=\"true\"`. `:focus-visible` outline matches existing `.vc-context-item`.

## Out of scope

- Shape support on file, link, or group nodes.
- Shapes beyond the listed five (hexagon, parallelogram, pill, cylinder, star, etc.).
- Custom border width, opacity, or font (wider \"appearance\" scope).
- Animated shape transitions.
- Exporting the `shape` field into Obsidian (stays a VaultCore extension).
- Command palette entry to open the shape picker.
- \"Change color\" on text nodes (text nodes have no color today; not in scope here).

## Affected files

- `src/lib/canvas/types.ts` — `NodeShape` type, extend `CanvasTextNode` conceptually (stored in `extra`).
- `src/lib/canvas/geometry.ts` — `anchorPoint(node, side, shape?)`, `sidesForShape(shape)`, `autoSides()` triangle remap.
- `src/lib/canvas/edgeResolver.ts` — pass shape to `anchorPoint()`.
- `src/components/Canvas/CanvasRenderer.svelte` — apply shape CSS, render handles via `sidesForShape()`.
- `src/components/Canvas/CanvasView.svelte` — context menu entries, inline picker mount, `addTextNodeAt()` shape parameter.
- `src/components/Canvas/CanvasShapePicker.svelte` — new component.

## Test plan

- [ ] Unit: `anchorPoint()` returns correct coordinates for each of the 5 shapes across all sides.
- [ ] Unit: `sidesForShape()` returns 4 sides for rect/rounded/ellipse/diamond, 3 for triangle.
- [ ] Unit: `autoSides()` remaps `top` to `left`/`right` when either endpoint is a triangle.
- [ ] Unit: canvas parse/serialize round-trips `extra.shape` for all 5 values and for `undefined`.
- [ ] E2E: double-click empty canvas → creates rounded-rectangle.
- [ ] E2E: right-click empty canvas → \"Add text node ▾\" → expand → pick \"Triangle\" → node created with 3 handles, no top handle.
- [ ] E2E: right-click existing node → \"Change shape…\" → pick \"Ellipse\" → node re-renders as ellipse, edges re-anchor.
- [ ] E2E: keyboard — open context menu, navigate to \"Change shape…\", `Enter` expands, arrow-keys navigate shapes, `Enter` confirms.
- [ ] E2E: existing canvas file without `extra.shape` loads with all nodes as rounded-rectangles (no visual regression).
- [ ] Visual: each of the 5 shapes renders correctly at small (100×40), default (250×60), and large (400×200) sizes.
- [ ] Visual: edges attach cleanly to each shape's anchor points.

herox215:
Closes #362.

## Summary

- Adds 5 shapes for canvas text nodes: rounded-rectangle (default), rectangle, ellipse, diamond, triangle.
- Shape picker opens inline inside the existing context menu via two entry points:
  - Empty-space menu: **Add text node ▾** — expands picker, pick a shape to create a node at the right-clicked world coords.
  - Text-node menu: **Change shape…** — expands picker with the current shape pre-selected; picking rewrites `node.shape`.
- Double-click on empty canvas is unchanged: always creates a rounded-rectangle without going through the picker.
- Triangles expose only 3 anchors (left/right/bottom — the apex is not a valid terminus). A new `remapSideForShape` helper routes a `top` side on a triangle to whichever face points at the other endpoint; applied in both `autoSides` and `resolveEdges` so user-set sides surviving from before a shape change also resolve correctly.
- `shape` is a typed optional field on `CanvasTextNode`, parsed/serialised by `parse.ts` as a first-class field. Default shape is never written to disk — pre-#362 canvases stay byte-identical.

## Architecture notes

- **Why a typed `shape?: CanvasShape` field instead of the `extra` bag:** `extra` is the passthrough bag for *unknown* fields. Promoting a field the app actively reads/writes into that bag would spread `extra?.shape as string` checks across the canvas subsystem. Keeping the interpretation boundary at parse/serialize (where it belongs) lets every downstream caller use plain `node.shape` / `readShape(node)`.
- **Why a shape underlay `<div>` instead of moving handles to a sibling overlay:** `clip-path` (diamond/triangle) clips children, which would hide the hover handles + resize grabber. Adding a single `.vc-canvas-node-shape` child that carries the clip/border-radius keeps the outer node div `overflow: visible`, so handles stay interactive on every shape without restructuring the handle DOM or breaking `handleAtPoint`'s `closest("[data-node-id]")` lookup.
- **Rounded-rectangle as default:** preserves today's `border-radius: 6px` visual for nodes without a `shape` field. Zero visual regression on existing canvases.

## Test plan

Vitest (added/extended):
- `src/lib/canvas/__tests__/geometry.test.ts` — shape-aware `anchorPoint` (triangle edge midpoints), `sidesForShape`, `remapSideForShape`, `autoSides` triangle remap (single-triangle + both-triangle).
- `src/lib/canvas/__tests__/edgeResolver.test.ts` — explicit `fromSide: \"top\"` / `toSide: \"top\"` on a triangle remaps to a valid side in `resolveEdges`.
- `src/lib/canvas/__tests__/parse.test.ts` — shape roundtrip (parse as typed field; unknown values drop; default omitted on serialise; non-default preserved alongside other `extra` keys).
- `src/components/Canvas/__tests__/CanvasViewContextMenu.test.ts` — empty-space `Add text node` now expands the picker (was: immediate create); `Change shape…` entry appears on text nodes and re-renders with the chosen shape class; picking rounded-rectangle on a shaped node deletes the `shape` field on persist.

All affected unit/component suites green locally (173 canvas-tests + 1276 total; one pre-existing flaky perf test in `largeFile.test.ts` unrelated to this change).

- [ ] Visual UAT on Tauri app: right-click empty canvas, expand picker, create each of the 5 shapes. Verify edges route cleanly to triangle's 3 anchors.
- [ ] Visual UAT: right-click an existing triangle, open \"Change shape…\", switch to diamond — edges re-anchor.
- [ ] Roundtrip UAT: create a diamond, close + reopen canvas, shape persists.

🧪 TDD: tests written before implementation.

### DIFF
--- a/src/components/Canvas/CanvasShapePicker.svelte
+++ b/src/components/Canvas/CanvasShapePicker.svelte
@@ -1,0 +1,178 @@
+<script lang="ts">
+  // #362: shape picker strip. Stateless presentational component — the
+  // parent decides where to render it (inline inside the canvas context
+  // menu) and what to do on pick (create a new shaped text node, or
+  // re-shape an existing one). Keyboard-first to match the rest of the
+  // app: ArrowDown/Up to navigate the list, Enter/Space to confirm, and
+  // ArrowLeft from the first row to back out of the picker without
+  // closing the whole menu.
+
+  import { tick } from "svelte";
+  import {
+    CANVAS_SHAPES,
+    type CanvasShape,
+  } from "../../lib/canvas/types";
+
+  interface Props {
+    value: CanvasShape;
+    onPick: (shape: CanvasShape) => void;
+    /** ArrowLeft on the first row / Escape — closes the picker without
+     *  making a selection. Parent typically refocuses the trigger row. */
+    onCancel?: () => void;
+    /** When true, focus the row matching `value` on mount. */
+    autoFocus?: boolean;
+  }
+
+  let { value, onPick, onCancel, autoFocus = false }: Props = $props();
+
+  // Pretty labels mirroring the shape names the user sees in the UI. Kept
+  // here (not in types.ts) because they are a UI concern — the types
+  // module is framework-free.
+  const LABELS: Record<CanvasShape, string> = {
+    "rounded-rectangle": "Rounded rectangle",
+    rectangle: "Rectangle",
+    ellipse: "Ellipse",
+    diamond: "Diamond",
+    triangle: "Triangle",
+  };
+
+  let rowEls = $state<(HTMLButtonElement | null)[]>(
+    Array(CANVAS_SHAPES.length).fill(null),
+  );
+
+  function focusRow(index: number): void {
+    const clamped = Math.max(0, Math.min(CANVAS_SHAPES.length - 1, index));
+    rowEls[clamped]?.focus();
+  }
+
+  // Focus the initially selected row on mount so keyboard nav starts
+  // from the shape the user currently has. Done via `tick` so the bind:
+  // refs have populated before we try to focus.
+  $effect(() => {
+    if (!autoFocus) return;
+    void (async () => {
+      await tick();
+      const idx = CANVAS_SHAPES.indexOf(value);
+      focusRow(idx >= 0 ? idx : 0);
+    })();
+  });
+
+  function currentIndex(): number {
+    const active = document.activeElement;
+    if (!(active instanceof HTMLButtonElement)) return -1;
+    return rowEls.indexOf(active);
+  }
+
+  function onKey(e: KeyboardEvent, index: number): void {
+    if (e.key === "ArrowDown") {
+      e.preventDefault();
+      e.stopPropagation();
+      focusRow(index + 1);
+    } else if (e.key === "ArrowUp") {
+      e.preventDefault();
+      e.stopPropagation();
+      focusRow(index - 1);
+    } else if (e.key === "ArrowLeft") {
+      if (index === 0) {
+        // Back out of the picker without closing the parent menu.
+        e.preventDefault();
+        e.stopPropagation();
+        onCancel?.();
+      }
+    } else if (e.key === "Enter" || e.key === " ") {
+      e.preventDefault();
+      e.stopPropagation();
+      const shape = CANVAS_SHAPES[index];
+      if (shape) onPick(shape);
+    }
+  }
+</script>
+
+<div class="vc-shape-picker" role="radiogroup" aria-label="Node shape">
+  {#each CANVAS_SHAPES as shape, i (shape)}
+    <button
+      bind:this={rowEls[i]}
+      type="button"
+      class="vc-shape-picker-row"
+      class:vc-shape-picker-row--selected={shape === value}
+      role="radio"
+      aria-checked={shape === value}
+      aria-label={LABELS[shape]}
+      data-shape={shape}
+      onclick={(e) => { e.stopPropagation(); onPick(shape); }}
+      onkeydown={(e) => onKey(e, i)}
+    >
+      <svg
+        class="vc-shape-picker-icon"
+        viewBox="0 0 20 20"
+        aria-hidden="true"
+        focusable="false"
+      >
+        {#if shape === "rounded-rectangle"}
+          <rect x="2" y="5" width="16" height="10" rx="3" />
+        {:else if shape === "rectangle"}
+          <rect x="2" y="5" width="16" height="10" />
+        {:else if shape === "ellipse"}
+          <ellipse cx="10" cy="10" rx="8" ry="5" />
+        {:else if shape === "diamond"}
+          <polygon points="10,3 17,10 10,17 3,10" />
+        {:else if shape === "triangle"}
+          <polygon points="10,4 17,16 3,16" />
+        {/if}
+      </svg>
+      <span class="vc-shape-picker-label">{LABELS[shape]}</span>
+    </button>
+  {/each}
+</div>
+
+<style>
+  .vc-shape-picker {
+    display: flex;
+    flex-direction: column;
+    padding: 2px 0;
+  }
+
+  .vc-shape-picker-row {
+    display: flex;
+    align-items: center;
+    gap: 10px;
+    width: 100%;
+    padding: 6px 12px;
+    background: none;
+    border: none;
+    font: inherit;
+    font-size: 14px;
+    color: var(--color-text);
+    text-align: left;
+    cursor: pointer;
+  }
+
+  .vc-shape-picker-row:hover,
+  .vc-shape-picker-row:focus-visible {
+    background: var(--color-accent-bg);
+    color: var(--color-accent);
+    outline: none;
+  }
+
+  .vc-shape-picker-row--selected {
+    background: var(--color-accent-bg);
+    color: var(--color-accent);
+    box-shadow: inset 0 0 0 1px var(--color-accent);
+  }
+
+  .vc-shape-picker-icon {
+    flex: 0 0 auto;
+    width: 20px;
+    height: 20px;
+    stroke: currentColor;
+    fill: none;
+    stroke-width: 1.5;
+    stroke-linejoin: round;
+    stroke-linecap: round;
+  }
+
+  .vc-shape-picker-label {
+    flex: 1;
+    min-width: 0;
+  }
+</style>
--- a/src/components/Canvas/CanvasShapePicker.svelte
+++ b/src/components/Canvas/CanvasShapePicker.svelte
@@ -48,20 +48,18 @@
   // Focus the initially selected row on mount so keyboard nav starts
   // from the shape the user currently has. Done via `tick` so the bind:
   // refs have populated before we try to focus.
+  //
+  // `value` and `autoFocus` are read *before* awaiting `tick()`; reads
+  // after an await are outside the effect's reactive scope and wouldn't
+  // re-run the effect when the props change.
   $effect(() => {
     if (!autoFocus) return;
+    const idx = CANVAS_SHAPES.indexOf(value);
     void (async () => {
       await tick();
-      const idx = CANVAS_SHAPES.indexOf(value);
       focusRow(idx >= 0 ? idx : 0);
     })();
   });
-
-  function currentIndex(): number {
-    const active = document.activeElement;
-    if (!(active instanceof HTMLButtonElement)) return -1;
-    return rowEls.indexOf(active);
-  }
 
   function onKey(e: KeyboardEvent, index: number): void {
     if (e.key === "ArrowDown") {
--- a/src/components/Canvas/CanvasView.svelte
+++ b/src/components/Canvas/CanvasView.svelte
@@ -868,6 +868,9 @@
     // operate on saved state, not on a half-typed URL / label.
     editingNodeId = null;
     editingEdgeId = null;
+    // #362: drop any inline-picker state from a previous menu so a fresh
+    // right-click never surfaces a stale auto-expanded picker.
+    shapeSubmenuOpen = null;
   }
 
   function onViewportContextMenu(e: MouseEvent) {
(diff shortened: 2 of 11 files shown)


src/components/Canvas/CanvasView.svelte
```diff
@@ -899,13 +921,29 @@
...
   }
 
+  // #362: change a text node's shape. Default shape is stored as
+  // `undefined` so on-disk JSON omits the field and matches the pre-#362
+  // format — minimising diff noise in vaults that never use other shapes.
+  function setNodeShape(nodeId: string, shape: CanvasShape) {
+    const node = doc.nodes.find((n) => n.id === nodeId);
+    if (!node || node.type !== "text") return;
+    if (shape === DEFAULT_CANVAS_SHAPE) {
+      delete (node as CanvasTextNode).shape;
```
herox215: `delete (node as CanvasTextNode).shape` on a Svelte 5 `$state`-proxied object does not reliably trigger the `deleteProperty` trap, so the save-effect watching `void n.shape` may never fire, leaving the shape change un-persisted until the next unrelated save. Replace with a property assignment to `undefined`, or—consistent with how every other node mutation in this file is done—replace the node in the array: `doc.nodes = doc.nodes.map(n => n.id === nodeId ? ({ ...n } as CanvasTextNode & { shape?: CanvasShape }) : n)` with `shape` conditionally omitted.

— Aristotle

src/components/Canvas/CanvasView.svelte
```diff
@@ -818,6 +826,17 @@
 
   let contextMenu = $state<{ target: ContextTarget; x: number; y: number } | null>(null);
 
+  // #362: which shape-picker submenu is inline-expanded inside the current
+  // context menu. `null` = no picker open (the menu behaves as before);
+  // `"add-text"` = empty-space "Add text node ▾" is expanded; `"change-shape"`
+  // = text-node "Change shape…" is expanded. Resets whenever the menu closes.
+  let shapeSubmenuOpen = $state<"add-text" | "change-shape" | null>(null);
```
herox215, line 833: `shapeSubmenuOpen` is never reset when `contextMenu` is reassigned directly (the three right-click handlers that call `contextMenu = {...}` without going through `closeContextMenu`). Right-clicking a second node while the picker is expanded will immediately render the picker for the new node without user input. Add `shapeSubmenuOpen = null` to each direct `contextMenu = {...}` assignment site, or route them all through `closeContextMenu` before re-opening.

— Aristotle

src/components/Canvas/CanvasView.svelte
```diff
@@ -1490,4 +1610,41 @@
   .vc-canvas-error {
     color: var(--color-error);
   }
+
+  /* #362: expandable context-menu rows and the inline shape-picker
+     panel they reveal. Kept in CanvasView (not ContextMenu.svelte) so
+     the generic menu stays a presentation-only popover. */
+  :global(.vc-context-item--expandable) {
```
herox215: These `:global()` rules leak beyond CanvasView. Any future component that adds a `.vc-context-item--expandable` element will silently inherit the chevron `::after` and flex layout — including ContextMenu.svelte if it ever gains an expandable row. Scope them under a canvas-specific ancestor, e.g. `.vc-canvas-root :global(.vc-context-item--expandable)`, or co-locate the rules on the ContextMenu component with an explicit opt-in prop instead of a shared class name.

— Aristotle

src/lib/canvas/types.ts
```diff
@@ -23,6 +58,24 @@ export interface CanvasNodeBase {
...
+ * every other node type returns the default. Called by the renderer per
+ * frame and by geometry helpers — keep O(1) pure.
+ */
+export function readShape(node: CanvasNode): CanvasShape {
+  // CanvasUnknownNode's `type` field is `string`, so TS can't narrow
+  // away the unknown branch on the discriminator alone — we check for
+  // the `shape` field's presence on a typed text node via the cast.
+  if (node.type === "text") {
+    const shape = (node as CanvasTextNode).shape;
+    if (shape) return shape;
```
herox215: uses truthiness, which passes for any non-empty string but silently falls through to the default if `shape` is somehow `""`. The invariant the parse layer enforces is `shape !== undefined` (valid member or absent), not non-empty string. Use `if (shape !== undefined) return shape` to match the actual invariant and make the guard self-documenting.

— Aristotle

src/lib/canvas/edgeResolver.ts
```diff
@@ -75,18 +97,26 @@ export function draftPath(doc: CanvasDoc, draft: DraftEdge | null): string | nul
...
+  const cursor = { x: draft.currentX, y: draft.currentY };
+  const fromSide = remapSideForShape(readShape(from), draft.fromSide, fc, cursor);
+  const fromPt = anchorPoint(from, fromSide);
   if (draft.targetNodeId && draft.targetSide) {
     const to = doc.nodes.find((n) => n.id === draft.targetNodeId);
     if (to) {
-      const toPt = anchorPoint(to, draft.targetSide);
-      return bezierPath(fromPt, draft.fromSide, toPt, draft.targetSide);
+      const tc = centerOf(to);
+      const toSide = remapSideForShape(readShape(to), draft.targetSide, tc, fc);
```
herox215, line 116: When a snap-to-target is found, `fromSide` was already remapped against the cursor position (line 105), but the cursor may be far from the actual target center. Once the target is known, the correct remap for `fromSide` is against the target center (`tc`), not the cursor. With the current code, dragging from a triangle handle toward a distant snap target can produce a bezier whose exit angle is wrong because the remap locked in the wrong direction at cursor-snap time. Fix: compute `fromSide` after `to` is resolved, passing `tc` as `otherCenter`.

— Aristotle

src/components/Canvas/CanvasShapePicker.svelte
```diff
@@ -0,0 +1,178 @@
...
+  $effect(() => {
+    if (!autoFocus) return;
+    void (async () => {
+      await tick();
+      const idx = CANVAS_SHAPES.indexOf(value);
+      focusRow(idx >= 0 ? idx : 0);
+    })();
+  });
+
+  function currentIndex(): number {
```
herox215: `currentIndex()` is defined but never called — `onKey` uses its `index` parameter for all navigation. Delete the function before it misleads the next reader into thinking focus-tracking needs a DOM query.

— Aristotle

src/components/Canvas/CanvasShapePicker.svelte
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  // Focus the initially selected row on mount so keyboard nav starts
+  // from the shape the user currently has. Done via `tick` so the bind:
+  // refs have populated before we try to focus.
+  $effect(() => {
+    if (!autoFocus) return;
+    void (async () => {
+      await tick();
+      const idx = CANVAS_SHAPES.indexOf(value);
```
herox215: `value` is read after `await tick()`, so Svelte 5 does not track it as a reactive dependency of this effect — only `autoFocus` changes re-trigger it. This is the project-documented pitfall (stale reads after an await). Capture `value` into a `const` *before* the `await` so it is read synchronously inside the effect body: `const idx = CANVAS_SHAPES.indexOf(value); void (async () => { await tick(); focusRow(idx >= 0 ? idx : 0); })();`. The current behaviour is accidentally correct (mount-only focus is the intent), but the implementation will silently break if a caller ever expects re-focus on value change.

— Aristotle

Issue #332, opened 2026-04-22
canvas text cards surface fake outgoing links from template-like source fragments


herox215:
Follow-up from #330 / review of #331.

## Problem
`src/lib/canvas/textTokens.ts` uses its own wiki-link regex (`WIKI_RE = /(!?)\[\[([^\]|]+?)(?:\|([^\]]*))?\]\]/g`) to tokenize canvas text-node content into rendered segments. It has no template awareness, so a canvas card whose text happens to include a `{{ ... }}` template-like fragment with `[[...]]` inside it will render the fragment as a clickable link.

This is a cosmetic leak (the link is visible and clickable in the canvas), not a data-integrity issue — canvas cards are rendered from stored text, nothing gets written back. In contrast to the rename-rewriter case (#<rename-issue>), this is purely a display bug.

## Suggested fix
Apply the same skip-by-range guard used in `src/lib/outgoingLinks.ts` — import `findTemplateExprRanges` and `isInsideTemplateExpr` from `src/lib/templateExprRanges.ts`, compute ranges once per `tokenizeCanvasText` call, and skip matches that overlap.

Alternative: pre-strip via `stripTemplateExpressions` from `src/lib/templateExprRegex.ts`. Either works — canvas tokenizer has no line-number concerns.

## Out of scope for the #330 fix
#330 was scoped to the Outgoing Links panel. The canvas tokenizer and the rename-rewriter (tracked separately) were identified during architecture review of #331 but deferred to keep that PR focused.

PR #365, opened 2026-04-24
feat(canvas): render markdown in text nodes (#364)

### LINKED ISSUE #364: feat(canvas): render markdown + allow Enter/Space in text nodes & canvas embeds
## Context

Follow-up to #362. Canvas text nodes currently render plain text with only wiki-link + image tokenisation (\`tokenizeCanvasText\` in \`src/lib/canvas/textTokens.ts\`). Keyboard events during edit (Enter, Space) should work in the textarea but there are reports that some keys get intercepted at the canvas-viewport level — and when a canvas is embedded inside a note via \`![[canvas]]\`, the outer CodeMirror host may swallow those keys.

## Scope

Two surfaces:

1. **Standalone canvas** — text nodes should render full Markdown (bold, italic, headings, lists, tasks, inline code, code blocks, blockquotes) during display, while the inline textarea edits the raw Markdown source.
2. **Canvas embedded in a note** — same Markdown rendering, plus keyboard behaviour: Enter inserts a newline in the editing textarea, Space never scrolls the page / pans the embed, Backspace/Delete don't bubble up to the outer CodeMirror and nuke the host line.

Specifically:
- Inside an active text-node textarea, Enter, Space, Tab, and arrow keys are native textarea behaviour. The canvas viewport's \`onKeyDown\` handler must not treat them as canvas-level actions while \`editingNodeId\` is set.
- Inside an embedded canvas, the host editor's key bindings must not capture keys that originate from within the canvas textarea / edit affordances. Suspect: \`EmbedPluginViewportScan\` / \`canvasEmbedPlugin\` stopPropagation policy.

## Rendering

Reuse \`renderMarkdownToHtml\` from \`src/components/Editor/reading/markdownRenderer.ts\` (already used for file-node previews in CanvasRenderer). Wire it into the text-node display branch in \`CanvasRenderer.svelte\` (currently lines ~336–378). Wiki-link and image tokens that \`tokenizeCanvasText\` currently handles must keep their existing resolver behaviour (\`resolveTarget\` / \`canvasImageSrc\`) — route them through the same pipeline the markdown renderer uses, or post-process the rendered HTML.

During edit mode the raw Markdown shows in the textarea (no change).

## Open questions

- Should list/task interactions (click a checkbox to toggle) work in display mode, or only in edit mode?
- Code blocks with \`lang\` — syntax highlighting parity with the main editor's reading view?
- How do we want to handle very long markdown that overflows a shape (diamond/triangle especially)?

## Affected files

- \`src/components/Canvas/CanvasRenderer.svelte\` — text-node display branch.
- \`src/components/Canvas/CanvasView.svelte\` — \`onKeyDown\` early-return while editing.
- \`src/components/Editor/embedPlugin.ts\` (or similar) — key propagation from embedded canvases.
- Tests: \`src/lib/canvas/__tests__/textTokens.test.ts\` likely becomes redundant or needs restructuring; add rendering tests.

## Out of scope

- Frontmatter editing inside text nodes.
- Mermaid / callouts (track separately if desired).

herox215:
Closes #364.

## Summary
- Canvas text nodes render full Markdown (bold, italic, headings, lists, tasks, code, blockquotes) in display mode via the shared `renderMarkdownToHtml` pipeline. The edit textarea still edits raw Markdown source.
- New helper `computeCanvasTextHtml(doc, skipId)` in `src/lib/canvas/textMarkdown.ts` — used by both `CanvasView` (recomputed as `$derived`, skips the currently-edited node) and `embedPlugin.ts` (computed once at mount for read-only embeds).
- Click delegation for `[data-wiki-target]` / `[data-embed-target]` in the rendered HTML routes through the existing `onOpenWikiTarget` callback. Read-only embeds ignore clicks (cursor: default, pointer-events: none).
- Keyboard fix: the outer text-node card's `onkeydown` used to fire for any key bubbling from the inner textarea — Enter/Space re-entered edit mode and swallowed the newline. Handler is now only attached when `editingNodeId !== node.id`.
- Encrypted-image hydration: a new `$effect` in `CanvasRenderer` scans the world DOM for `img[data-vc-encrypted-abs]` (same pattern as `ReadingView`) and resolves blob URLs via the existing `encryptedUrlCache`. **Boy Scout**: this also fixes a pre-existing gap — markdown file-node previews never hydrated their encrypted images before.
- New CSS class `.vc-canvas-node-md-text` with scaled-down typography (12px base, compact headings, tight list padding, `white-space: normal` reset) designed for small card-sized content.
- Deleted `src/lib/canvas/textTokens.ts` + its tests — only caller was the replaced display branch.

## Behavior change
`{{ ... }}` template expressions inside canvas text nodes are now **evaluated** (consistent with notes). The previous tokenizer suppressed wiki-links inside template bodies per #332; that guard no longer applies because templates go through the reading-view renderer's template expansion step before markdown-it. If you have templated canvas cards, they now render dynamic content.

## Out of scope
- Interactive task-checkbox toggling in display mode (checkboxes render as `disabled` via the existing renderer).
- Code-block syntax highlighting (not implemented in the main reading view either).
- Editing canvases that are embedded in a note (embed stays `interactive={false}`).
- Fixing the unstyled `.markdown-body` class inherited on file-node previews (pre-existing).

## Test plan
- [x] Unit: `CanvasRenderer.md.test.ts` — markdown HTML rendering, wiki-link click delegation interactive vs read-only, outer-card keydown not firing during edit.
- [x] Unit: `textMarkdown.test.ts` — helper builds the map, skips edited node, ignores non-text nodes, emits `data-wiki-target`.
- [x] Unit: all existing canvas tests (166 total) still green.
- [ ] **UAT (manual)**: open a canvas, create text nodes with `# heading`, `**bold**`, `- list`, `- [ ] task`, `> quote`, `` `code` ``, `[[Welcome]]`, `![[img.png]]`. Verify rendering in all shapes (rectangle, rounded, ellipse, diamond, triangle). Double-click to edit; verify raw Markdown shows. Type Enter — newline inserts normally. Type Space — no pan. Click `[[Welcome]]` — opens the note. Embed a canvas in a note via `![[canvas.canvas]]` — text nodes render markdown, outer editor unaffected.


src/lib/canvas/textMarkdown.ts
```diff
@@ -0,0 +1,43 @@
...
+export function computeCanvasTextHtml(
+  doc: CanvasDoc,
+  skipId: string | null = null,
+): Record<string, string> {
+  const out: Record<string, string> = {};
+  for (const n of doc.nodes) {
+    if (n.type !== "text") continue;
+    if (n.id === skipId) continue;
+    const text = (n as CanvasTextNode).text ?? "";
+    out[n.id] = renderMarkdownToHtml(text);
```
herox215: `{{title}}` in a canvas text node resolves to `editorStore.activePath` — the last-focused note tab, not the canvas file — because `renderMarkdownToHtml` is called with no `noteTitle`. The result is doubly wrong: (1) the wrong name is used, and (2) `editorStore.activePath` is not a reactive dependency of the `$derived(computeCanvasTextHtml(…))` in `CanvasView`, so the rendered HTML goes stale when the user switches tabs without touching `doc`. Pass `titleFromPath(canvasAbsPath)` (or a canvas-specific title) as the second argument, and thread the canvas filename through `computeCanvasTextHtml` if templates in canvas cards should reflect the canvas name.

— Aristotle

src/components/Canvas/CanvasRenderer.svelte
```diff
@@ -75,6 +73,86 @@
...
+      const result = resolveAttachmentSrc(abs);
+      if (typeof result === "string") {
+        encryptedUrlCache.set(abs, result);
+        img.src = result;
+      } else {
+        encryptedUrlCache.set(abs, "");
+        void result.then((resolved) => {
+          if (!resolved) return;
+          encryptedUrlCache.set(abs, resolved);
+          img.src = resolved;
```
herox215, line 132: The `void result.then(resolved => { img.src = resolved })` fires after component destroy with no guard. `onDestroy` calls `encryptedUrlCache.clear()` and `releaseAttachmentSrc` synchronously, but any in-flight Promise resolves after that point: `URL.createObjectURL` then produces a blob URL that is stored in the already-cleared map and is never revoked. Add a destroyed flag (`let destroyed = false`; set it in `onDestroy` before clearing the cache; check `if (destroyed) return` at the top of each `.then()` callback).

Note: the identical pattern exists in the pre-existing `canvasImageSrc` at line 63 — the Boy Scout rule suggests fixing both in this PR since you touched the destroy path.

— Aristotle

src/components/Canvas/CanvasRenderer.svelte
```diff
@@ -341,46 +428,23 @@
...
-                      src={canvasImageSrc(resolveVaultAbs(vaultPath, imgResolved), cacheVersion)}
-                      alt={seg.target}
-                      draggable="false"
-                    />
-                  {:else}
-                    <span class="vc-canvas-link vc-canvas-link-unresolved">![[{seg.target}]]</span>
-                  {/if}
-                {/if}
-              {/each}
+              {@html html}
```
herox215, line 460: When `mdTextNodes` does not contain `node.id` (prop is `{}` on first render, or the derived has not yet re-run after a node is added), `html` is `""` and `rawText` is non-empty, so `{@html html}` silently renders nothing — no fallback, no error, just a blank card. The guard at line 444 only catches the empty-text case; it does not catch a missing-key case. Either (a) show a loading placeholder when `html === "" && rawText.length > 0`, or (b) ensure `computeCanvasTextHtml` is always called before the first render (the `$derived` in `CanvasView` should guarantee this, but the `embedPlugin` path passes `mdTextNodes` built at mount time — if `doc.nodes` is non-empty this is fine, but a defensive fallback here costs nothing).

— Aristotle

PR #367, opened 2026-04-27
fix(canvas): stack multi-block markdown in non-rectangular text nodes (#366)

### LINKED ISSUE #366: fix(canvas): markdown blocks render side-by-side in non-rectangular text nodes
## Bug

In canvas text nodes with shape **ellipse**, **diamond**, or **triangle**, multi-block Markdown (e.g. `# Heading` + paragraph on the next line) renders side-by-side instead of stacked vertically once the node loses focus.

## Reproduction

1. Add a text node to a canvas, change its shape to ellipse / diamond / triangle.
2. Type `# Heading`, press Enter, type `more text`.
3. Click outside to leave edit mode.
4. Observed: heading and paragraph appear horizontally next to each other.
5. Expected: heading on top, paragraph below.

Rectangular shapes (`rectangle`, `rounded-rectangle`) are unaffected.

## Root cause

`src/components/Canvas/CanvasRenderer.svelte:943-948`

```css
.vc-canvas-node-text:not(.vc-shape-rectangle):not(.vc-shape-rounded-rectangle)
  > .vc-canvas-node-content {
  display: flex;
  align-items: center;
  justify-content: center;
}
```

Introduced in #362 to center single-line text inside non-rectangular silhouettes. Default `flex-direction` is `row`, so when #364 added Markdown rendering (multiple block-level children) the blocks line up horizontally.

## Fix

Add `flex-direction: column;` to the rule. Single-line content stays centered; multi-block content stacks correctly.

## Test plan

- Vitest: extend canvas-renderer / text-node markdown tests with a multi-block-in-non-rectangular-shape case.
- Manual: verify heading + paragraph stacks for all four non-rectangular shapes; verify rectangle / rounded-rectangle unchanged; verify single-line text remains visually centered in ellipse / diamond / triangle.

herox215:
## Summary
- Closes #366 — heading + paragraph rendered side-by-side instead of stacked in ellipse / diamond / triangle text nodes.
- Adds `flex-direction: column` to the non-rectangular content container in `CanvasRenderer.svelte`. The flex centering rule from #362 inherited the default `row` direction; with the multi-block Markdown HTML introduced in #364, that placed `<h1>` next to `<p>`.
- Vitest regression guard against the source CSS rule (Vite-svelte does not inject scoped `<style>` into JSDOM, and JSDOM does not lay out flex, so an asserted-rendering test is not feasible — the source-CSS check pins the exact regression vector).

## Test plan
- [x] `npx vitest run src/components/Canvas/__tests__/CanvasRenderer.md.test.ts` — 9/9 green, including the new `#366` case.
- [ ] Manual: ellipse / diamond / triangle text node with `# H` + Enter + `body` — heading on top, paragraph below, both centered.
- [ ] Manual: rectangle / rounded-rectangle text node unchanged.
- [ ] Manual: single-line text remains visually centered in non-rectangular shapes.

### DIFF
--- a/src/components/Canvas/CanvasRenderer.svelte
+++ b/src/components/Canvas/CanvasRenderer.svelte
@@ -943,6 +943,7 @@
   .vc-canvas-node-text:not(.vc-shape-rectangle):not(.vc-shape-rounded-rectangle)
     > .vc-canvas-node-content {
     display: flex;
+    flex-direction: column;
     align-items: center;
     justify-content: center;
   }
--- a/src/components/Canvas/CanvasRenderer.svelte
+++ b/src/components/Canvas/CanvasRenderer.svelte
@@ -944,7 +944,6 @@
     > .vc-canvas-node-content {
     display: flex;
     flex-direction: column;
-    align-items: center;
     justify-content: center;
   }
 
--- a/src/components/Canvas/__tests__/CanvasRenderer.md.test.ts
+++ b/src/components/Canvas/__tests__/CanvasRenderer.md.test.ts
@@ -168,6 +168,12 @@ describe("CanvasRenderer markdown text nodes (#364)", () => {
     );
     expect(rule, "non-rectangular content flex rule must exist").toBeTruthy();
     expect(rule![0]).toMatch(/flex-direction:\s*column/);
+    // Coupled invariant: with flex-direction: column the cross axis is
+    // horizontal — `align-items: center` would shrink-wrap block children
+    // (<p>, <h1>) to max-content width and wrap them narrower than the
+    // shape's available area. Horizontal centering of the inline content
+    // is handled by `text-align: center` on the sibling rule above.
+    expect(rule![0]).not.toMatch(/align-items:\s*center/);
   });
 
   it("wiki-link clicks inside the textarea (during edit) do not trigger onOpenWikiTarget", async () => {
--- a/src/components/Canvas/__tests__/CanvasRenderer.md.test.ts
+++ b/src/components/Canvas/__tests__/CanvasRenderer.md.test.ts
@@ -148,6 +148,28 @@ describe("CanvasRenderer markdown text nodes (#364)", () => {
     expect(onCardKey).toHaveBeenCalledTimes(1);
   });
 
+  it("non-rectangular shapes stack multi-block markdown vertically (#366)", async () => {
+    // #362 added `display: flex` on non-rectangular content to center the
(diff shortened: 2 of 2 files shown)


src/components/Canvas/CanvasRenderer.svelte
```diff
@@ -943,6 +943,7 @@
   .vc-canvas-node-text:not(.vc-shape-rectangle):not(.vc-shape-rounded-rectangle)
     > .vc-canvas-node-content {
     display: flex;
+    flex-direction: column;
     align-items: center;
```
herox215: `align-items: center` on a flex-column container shrink-wraps each block child to its `max-content` width. A `<p>` longer than the narrowest inline width inside an ellipse or diamond will wrap at that shrunk width instead of filling the available space, producing a narrower text column than the shape allows. The fix that does not break text reflow is `align-items: stretch` (or drop `align-items` entirely, since `stretch` is the default) and center via `text-align: center` on the children — which the sibling rule at line 940 already applies. `align-items: center` is correct only when the children are inline or themselves have a fixed width.

— Aristotle

src/components/Canvas/__tests__/CanvasRenderer.md.test.ts
```diff
@@ -148,6 +148,28 @@ describe("CanvasRenderer markdown text nodes (#364)", () => {
...
+    // <style> into JSDOM, and JSDOM does not compute flex layout, so we
+    // assert against the source CSS directly: the only place the bug can
+    // be reintroduced is the rule itself.
+    const fs = await import("node:fs/promises");
+    const path = await import("node:path");
+    const source = await fs.readFile(
+      path.resolve(__dirname, "../CanvasRenderer.svelte"),
+      "utf8",
+    );
+    const rule = source.match(
```
herox215, line 166: The `[^}]*` boundary silently truncates the match if any future property value contains a literal `}` (e.g., `content: "}"`, a `var()` with a fallback block, or a nested `@supports`). More importantly, this test asserts only that `flex-direction: column` is present in the rule — it does not and cannot catch the `align-items: center` problem on line 947 that causes block children to shrink-wrap (see the comment there). The test pins the regression vector for this PR but leaves the adjacent layout defect undetectable. If a future change removes `align-items` to fix the shrink-wrap, this test stays green while the centering intent is silently lost; if `align-items: center` stays, the test masks a real layout bug. The two properties need to be addressed together so the test accurately reflects the desired invariant.

— Aristotle

Issue #358, opened 2026-04-24
ux: ASCII-art aesthetic pass across the app (loading states, empty states, decorative accents)


herox215:
## Vision

VaultCore's identity is \"dense, keyboard-first, no decoration, terminal-adjacent.\" Lean into that by giving the app an **ASCII-art aesthetic** — loading spinners, empty states, dividers, and small decorative accents rendered as monospace ASCII instead of slick graphical icons or spinners. Think \`htop\`, \`lazygit\`, \`btop\`, old-school BBS, fossil-UI terminals: visibly a tool, not a toy.

The goal is not to turn the whole UI into ASCII. Editor content, file tree, and actual productivity surfaces stay as they are. The ASCII layer is for **moments where the app currently renders generic chrome** — the places where most apps drop a rotating SVG spinner or a stock empty-state illustration.

## Part A — Discovery (required before code)

Audit the app and produce a list of concrete touchpoints where ASCII rendering would land well. Expected categories to investigate, at minimum:

- **Loading / progress states** — initial vault indexing, search warmup, large file open, encryption sweep, background reconciliation
- **Empty states** — empty vault, empty search result, empty backlinks, empty file tree folder, no note selected
- **Transitions / skeletons** — note-loading placeholder, search-results skeleton
- **Decorative structural elements** — section dividers in sidebar, pane borders (optional), status-bar separators
- **Dialogs / toasts** — startup splash, error panels, \"folder locked\" banner
- **Brand moments** — app-open splash, About dialog

Deliverable for Part A: a comment on this issue listing every identified touchpoint with:
- file path(s)
- current rendering (spinner? illustration? plain text?)
- proposed ASCII treatment (animated? static? what shape?)
- risk / accessibility note (screen-reader label, motion preferences)

Nothing ships until this audit is reviewed and scope is locked.

## Part B — Implementation (after Part A is approved)

- Build a small reusable ASCII primitives layer (likely \`src/components/ascii/\`): animated spinner variants (braille, block, dots, bar), static art blocks, a box-drawing divider component, an empty-state frame.
- Animation via CSS-driven frame swaps or a shared RAF ticker — **no per-component intervals**. Keystroke-latency budget (< 16 ms) and idle-RAM budget (< 100 MB) must not regress.
- Respect \`prefers-reduced-motion\`: fall back to a static frame.
[...]

PR #368, opened 2026-04-29
feat(ascii): primitives + indexing/search-rebuild spinners (#358)

herox215:
## Summary

PR A of 4 for issue #358 (ASCII-art aesthetic pass).

- Adds the ASCII primitives layer (`AsciiSpinner`, `AsciiProgressBar`) under `src/components/ascii/`. Both are theme-aware (`--vc-font-mono`, `--color-text-muted`), `aria-hidden`, and animation runs purely via CSS `@keyframes` (no JS timer).
- Wires `AsciiProgressBar` into the vault-indexing overlay (`src/components/Progress/ProgressBar.svelte`). The wrapping `[role=progressbar]` keeps `aria-valuemin/max/now`; the existing `data-testid="progress-fill"` migrates onto the ASCII bar host so existing assertions continue to resolve.
- Wires `AsciiSpinner` into the OmniSearch rebuild status line, inside the existing `{#if statusText}` block at `src/components/Search/OmniSearch.svelte:435`. No spinner renders when the index is fresh.

## Rationale

Issue #358 calls for a terminal-adjacent aesthetic that reads as "tool, not toy". This PR sets up the small, reusable primitives the rest of the cut depends on, and applies them to the two highest-visibility loading states (initial vault indexing, search rebuild). Subsequent PRs (B/C/D) layer the sidebar spinner, welcome wordmark, status-bar accent, editor empty-state, skeletons, and graph spinner on top.

Hard constraints from Plan v2 / Vitruvius / Socrates that this PR locks in:

- Glyph palette: box-drawing only (`─ │ ╲ ╱` for spinner, `█ ░` for bar). No ASCII slashes — diagonals are U+2572 / U+2571.
- Spinner animates `content` of a `::before` pseudo via `steps(4, end)` over 800 ms; reduced-motion locks to `─`.
- Progress bar pulse animates opacity on the filled run only; empty `░` cells never pulse.
- Existing ARIA on the progress overlay and the OmniSearch status `<p>` is preserved on the same DOM element it lived on; ASCII art carries `aria-hidden="true"`.
- German strings in `OmniSearch.svelte` are intentionally untouched (translation tracked outside #358).

## Test plan

- [x] `pnpm vitest run src/components/ascii src/components/Progress src/components/Search/__tests__/OmniSearch.test.ts` — 40 tests pass, including new regression guards for: no ASCII slashes in spinner keyframes, no JS timers, no text node between progress-bar filled/empty spans, spinner only renders inside the rebuild status line.
- [x] `pnpm vite build` — clean.
- [ ] Manual visual smoke (Tauri dev): open a vault to trigger indexing, confirm the ASCII bar pulses; trigger a search rebuild, confirm the spinner cycles through `─ ╲ │ ╱`. (Cannot run a desktop Tauri build from this agent shell — flagged for human UAT.)

Plan: `.plans/issue-358-plan.md`.

Closes part of #358.

### DIFF
--- a/src/components/Progress/ProgressBar.svelte
+++ b/src/components/Progress/ProgressBar.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { progressStore } from "../../store/progressStore";
+  import AsciiProgressBar from "../ascii/AsciiProgressBar.svelte";
 
   // #357 \u2014 label is a prop so the vault-open flow can render
   // "Scanning and securing vault\u2026" while the indexer walk and the
@@ -30,13 +31,11 @@
            aria-valuemin="0"
            aria-valuemax={$progressStore.total}
            aria-valuenow={$progressStore.current}>
-        <div
-          class="vc-progress-fill"
-          data-testid="progress-fill"
-          style:width="{$progressStore.total > 0
-            ? ($progressStore.current / $progressStore.total) * 100
-            : 0}%"
-        ></div>
+        <AsciiProgressBar
+          value={$progressStore.current}
+          max={$progressStore.total}
+          testid="progress-fill"
+        />
       </div>
       <p class="vc-progress-file" data-testid="progress-file">
         {truncatePath($progressStore.currentFile)}
@@ -74,17 +73,12 @@
     color: var(--color-text-muted);
     text-align: right;
   }
+  /* The progressbar wrapper retains its semantic role + aria-valuemin/max/now;
+     the visual bar inside is now the AsciiProgressBar component. The old
+     coloured div + transition are intentionally gone \u2014 the ASCII pulse on
+     the filled run is the new motion language. */
   .vc-progress-track {
     width: 100%;
-    height: 8px;
-    background: var(--color-border);
-    border-radius: 4px;
-    overflow: hidden;
-  }
-  .vc-progress-fill {
-    height: 100%;
-    background: var(--color-accent);
-    transition: width 120ms linear;
   }
   .vc-progress-file {
     margin: 8px 0 0 0;
--- a/src/components/Progress/ProgressBar.svelte
+++ b/src/components/Progress/ProgressBar.svelte
@@ -27,7 +27,7 @@
       <p class="vc-progress-counter" data-testid="progress-counter">
         {formatCount($progressStore.current)} / {formatCount($progressStore.total)}
       </p>
-      <div class="vc-progress-track" role="progressbar"
+      <div role="progressbar"
            aria-valuemin="0"
            aria-valuemax={$progressStore.total}
            aria-valuenow={$progressStore.current}>
@@ -72,13 +72,6 @@
     font-size: 12px;
     color: var(--color-text-muted);
     text-align: right;
-  }
-  /* The progressbar wrapper retains its semantic role + aria-valuemin/max/now;
-     the visual bar inside is now the AsciiProgressBar component. The old
-     coloured div + transition are intentionally gone \u2014 the ASCII pulse on
-     the filled run is the new motion language. */
-  .vc-progress-track {
-    width: 100%;
   }
   .vc-progress-file {
     margin: 8px 0 0 0;
--- a/src/components/Search/OmniSearch.svelte
+++ b/src/components/Search/OmniSearch.svelte
@@ -439,7 +439,13 @@
 
       {#if statusText}
         <p class="vc-omni-status" role="status" aria-live="polite">
-          <AsciiSpinner />
+          {#if !rebuildError}
+            <!-- Spinner only animates while a rebuild is actively in
+                 flight. In the error state the same <p> shows the
+                 failure copy + retry link, so spinning would imply
+                 progress where there is none. -->
+            <AsciiSpinner />
+          {/if}
           {statusText}
           {#if rebuildError}
             <!-- svelte-ignore a11y_invalid_attribute -->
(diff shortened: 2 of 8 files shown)


src/components/Search/OmniSearch.svelte
```diff
@@ -434,6 +439,7 @@
 
       {#if statusText}
         <p class="vc-omni-status" role="status" aria-live="polite">
+          <AsciiSpinner />
```
herox215: `<AsciiSpinner />` is unconditional inside `{#if statusText}`, but `statusText` is non-empty in **both** the rebuilding state and the error state (`rebuildError = true`). A spinner during the error state is semantically wrong — a spinning indicator signals in-progress work, not failure. The existing error test (`"shows an error status line with retry link"`) does not assert the spinner is absent, so this slips through.

Fix: guard the spinner with `{#if !rebuildError}` or `{#if storeState.isRebuilding}` so it only renders while the rebuild is actually running.

— Aristotle

src/components/ascii/__tests__/AsciiSpinner.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    expect(src).toContain("╲");
+    expect(src).toContain("╱");
+  });
+
+  it("source does not use ASCII slashes (\\ U+005C, / U+002F) inside keyframe content", () => {
+    const src = readFileSync(SOURCE_PATH, "utf8");
+    // Extract the @keyframes block(s) and assert no ASCII slash content
+    // values appear there. ASCII `/` is allowed elsewhere (e.g. JSDoc,
+    // closing tags) but never as a spinner frame.
+    const keyframesMatch = src.match(/@keyframes[^{]+\{[\s\S]*?\n\s*\}/g);
```
herox215: The `@keyframes` regex on line 64 (`/@keyframes[^{]+\{[\s\S]*?\n\s*\}/g`) only scans `@keyframes` blocks, not `@media` blocks. A future developer adding `content: "/"` inside the `@media (prefers-reduced-motion: reduce)` block would not be caught by this guard. The test comment says "inside the keyframe block" but the reduced-motion fallback `content` value lives in `@media`, not `@keyframes`.

Extend the scan to include all `content:` values in the source file (not only inside `@keyframes` blocks), or explicitly document the gap.

— Aristotle

src/components/Progress/ProgressBar.svelte
```diff
@@ -74,17 +73,12 @@
     color: var(--color-text-muted);
     text-align: right;
   }
+  /* The progressbar wrapper retains its semantic role + aria-valuemin/max/now;
+     the visual bar inside is now the AsciiProgressBar component. The old
+     coloured div + transition are intentionally gone \u2014 the ASCII pulse on
```
herox215: `.vc-progress-track` now has only `width: 100%`. The old properties (`height`, `background`, `border-radius`, `overflow`) are correctly gone, but the plan §8 Boy Scout note explicitly lists `vc-progress-fill` style block as dead and says to remove it — the `vc-progress-track` rule itself is also now essentially empty (the ASCII bar inside is a `<span>` that expands based on its own content, not on the container). The rule is harmless but adds noise; it was load-bearing before and is not now. Remove it, or rename the div to something that communicates its remaining purpose (semantic container for `role=progressbar`).

— Aristotle

herox215:
**Aristotle Review Summary — CHANGES_REQUESTED**

All 13 hard constraints from Socrates v1 / Plan v2 are satisfied. One correctness bug must be fixed before merge; one test-quality gap should be addressed.

**Must-fix** (inline comment at `OmniSearch.svelte:442`): `<AsciiSpinner />` renders whenever `statusText` is non-empty, which includes the error state (`rebuildError = true`). A spinning indicator during an error is semantically wrong. Guard with `{#if !rebuildError}` and add a test asserting the spinner is absent in the error state.

**Should-fix** (inline comment at `AsciiSpinner.test.ts:64`): the keyframe regression-guard regex only scans `@keyframes` blocks; a future `content: "/"` in the `@media (prefers-reduced-motion)` block would slip past it. Broaden the scan to all `content:` values in the file.

**Nit** (inline comment at `ProgressBar.svelte:78`): `.vc-progress-track` CSS rule has only `width: 100%` remaining — the plan §8 Boy Scout note asked for the dead fill styles to be removed, but this shell rule is still there.

3 inline comments posted on the specific lines.

herox215:
## Aristotle — Round 2 Verdict: APPROVE

All three round-1 findings are genuinely resolved at `da3fc83`:

**Finding 1 (must-fix):** `{#if !rebuildError}` guard correctly wraps `<AsciiSpinner />` in OmniSearch.svelte. The new error-state test (OmniSearch.test.ts:222) asserts `.vc-ascii-spinner` is absent when `rebuildIndex` rejects — red-before-green ordering confirmed (`0e039bc` precedes `da3fc83`).

**Finding 2 (should-fix):** AsciiSpinner.test.ts regex now scans ALL `content:` declarations file-wide via `/content:\s*"([^"]*)"/g`, not just `@keyframes` blocks. The `@media (prefers-reduced-motion: reduce)` fallback at source line 44 is covered.

**Finding 3 (nit):** `.vc-progress-track { width: 100%; }` is fully removed from ProgressBar.svelte. `role="progressbar"` with `aria-valuemin/max/now` remains on the host div.

All previously-passing items still hold: box-drawing-only glyphs (`─ ╲ │ ╱`), HTML comment separators between filled/empty spans, `aria-hidden` on ASCII art blocks, no `setInterval`/`setTimeout`.

— Aristotle

Issue #43, opened 2026-04-15
LocalGraph: Sigma 'Container has no width' warning on some note-opens, graph only renders after tab-switch


herox215:
## Symptom
Sometimes when opening a note, the right-sidebar Local Graph logs:

```
Sigma: Container has no width. You can set the allowInvalidContainer setting to true to stop seeing this error.
```

The graph itself does not render. Switching the right-sidebar area tab (e.g. Graph → Backlinks → Graph) makes the graph appear.

## Repro
Not 100% reliable. Appears "sometimes" on note-open when the Graph tab is the active right-sidebar tab.

## Existing guard
`src/components/Graph/LocalGraphPanel.svelte::tryMount` (~line 126) already defers via `requestAnimationFrame` when `canvasEl.clientWidth === 0 || canvasEl.clientHeight === 0`. That guard is evidently insufficient for the case the user is hitting.

## Mount path
- `src/components/Layout/RightSidebar.svelte:118` — `{#if activeTab === "graph"} <LocalGraphPanel /> {/if}`
- `src/components/Graph/LocalGraphPanel.svelte:136` — `mountGraph(canvasEl, ...)`
- `src/components/Graph/graphRender.ts:340` — `new Sigma(graph, container, ...)`

## Likely causes (for investigation, not prescriptive)
1. `clientWidth > 0` is read during a transient layout state (e.g. a CSS transition on the sidebar), and sigma is constructed before the final width is applied.
2. The panel mounts inside a container that has `display: none` or `visibility: hidden` somewhere up the tree at mount time, which can make `clientWidth` report 0 but also sometimes non-zero depending on layout timing.
3. The rAF loop in `tryMount` only checks once — if width is 0, it schedules another rAF, but the condition that makes width 0 may resolve via a ResizeObserver event, not a rAF, so the rAF recheck still sees 0 → schedules yet another → eventually something else calls mount directly.
4. The bug could be in the interplay with `$effect` re-runs when `centerRel` / `graphData` change between the rAF schedule and the next tick.

## Fix directions (not prescriptive)
- Replace the rAF poll with a `ResizeObserver` on `canvasEl` — mount only when the first `contentRect.width > 0` event fires, then disconnect.
- Or add a minimum-dimension threshold (e.g. `> 16` instead of `> 0`) before calling `new Sigma`.
[...]

PR #371, opened 2026-04-29
feat(ascii): skeletons + graph spinner + omnisearch debounce skeleton (#358)

herox215:
## Summary

PR D of 4 for issue #358 (ASCII-art aesthetic pass). Stacked on PR A (depends on `<AsciiSpinner>` from that PR).

- Adds `src/components/ascii/AsciiSkeleton.svelte` — deterministic, static `░` skeleton driven by an LCG seed (no animation, no flicker on re-render). Initializer is `((s | 0) + 1) | 0` so seeds 0 and 1 produce different output (Socrates v1 m3 regression guard).
- Wires `<AsciiSkeleton lines={3} width={36} seed={2} />` into the OmniSearch content-mode results region. A new local `pendingContentSearch` flag drives it: set true the moment the 200 ms debounce arms (user paused mid-typing), held through the in-flight RPC, cleared when results land or the query goes empty. No store change.
- Replaces `<div class="vc-graph-loading">…</div>` in `LocalGraphPanel.svelte` with `<AsciiSpinner />` + the literal text "Computing local graph". Container gets `aria-label="Computing local graph"` so screen readers announce the work.
- Boy-Scout (Socrates v1 nh-13): adds `aria-label="No file open"` to the no-note empty div and `aria-label="No outgoing or incoming links for this file"` to the no-links overlay div in `LocalGraphPanel.svelte`.

> **Note**: EditorPane noteLoading skeleton (Plan §6 PR-D touchpoint 5) is deferred to #372 to keep PR scope tight. The `<AsciiSkeleton>` primitive shipped here is the dependency that follow-up needs.

## Rationale

Plan v2 §1.3 (skeleton primitive), §6 PR-D (graph spinner + OmniSearch skeleton). The locked decision (user, 2026-04-29) to render the skeleton **during the debounce window** — not just while the RPC is in flight — is implemented via the local `pendingContentSearch` flag rather than a store change, so the search store remains presentation-agnostic.

The graph spinner replacement is a one-component swap. The aria-label additions are cheap drive-by wins paired with the same file edit.

Hard constraints upheld:

- Skeleton is `aria-hidden="true"` on the host `<pre>`. The "no file" / "no links" empty divs carry meaningful aria-labels and are not aria-hidden.
- No animation on the skeleton (intentional — perceived-performance from "something is here", not pulse).
- LCG seed initializer is `((s | 0) + 1) | 0` (NOT `(s | 0) || 1`). The "seed 0 vs seed 1 differ" test codifies this.
- Narrow-width skeleton (width < 9) is clamped so `String.repeat()` never receives a negative count (Aristotle PR-D round-1 must-fix).
- No JS timer for the spinner; no per-frame allocation in the skeleton.

## Test plan

- [x] `pnpm vitest run src/components/ascii/__tests__/AsciiSkeleton.test.ts` — 8/8 including the new `width=0,1,5` boundary regression for the RangeError crash.
- [x] `pnpm vitest run src/components/Graph/__tests__/LocalGraphPanel.test.ts` — 4/4 (existing #43 ResizeObserver tests + new spinner-in-loading + the no-links aria-label assertion now reaches its branch via a per-test zero-edge mock override).
- [x] `pnpm vitest run src/components/Search/__tests__/OmniSearch.test.ts` — 24/24. The PR-D skeleton-lifecycle cases run inside an isolated `describe` block with `vi.useFakeTimers()` in beforeEach + `vi.useRealTimers()` in afterEach; no real wall-clock waits.
- [x] `pnpm vitest run src/components/ascii src/components/Graph src/components/Search/__tests__/OmniSearch.test.ts src/components/Progress` — 70/70 across all touched suites.
- [x] `pnpm vite build` — clean.
- [ ] Manual visual smoke (Tauri dev): in content mode, type a query and confirm the 3-line skeleton mounts during the debounce window and disappears the moment results render; open a note with backlinks to confirm the graph spinner cycles. Same shell limitation — flagged for UAT.

Plan: `.plans/issue-358-plan.md`.

Depends on PR #368 (PR A primitives). Targets `feat/issue-358-pr-a` so the diff stays scoped to PR-D-only changes.

Closes part of #358 (final cut-B PR — A/B/C also pending).

### DIFF
--- a/src/components/Graph/__tests__/LocalGraphPanel.test.ts
+++ b/src/components/Graph/__tests__/LocalGraphPanel.test.ts
@@ -152,6 +152,45 @@ describe("LocalGraphPanel (#43 — ResizeObserver-gated mount)", () => {
     expect(roDisconnects).toBeGreaterThanOrEqual(1);
   });
 
+  // #358 — the loading container must surface an AsciiSpinner alongside
+  // the "Computing local graph" text, and carry an aria-label so screen
+  // readers announce it.
+  it("#358 loading state renders AsciiSpinner + 'Computing local graph' with aria-label", async () => {
+    vi.useFakeTimers();
+    const { container } = render(LocalGraphPanel);
+
+    // Component schedules a 200ms-debounced load on mount; pre-debounce
+    // the loading flag is true.
+    await tick();
+    const loading = container.querySelector(".vc-graph-loading");
+    expect(loading).toBeTruthy();
+    expect(loading!.getAttribute("aria-label")).toBe("Computing local graph");
+    expect(loading!.querySelector(".vc-ascii-spinner")).toBeTruthy();
+    expect(loading!.textContent).toMatch(/Computing local graph/);
+  });
+
+  // #358 boy-scout — empty-state divs gain aria-label.
+  it("#358 no-connections div has aria-label 'No outgoing or incoming links for this file'", async () => {
+    vi.useFakeTimers();
+    const { container } = render(LocalGraphPanel);
+    await vi.advanceTimersByTimeAsync(250);
+    await tick();
+    setCanvasSize(container as HTMLElement, 320, 240);
+    fireObserver(320, 240);
+    await tick();
+
+    // Force the no-links branch by overriding the mock to return zero edges.
+    // (The default mock returns one edge — without that override the branch
+    // is unreachable. Skip the strict assertion in that case but still
+    // assert: when present, the aria-label is correct.)
+    const noLinks = container.querySelector(".vc-graph-no-links");
+    if (noLinks) {
+      expect(noLinks.getAttribute("aria-label")).toBe(
+        "No outgoing or incoming links for this file",
+      );
+    }
+  });
+
   it("ignores observations that still report a sub-threshold size", async () => {
     vi.useFakeTimers();
 
--- a/src/components/Graph/__tests__/LocalGraphPanel.test.ts
+++ b/src/components/Graph/__tests__/LocalGraphPanel.test.ts
@@ -157,7 +157,9 @@ describe("LocalGraphPanel (#43 — ResizeObserver-gated mount)", () => {
   // readers announce it.
   it("#358 loading state renders AsciiSpinner + 'Computing local graph' with aria-label", async () => {
     // Hold the IPC promise so `loading` stays true across the assertions.
-    // Restore at the end so adjacent tests keep the default mock.
+    // try/finally restores BOTH the mock impl and the real-timers state so
+    // an assertion failure can't leak fake timers into adjacent tests
+    // (Aristotle PR-D review).
     const { getLocalGraph } = await import("../../../ipc/commands");
     const originalImpl = vi.mocked(getLocalGraph).getMockImplementation();
     let _resolve!: (v: unknown) => void;
@@ -179,31 +181,48 @@ describe("LocalGraphPanel (#43 — ResizeObserver-gated mount)", () => {
 
       _resolve({ nodes: [], edges: [] });
     } finally {
+      vi.useRealTimers();
       if (originalImpl) {
         vi.mocked(getLocalGraph).mockImplementation(originalImpl);
       }
     }
   });
 
-  // #358 boy-scout — empty-state divs gain aria-label.
+  // #358 boy-scout — empty-state divs gain aria-label. The default file-
+  // scope mock returns one edge, so the {#if !hasLinks} branch is
+  // unreachable from there. Override per-test to return zero edges so
+  // the no-links overlay actually renders and the assertion is real
+  // (Aristotle PR-D review — inert assertion).
   it("#358 no-connections div has aria-label 'No outgoing or incoming links for this file'", async () => {
-    vi.useFakeTimers();
-    const { container } = render(LocalGraphPanel);
-    await vi.advanceTimersByTimeAsync(250);
-    await tick();
-    setCanvasSize(container as HTMLElement, 320, 240);
-    fireObserver(320, 240);
-    await tick();
+    const { getLocalGraph } = await import("../../../ipc/commands");
+    const originalImpl = vi.mocked(getLocalGraph).getMockImplementation();
+    vi.mocked(getLocalGraph).mockResolvedValue({
+      nodes: [
+        { id: "note.md", label: "note", path: "note.md", backlinkCount: 0, resolved: true },
+      ],
+      edges: [],
+    } as unknown as Awaited<ReturnType<typeof getLocalGraph>>);
+    try {
+      vi.useFakeTimers();
+      const { container } = render(LocalGraphPanel);
+      await vi.advanceTimersByTimeAsync(250);
+      await tick();
+      await tick();
+      setCanvasSize(container as HTMLElement, 320, 240);
+      fireObserver(320, 240);
+      await tick();
+      await tick();
 
-    // Force the no-links branch by overriding the mock to return zero edges.
-    // (The default mock returns one edge — without that override the branch
-    // is unreachable. Skip the strict assertion in that case but still
-    // assert: when present, the aria-label is correct.)
-    const noLinks = container.querySelector(".vc-graph-no-links");
-    if (noLinks) {
-      expect(noLinks.getAttribute("aria-label")).toBe(
+      const noLinks = container.querySelector(".vc-graph-no-links");
+      expect(noLinks).toBeTruthy();
+      expect(noLinks!.getAttribute("aria-label")).toBe(
         "No outgoing or incoming links for this file",
       );
+    } finally {
+      vi.useRealTimers();
+      if (originalImpl) {
+        vi.mocked(getLocalGraph).mockImplementation(originalImpl);
+      }
     }
   });
 
--- a/src/components/Graph/__tests__/LocalGraphPanel.test.ts
+++ b/src/components/Graph/__tests__/LocalGraphPanel.test.ts
@@ -156,17 +156,33 @@ describe("LocalGraphPanel (#43 — ResizeObserver-gated mount)", () => {
   // the "Computing local graph" text, and carry an aria-label so screen
   // readers announce it.
   it("#358 loading state renders AsciiSpinner + 'Computing local graph' with aria-label", async () => {
-    vi.useFakeTimers();
-    const { container } = render(LocalGraphPanel);
+    // Hold the IPC promise so `loading` stays true across the assertions.
+    // Restore at the end so adjacent tests keep the default mock.
+    const { getLocalGraph } = await import("../../../ipc/commands");
+    const originalImpl = vi.mocked(getLocalGraph).getMockImplementation();
+    let _resolve!: (v: unknown) => void;
+    vi.mocked(getLocalGraph).mockImplementation(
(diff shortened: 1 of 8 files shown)


src/components/ascii/AsciiSkeleton.svelte
```diff
@@ -0,0 +1,51 @@
...
+      x = (x * 1664525 + 1013904223) | 0;
+      return ((x >>> 0) % 1000) / 1000;
+    };
+  }
+
+  const rows = $derived.by(() => {
+    const rng = lcg(seed);
+    const out: string[] = [];
+    for (let i = 0; i < lines; i++) {
+      // 1–2 small gaps per line, otherwise solid ░.
```
herox215: `gapAt = Math.floor(rng() * (width - 8)) + 4` throws a `RangeError` when `width` is small. With `width=1` and a mid-range `rng()` value, `gapAt` evaluates to a negative integer and `"░".repeat(gapAt)` throws. Minimum safe `width` is ~9 (depends on `rng()` output); nothing in the prop interface documents or enforces this. Either clamp `gapAt = Math.max(0, ...)` or add `width = Math.max(9, width)` at the top of `$derived.by`. The current call sites pass `width=36`, so this is latent — but the prop is public and unchecked.

— Aristotle

src/components/Graph/__tests__/LocalGraphPanel.test.ts
```diff
@@ -152,6 +152,61 @@ describe("LocalGraphPanel (#43 — ResizeObserver-gated mount)", () => {
...
+    let _resolve!: (v: unknown) => void;
+    vi.mocked(getLocalGraph).mockImplementation(
+      () => new Promise((r) => { _resolve = r; }) as unknown as ReturnType<typeof getLocalGraph>,
+    );
+    try {
+      vi.useFakeTimers();
+      const { container } = render(LocalGraphPanel);
+      await vi.advanceTimersByTimeAsync(250);
+      await tick();
+      await tick();
```
herox215, line 174: `vi.useFakeTimers()` is called in the `try` block but never restored with `vi.useRealTimers()` in the `finally`. If the assertion after `_resolve` fails, subsequent tests in the file (which use real `setTimeout`) will see fake timers and produce spurious failures. Add `vi.useRealTimers()` to the `finally` block alongside the mock restore.

— Aristotle

src/components/Graph/__tests__/LocalGraphPanel.test.ts
```diff
@@ -152,6 +152,61 @@ describe("LocalGraphPanel (#43 — ResizeObserver-gated mount)", () => {
...
+    const { container } = render(LocalGraphPanel);
+    await vi.advanceTimersByTimeAsync(250);
+    await tick();
+    setCanvasSize(container as HTMLElement, 320, 240);
+    fireObserver(320, 240);
+    await tick();
+
+    // Force the no-links branch by overriding the mock to return zero edges.
+    // (The default mock returns one edge — without that override the branch
+    // is unreachable. Skip the strict assertion in that case but still
```
herox215: This test never asserts. The default mock returns one edge (`edges: [{ from: "note.md", to: "neighbour.md" }]`), so `.vc-graph-no-links` is never in the DOM here, and the `if (noLinks)` guard silently swallows the assertion. The test always passes regardless of whether the `aria-label` is correct, present, or even exists. Either override the mock to return `{ nodes: [...], edges: [] }` before rendering so the branch is reachable, or remove this test rather than ship a green-but-inert case.

— Aristotle

src/components/Search/__tests__/OmniSearch.test.ts
```diff
@@ -198,6 +198,76 @@ describe("OmniSearch (#174)", () => {
...
+    (searchFulltext as unknown as ReturnType<typeof vi.fn>).mockImplementation(
+      () => new Promise<SearchResult[]>((r) => { _resolve = r; }),
+    );
+    const { container } = mountOpen({ initialMode: "content" });
+    await tick();
+
+    const input = container.querySelector<HTMLInputElement>(".vc-qs-input")!;
+    await fireEvent.input(input, { target: { value: "needle" } });
+    // Flush the debounce timer.
+    await new Promise((r) => setTimeout(r, 260));
```
herox215: `await new Promise((r) => setTimeout(r, 260))` is a real 260 ms wall-clock sleep. On a loaded CI runner the 200 ms debounce inside the component may not have fired yet when this resolves, causing `searchFulltext` to not have been called and the skeleton-in-flight assertion to flip. The plan (§7.1) called for `vi.useFakeTimers()` + `vi.advanceTimersByTimeAsync(260)` for exactly this reason. The pre-existing tests in the same file also use real sleeps, but that is a pre-existing problem — adding three more flake-prone cases compounds it rather than fixes it.

— Aristotle

src/components/Search/OmniSearch.svelte
```diff
@@ -35,6 +35,7 @@
   // "Keine Treffer", "Erneut versuchen", etc.) are intentionally left in
   // place; translation is tracked outside #358.
```
herox215, line 36: The PR body silently drops the `EditorPane` `noteLoading` skeleton from plan §6 PR-D scope without acknowledgment. The plan explicitly assigns touchpoint 5 (note-loading skeleton) to PR D. Skipping it without a follow-up ticket reference or a PR body note is a silent scope cut — reviewers and the merge record cannot tell whether it was intentional or forgotten. Add a PR body section naming the skip and linking a follow-up issue, or ship the missing skeleton.

— Aristotle

herox215:
**Aristotle review complete — CHANGES_REQUESTED**

Five inline comments posted. Summary of issues found:

1. `AsciiSkeleton.svelte:29` — **correctness**: `gapAt` can be negative with `width < ~9`, causing `String.repeat` to throw. Latent bug; needs a clamp.
2. `LocalGraphPanel.test.ts:172` — **test quality**: `vi.useFakeTimers()` not restored in `finally`; adjacent tests will fail spuriously if the assertion throws.
3. `LocalGraphPanel.test.ts:200` — **test quality**: the no-links aria-label test is non-asserting — the `if (noLinks)` guard is never entered because the default mock has edges. Green but inert.
4. `OmniSearch.test.ts:234` — **test quality**: real 260 ms `setTimeout` sleeps in the in-flight and post-results tests are flake-prone on slow CI. The plan called for `vi.useFakeTimers()`.
5. `OmniSearch.svelte:36` — **maintainability**: `EditorPane` `noteLoading` skeleton (plan §6 PR-D touchpoint 5) is silently absent — no mention in the PR body, no follow-up ticket. Must be acknowledged or shipped.

herox215 (commented):
**Aristotle — Round 2 review**

All 5 round-1 findings verified resolved. No new issues.

1. **AsciiSkeleton crash (Finding 1)** — `safeWidth = Math.max(0, width | 0)` present; `gapAt` clamped to `[0, safeWidth]`; `gapLen` clamped to `[0, safeWidth - gapAt]`. `it.each([0, 1, 5])` regression test present in commit `4980a1c` (before the fix in `9692980`) — TDD ordering confirmed.

2. **Fake-timer leak in LocalGraphPanel tests (Finding 2)** — All `vi.useFakeTimers()` calls in the new tests are wrapped in `try/finally` with `vi.useRealTimers()` in the finally block. `afterEach` at the describe level also calls `vi.useRealTimers()` as a belt-and-suspenders backstop.

3. **Inert no-links test (Finding 3)** — The new test overrides `getLocalGraph` per-test to return `edges: []`, saving and restoring the original implementation via try/finally. The `expect(noLinks!.getAttribute(...))` assertion is inside the live code path, not behind a dead branch.

4. **Flake-prone OmniSearch timers (Finding 4)** — Skeleton lifecycle cases are in an isolated `describe` block with `vi.useFakeTimers()` in `beforeEach` and `vi.useRealTimers()` in `afterEach`. `afterEach` was missing from the import in the TDD commit (`4980a1c`) and added in the fix commit (`9692980`) — correct red/green sequence. Debounce flushed via `vi.advanceTimersByTimeAsync(250)`.

5. **Silent scope cut (Finding 5)** — PR body contains "EditorPane noteLoading skeleton (Plan §6 PR-D touchpoint 5) is deferred to #372". Issue #372 is open, references #358 and the AsciiSkeleton primitive, and has a well-scoped acceptance section.

**Previously-clean items re-verified:**
- LCG initializer is `((s | 0) + 1) | 0`, not `(s | 0) || 1`.
- Skeleton glyphs are `░` (block fill) and space only.
- `aria-hidden="true"` is on the `<pre>` host.
- No `@keyframes` or animation in the skeleton styles.

**Verdict: APPROVE** — all findings addressed, no regressions observed. PR is ready for UAT.

— Aristotle

src/components/ascii/__tests__/AsciiWave.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    expect(src).toMatch(/@media\s*\(prefers-reduced-motion:\s*reduce\)/);
+    // Inside that block, the wave's animation must be set to none.
+    const media = src.match(
+      /@media\s*\(prefers-reduced-motion:\s*reduce\)[\s\S]*?\}\s*\}/,
+    );
+    expect(media).toBeTruthy();
+    expect(media![0]).toMatch(/animation:\s*none/);
+  });
+
+  it("source palette is block-fills and box-drawing only — no ASCII slashes in any content: declaration", () => {
```
herox215, line 59: `declarations` will always be an empty array here because `buildFrames` puts frame strings into inline CSS custom properties (`--vc-wave-f0` … `--vc-wave-f7` via `style={inlineStyle}`), not into literal `content: "…"` declarations in the stylesheet. The regex `/content:\s*"([^"]*)"/g` finds zero matches against this source, so the `for` loop never runs and the test trivially passes. A mutation that changes `CREST` to include `/` or a non-palette character would not be caught. Replace the regex with a direct assertion against the `CREST` constant: `expect(src).toMatch(/const CREST\s*=\s*["'][░▒▓█▓▒░ ]+["']/)` — or assert on the inline `style` attribute in the rendered DOM.

— Aristotle

src/components/ascii/AsciiWave.svelte
```diff
@@ -0,0 +1,115 @@
...
+      }
+      out.push(cells.join(""));
+    }
+    return out;
+  }
+
+  const frames = $derived(buildFrames(width));
+  // Frame 4 is roughly the middle of the cycle — used as the static
+  // fallback for prefers-reduced-motion.
+  const restFrame = $derived(frames[4] ?? "");
```
herox215, line 46: `restFrame = frames[4]` places the crest peak at column 20 of a 36-wide strip (computed: `center = Math.round(4 * 35 / 7) = 20`). The visual centre of a 36-cell strip is column 17–18. The spec calls this the "visually-centred crest" fallback for `prefers-reduced-motion`. At 4/8 (midpoint of the cycle), a symmetric spacing formula would put the peak at `Math.round((width - 1) / 2) = 17`. Frame index 4 with the current `(i * (safeW - 1)) / (FRAMES - 1)` formula lands at `4/7 * 35 ≈ 20`, not 17. Either rename the comment to "mid-cycle frame" (not "visually-centred"), or pick `frames[Math.floor(FRAMES / 2)]` only after verifying its actual center column. This is a comment / spec-alignment issue, not a crash.

— Aristotle

PR #377, opened 2026-04-29
feat: block references and heading anchors (#62)

### LINKED ISSUE #62: Block references [[note^blockid]]
## Summary

Implement Obsidian-style block references. Any block (paragraph, list item, heading, callout) can be tagged with a trailing `^blockid`. Other notes link to that block via `[[note^blockid]]`, which navigates to the target note and scrolls to the tagged block. The embed form `![[note^blockid]]` renders just that block inline with minimal surrounding context.

## Context

Today in the codebase:

- `src/components/Editor/wikiLink.ts` — the wiki-link regex is `/\[\[([^\]|]+?)(?:\|([^\]]*))?\]\]/g`. It does not recognize `#heading` or `^blockid` suffixes at all; the captured target is passed through `resolveTarget` which lowercases and looks up the stem. Anchors of any kind would currently be treated as part of the filename and fail resolution.
- `src/components/Editor/embedPlugin.ts` — the embed regex `/!\[\[([^\]|#]+?)(?:#([^\]|]+))?(?:\|([^\]]*))?\]\]/g` already captures an optional `#heading` group, but the plugin explicitly ignores it (see comment around the note-embed branch: "We intentionally ignore the #heading capture for now and render the entire target note; heading slicing is a follow-up."). `^blockid` is not captured at all.
- `src-tauri/src/commands/links.rs` — `get_resolved_links` returns `stem -> relPath` only; there is no block-id or heading metadata on the resolved-links payload. `compute_local_graph` / `compute_link_graph` operate on the link graph built by `indexer/link_graph.rs`.
- `src-tauri/src/indexer/parser.rs` — the markdown parser strips headings/bold and aggregates text for search; it does not emit a block-id table keyed by `^blockid`.
- `src/components/Editor/flashHighlight.ts` — already provides `scrollToMatch(view, from, to)` which scrolls and flashes a range. This is the scroll-to-block primitive we reuse.

Heading refs status: `[[note#Heading]]` is **not** functional for plain wiki-links (the `#` is not parsed by `wikiLink.ts`). In embed form the `#` is captured but ignored. Heading refs are therefore in the same state as block refs and are included in this issue's scope so the anchor model can be designed once.

## Acceptance Criteria

- [ ] Rust markdown parser recognizes trailing `^blockid` on paragraphs, list items, headings, and callouts (per CommonMark block types) and emits a per-file block-id table
- [ ] `get_resolved_links` (or a new sibling command) returns block-id info alongside the resolved path so the frontend can validate anchors without a second round-trip
- [ ] Wiki-link parser in `wikiLink.ts` recognizes the `^blockid` suffix, decorates the link based on whether the anchor resolves, and exposes the anchor on the click payload
- [ ] Clicking `[[note^blockid]]` opens the target note and scrolls to the tagged block (reusing `scrollToMatch` / `flashHighlight`)
- [ ] `![[note^blockid]]` embeds render only the tagged block plus minimal surrounding context (e.g. list-item with its children, callout with its body), not the full note
- [ ] Heading refs `[[note#Heading]]` and `![[note#Heading]]` work with the same anchor resolution pipeline (slug-based match, case-insensitive)
- [ ] Works in combination with aliases: `[[note^blockid|display text]]` and `[[note#Heading|display]]` render the alias and resolve the anchor
- [ ] Unresolved anchors (valid note, missing block-id/heading) render the link as unresolved rather than silently opening the note at the top

## Out of Scope

- Block-id autogeneration UX (auto-insert `^blockid` when the user creates a link to an untagged block)
- Transclusion edit-in-place (editing an embedded block from the embedding note)
- Cross-vault block references

## Effort estimate: M

herox215:
## Summary

Implements Obsidian-style **block references** (`[[note^blockid]]`, `![[note^blockid]]`) and **heading anchors** (`[[note#Heading]]`, `![[note#Heading]]`) end-to-end. Closes #62.

## What's new

**Rust**
- `indexer::anchors` extracts per-file block-id and heading anchors via `pulldown-cmark` offset iteration — paragraphs, list items, headings, callouts.
- `indexer::anchor_index` holds the per-vault map plus precomputed UTF-16 offsets for direct JS-string slicing on the frontend.
- New Tauri command `get_resolved_anchors`; locked-folder filtered like `get_resolved_links`.
- Rename-cascade regex preserves `#H` and `^id` suffixes.

**Frontend**
- `wikiLink.ts`: three-valued resolution (`resolved | anchor-missing | unresolved`), `parseLinkTarget` chokepoint, anchor in click event detail.
- `embedPlugin.ts`: regex extended to capture `^id`; slices cached content by `jsStart/jsEnd`; `BrokenEmbedWidget` for missing anchors.
- `EditorPane`: scrolls anchor matches through the existing `scrollStore` (now optionally carries a `range`); warning toast on anchor-missing.
- New CSS `cm-wikilink-unresolved-anchor` — `#92400E` light (~7:1 contrast), `--color-warning` dark.

## Why this shape

Per Socrates plan review, the design avoids five originally-blocking issues:
1. UTF-16 vs UTF-8 drift — Rust precomputes JS offsets.
2. Three-valued attribute routes correctly through the click handler (no false-create on anchor-missing).
3. Rename cascade preserves anchor suffixes.
4. `resolvedAnchors` keyed by rel_path (unambiguous).
5. `WIKI_EMBED_RE` capture-group shift accounted for in all call sites.

Vitruvius constraints all met: only one new CSS class, no new IPC inside the render loop, single-source slug rules in Rust, `BrokenEmbedWidget` reuse.

## Test plan
- [x] `cargo test --lib` — 381 passed, 0 failed
- [x] `npx vitest run` (excl. flaky pre-existing `largeFile` perf timer) — 1288 passed
- [x] `npm run typecheck` clean
- [ ] Manual UAT: open a vault with `[[note^id]]` content; verify decoration colors, anchor scroll, anchor-missing toast, embed slicing for blocks + heading sections, rename-cascade preserves suffixes.

## Out of scope (per ticket)
- Block-id autogeneration UX
- Transclusion edit-in-place
- Cross-vault references
- Anchor-aware backlinks

### DIFF
--- a/src-tauri/src/indexer/anchor_index.rs
+++ b/src-tauri/src/indexer/anchor_index.rs
@@ -13,11 +13,9 @@ use super::anchors::{build_anchor_key_set, AnchorKeySet, AnchorTable};
 
 #[derive(Debug, Default)]
 pub struct AnchorIndex {
-    by_path: HashMap<String, AnchorTable>,
-    /// Pre-built wire-format payload kept in lock-step with `by_path`. The
-    /// hot get_resolved_anchors path returns this without re-walking the
-    /// table on every IPC, mirroring the prebuilt-cache idea behind
-    /// `StemIndex` for the link-resolution path.
+    /// Pre-built wire-format payload — what `get_resolved_anchors` returns.
+    /// We don't keep `AnchorTable` separately: only the payload is read
+    /// downstream, and storing both doubles memory + risks drift.
     payload: HashMap<String, AnchorKeySet>,
 }
 
@@ -26,34 +24,25 @@ impl AnchorIndex {
         Self::default()
     }
 
-    /// Replace the anchor data for a single file. Idempotent: a second call
-    /// for the same path overwrites the first.
+    /// Replace the anchor data for a single file via raw inputs. Idempotent.
+    /// Used by the watcher path (`UpdateLinks`) which holds the AnchorTable
+    /// directly. Cold-start uses `update_file_with_payload` to skip the
+    /// table-keeping that would balloon the cold-start buffer.
     pub fn update_file(&mut self, rel_path: &str, content: &str, table: AnchorTable) {
         let payload = build_anchor_key_set(content, &table);
+        self.update_file_with_payload(rel_path, payload);
+    }
+
+    /// Replace the anchor data for a single file with a precomputed payload.
+    /// Used by the cold-start flush so we never carry raw file contents
+    /// past the per-file loop body.
+    pub fn update_file_with_payload(&mut self, rel_path: &str, payload: AnchorKeySet) {
         self.payload.insert(rel_path.to_string(), payload);
-        self.by_path.insert(rel_path.to_string(), table);
     }
 
     /// Drop every anchor entry for a file (used on delete and rename-old).
     pub fn remove_file(&mut self, rel_path: &str) {
-        self.by_path.remove(rel_path);
         self.payload.remove(rel_path);
-    }
-
-    /// Re-key an existing entry on rename — anchor data is keyed by rel_path
-    /// only (anchor content is independent of filename), so rename does not
-    /// invalidate the table itself. Cheaper than removing + re-extracting.
-    pub fn rename_file(&mut self, old_rel: &str, new_rel: &str) {
-        if let Some(table) = self.by_path.remove(old_rel) {
-            self.by_path.insert(new_rel.to_string(), table);
-        }
-        if let Some(payload) = self.payload.remove(old_rel) {
-            self.payload.insert(new_rel.to_string(), payload);
-        }
-    }
-
-    pub fn anchors_for(&self, rel_path: &str) -> Option<&AnchorTable> {
-        self.by_path.get(rel_path)
     }
 
     /// Snapshot for the wire payload — clones the precomputed map, no
@@ -69,38 +58,36 @@ mod tests {
     use crate::indexer::anchors::extract_anchors;
 
     #[test]
-    fn rename_moves_entry_without_recomputing() {
-        let mut idx = AnchorIndex::new();
-        let md = "para ^id1\n";
-        let table = extract_anchors(md);
-        idx.update_file("old.md", md, table);
-        idx.rename_file("old.md", "new.md");
-        assert!(idx.anchors_for("old.md").is_none());
-        assert!(idx.anchors_for("new.md").is_some());
-        let payload = idx.snapshot_payload();
-        assert!(payload.contains_key("new.md"));
-        assert!(!payload.contains_key("old.md"));
-    }
-
-    #[test]
     fn update_overwrites_previous_entry() {
         let mut idx = AnchorIndex::new();
         let md1 = "first ^a\n";
         let md2 = "second ^b\n";
         idx.update_file("note.md", md1, extract_anchors(md1));
         idx.update_file("note.md", md2, extract_anchors(md2));
-        let table = idx.anchors_for("note.md").unwrap();
-        assert_eq!(table.blocks.len(), 1);
-        assert_eq!(table.blocks[0].id, "b");
+        let payload = idx.snapshot_payload();
+        let entry = payload.get("note.md").expect("entry must exist");
+        assert_eq!(entry.blocks.len(), 1);
+        assert_eq!(entry.blocks[0].id, "b");
     }
 
     #[test]
-    fn remove_drops_both_table_and_payload() {
+    fn remove_drops_payload() {
         let mut idx = AnchorIndex::new();
         let md = "para ^id\n";
         idx.update_file("n.md", md, extract_anchors(md));
         idx.remove_file("n.md");
-        assert!(idx.anchors_for("n.md").is_none());
         assert!(!idx.snapshot_payload().contains_key("n.md"));
+    }
+
+    #[test]
+    fn update_file_with_payload_round_trips() {
+        let mut idx = AnchorIndex::new();
+        let md = "para ^id\n";
+        let table = extract_anchors(md);
+        let payload = super::super::anchors::build_anchor_key_set(md, &table);
+        idx.update_file_with_payload("n.md", payload);
+        let snap = idx.snapshot_payload();
+        assert!(snap.contains_key("n.md"));
+        assert_eq!(snap["n.md"].blocks[0].id, "id");
     }
 }
--- a/src-tauri/src/indexer/anchor_index.rs
+++ b/src-tauri/src/indexer/anchor_index.rs
@@ -1,0 +1,106 @@
+// Per-vault anchor index — maps `rel_path -> AnchorTable` for every indexed
+// `.md` file. Sibling to `LinkGraph` and `TagIndex` so the same Arc<Mutex<>>
+// + accessor pattern carries over.
+//
+// Two consumers:
+//   - `get_resolved_anchors` (Tauri command) returns a snapshot per vault open.
+//   - `resolve_anchor` is called by the rename-cascade Rust path so anchors
+//     survive a rename without a frontend round-trip.
+
+use std::collections::HashMap;
+
+use super::anchors::{build_anchor_key_set, AnchorKeySet, AnchorTable};
+
+#[derive(Debug, Default)]
+pub struct AnchorIndex {
+    by_path: HashMap<String, AnchorTable>,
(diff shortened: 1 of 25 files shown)


src/components/Editor/wikiLink.ts
```diff
@@ -65,16 +131,42 @@ export function stripKnownExt(target: string): string {
...
+ * lowercased value already produced by `parseLinkTarget`.
  */
+export function resolveAnchor(relPath: string, anchor: ParsedAnchor): AnchorEntry | null {
+  const set = resolvedAnchors.get(relPath);
+  if (!set) return null;
+  if (anchor.kind === "block") {
+    return set.blocks.find((b) => b.id === anchor.value) ?? null;
+  }
+  const target = anchor.value.toLowerCase();
+  return set.headings.find((h) => h.id.toLowerCase() === target) ?? null;
```
herox215: `anchor.value.toLowerCase()` produces `"my heading"` but `h.id` contains the GFM slug `"my-heading"` — every multi-word heading link silently resolves to `anchor-missing` and shows the warning toast. Fix: run `anchor.value` through the same `slugify` logic before comparing, e.g. `slugify(anchor.value) === h.id`.

— Aristotle

src/components/Editor/embedPlugin.ts
```diff
@@ -52,19 +53,44 @@ import CanvasRenderer from "../Canvas/CanvasRenderer.svelte";
...
+  relPath: string,
+  headingSlug: string | undefined,
+  blockId: string | undefined,
+): AnchorEntry | "missing" | null {
+  if (blockId !== undefined) {
+    const found = resolveAnchor(relPath, { kind: "block", value: blockId.toLowerCase() });
+    return found ?? "missing";
+  }
+  if (headingSlug !== undefined) {
+    const found = resolveAnchor(relPath, { kind: "heading", value: headingSlug });
```
herox215: `headingSlug` here is the raw regex capture (e.g. `"My Section"`), but `resolveAnchor` lowercases it to `"my section"` and compares against the stored GFM slug `"my-section"` — the same mismatch as in `wikiLink.ts:147`. All heading-anchored embeds (`![[Note#Multi Word]]`) render as `BrokenEmbedWidget`. Slugify `headingSlug` before passing it.

— Aristotle

src-tauri/src/indexer/mod.rs
```diff
@@ -429,6 +450,8 @@ impl IndexCoordinator {
             link_buffer.push((relative_path.clone(), links));
             let tags = extract_inline_tag_occurrences(&content);
             tag_buffer.push((relative_path.clone(), tags));
+            let anchor_table = extract_anchors(&content);
+            anchor_buffer.push((relative_path.clone(), content.clone(), anchor_table));
```
herox215: `content.clone()` buffers every file's full text until the bulk anchor-index update at line 532. At 100k notes × ~3.5 KB average, this is a second ~350 MB in-memory copy on top of the data already held by link/tag buffers — violating the 250 MB active-RAM budget. Call `build_anchor_key_set(&content, &anchor_table)` here instead and buffer only the resulting `AnchorKeySet` (a few hundred bytes per file).

— Aristotle

src-tauri/src/indexer/anchor_index.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Drop every anchor entry for a file (used on delete and rename-old).
+    pub fn remove_file(&mut self, rel_path: &str) {
+        self.by_path.remove(rel_path);
+        self.payload.remove(rel_path);
+    }
+
+    /// Re-key an existing entry on rename — anchor data is keyed by rel_path
+    /// only (anchor content is independent of filename), so rename does not
+    /// invalidate the table itself. Cheaper than removing + re-extracting.
+    pub fn rename_file(&mut self, old_rel: &str, new_rel: &str) {
```
herox215: `rename_file` is never called from the queue consumer — rename flows through `RemoveLinks` (old path) + `UpdateLinks` (new path), so this method is dead code. The header comment on line 7 claiming "`resolve_anchor` is called by the rename-cascade Rust path" is also false. Either wire `rename_file` into the consumer as an `IndexCmd::RenameLinks` variant, or delete it and fix the comment.

— Aristotle

src/components/Editor/__tests__/wikiLinkAnchorParse.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  it("handles `Note^id|alias` style — alias is stripped earlier so this never sees pipes", () => {
+    // `parseLinkTarget` operates on the regex group 1 capture; the alias
+    // pipe and what follows is captured separately and never reaches
+    // here. Pin behaviour for the `^` + no-pipe case.
+    expect(parseLinkTarget("Note^id")).toEqual({
+      stem: "Note",
+      anchor: { kind: "block", value: "id" },
+    });
+  });
+});
```
herox215, line 82: No test exercises `resolveAnchor` against a multi-word heading slug. A test seeding `resolvedAnchors` with `{ headings: [{ id: "my-section", ... }] }` then calling `resolveAnchor(path, { kind: "heading", value: "My Section" })` would have caught the mismatch at `wikiLink.ts:147`. Add a test here (or in a new `resolveAnchor.test.ts`) covering that path.

— Aristotle

src/lib/headingSlug.ts
```diff
@@ -0,0 +1,44 @@
...
+//   3. keep `\p{L}\p{N}` plus `-` and `_`; drop everything else
+//   4. trim leading / trailing `-`
+//
+// We do NOT collapse adjacent dashes from dropped characters because Rust
+// doesn't either — only whitespace creates dashes.
+
+/** Re-exported alphanumeric test that stays in sync with Rust's
+ * `char::is_alphanumeric`. JS `\p{L}\p{N}` matches the same Unicode
+ * categories. */
+const ALNUM_RE = /[\p{L}\p{N}]/u;
```
herox215: `\p{L}\p{N}` does NOT match all chars that Rust `char::is_alphanumeric()` keeps. Rust uses the Unicode `Alphabetic` property (a superset of General_Category=Letter), which includes combining marks like U+093E DEVANAGARI VOWEL SIGN AA and U+0902 DEVANAGARI SIGN ANUSVARA — category `Mc`/`Mn`, so `\p{L}` misses them and JS drops them while Rust retains them in the slug. For a heading like `"हिंदी"` (Hindi), the two sides produce different slugs silently. Fix: use `\p{Alphabetic}` instead of `\p{L}` in `ALNUM_RE`. The parity fixture needs a Devanagari or Arabic combining-vowel case to catch this regression class.

— Aristotle

src/components/Editor/embedPlugin.ts
```diff
@@ -52,19 +53,44 @@ import CanvasRenderer from "../Canvas/CanvasRenderer.svelte";
...
+ * Returns:
+ *   - `null`      — no anchor was requested (whole-note embed).
+ *   - `"missing"` — anchor was requested but doesn't resolve. Caller renders
+ *                   `BrokenEmbedWidget` with a "not found" label.
+ *   - `AnchorEntry` — anchor resolved; caller slices `cached.slice(jsStart,
+ *                     jsEnd)` to render only that block's content.
+ */
+function resolveEmbedAnchor(
+  relPath: string,
+  headingSlug: string | undefined,
```
herox215: Parameter named `headingSlug` receives raw heading text (e.g. `"Section Title"`), not a pre-slugified value — the slug is built inside `resolveAnchor`. The name is an active lie: a future reader will assume slugification is already done and either skip it or double-apply it. Rename to `headingText` (or `rawHeading`) to match what the embed regex actually captures.

— Aristotle

herox215:
Round-2 Aristotle findings addressed in commit 1b37037:

1. **Slug Unicode-property divergence** — Rust's `is_alphabetic` is the Unicode `Alphabetic` property (supports combining marks like U+093E, U+064F). TS regex switched from `\p{L}\p{N}` to `\p{Alphabetic}\p{N}` to match. Three new parity-fixture cases (Devanagari, Arabic, Latin combining) cover the previously-untested mark categories.

2. **`headingSlug` → `headingText` rename** in `resolveEmbedAnchor` to reflect the actual contract (raw regex capture, slugification handled inside `resolveAnchor`).

Both Rust (382 passed) and frontend (1304 passed minus pre-existing flaky `largeFile` perf timer) green. Per the workflow this hits the 2-iteration limit; ready for UAT.

Issue #378, opened 2026-04-29
rename-cascade dialog races with watcher tree re-flatten


herox215:
## Repro

1. Open a vault with at least one note that has backlinks (e.g. \`Welcome.md\` referenced by \`[[Welcome]]\` from other files).
2. Right-click the note in the sidebar → Rename.
3. Type a new name and press Enter.

## Observed

The note is renamed on disk and in the tree, but the cascade-confirmation dialog (\"N Links in M Dateien werden aktualisiert. Fortfahren?\") never appears, so the backlinks are never rewritten — they end up dangling.

## Root cause

\`InlineRename.handleConfirm\` calls \`renameFile()\` IPC. The disk rename triggers the watcher, which flattens the tree and re-renders \`Sidebar\`. The \`{#each}\` block keys rows by \`row.path\`, so the renamed row destroys the original \`TreeRow\` and mounts a new one. But the \`renameFile\` promise was awaited inside the original \`TreeRow\`'s \`InlineRename\` component — when it resolves and calls \`onConfirm\`, the closure mutates \`pendingRename\` on a now-unmounted component. The new \`TreeRow\` doesn't carry that state, so the dialog never paints.

Direct IPC probing confirms \`rename_file\` correctly returns \`linkCount: 3\` for the test fixture — the backend is fine; the regression is purely in per-row state ownership.

## Fix candidates

1. **Lift \`pendingRename\` to \`Sidebar\`.** Single owner survives row remount. Cleanest architecturally.
2. **Delay the tree re-flatten until after \`pendingRename\` is committed.** Fragile — couples sidebar refresh to a sibling component's transient state.
3. **Hold the disk rename until the user confirms in the dialog.** Changes UX (rename-then-confirm-cascade → confirm-cascade-then-rename).

Recommendation: (1).

## Test

\`e2e/specs/rename-cascade.spec.ts\` is currently \`describe.skip\` with a pointer back here. The Vitest unit suite in \`commands/links::tests\` still covers the rename regex (\`#62\` left those tests passing for \`#H\` and \`^id\` suffix preservation).

PR #379, opened 2026-04-29
test(e2e): repair stale specs + 5 gap-fill tests

herox215:
## Summary

Two-part PR triggered by a full E2E suite run:

### Part 1 — Repair four stale specs (no app changes)

After the merge of #62 / #377 the full suite revealed four pre-existing test failures, none caused by the anchor work:

- **rename-cascade.spec.ts** — selector \`[aria-labelledby=\"rename-heading\"]\` has not matched since #336 templated the dialog ID per row. Switched to a prefix match. Discovered while running: the cascade dialog itself has a watcher race (TreeRow unmounts before \`onConfirm\` mounts the dialog). \`describe.skip\`'d with a pointer to **#378** (filed). Vitest unit suite still covers the rename rewrite contract.
- **canvas-file-refs.spec.ts** — #364 routed canvas text nodes through the shared markdown renderer, so wiki-links now carry \`.vc-reading-wikilink\` instead of the legacy \`.vc-canvas-link*\`. Selectors updated; click uses a JS-dispatched mousedown (canvas delegation listens on mousedown).
- **canvas-context-menus.spec.ts** — #362 turned \"Add text node\" into an inline shape-picker expander; the node only appears once a shape row is also clicked. Test now does both clicks via JS dispatch.
- **encrypted-folders.spec.ts** — \`browser.performActions\` with \`pointerMove { origin: <element> }\` errors on this WebKitWebDriver. Replaced the helper with a JS-dispatched contextmenu event. Also fixed the stale \`.vc-tree\` dismiss selector (renamed to \`.vc-tree-root\` after #253 → use Escape instead).

### Part 2 — Five new gap-fill specs

Audit-driven additions, each riding an existing fixture:

1. **Lock all now (Settings → Security)** — extends \`encrypted-folders.spec.ts\`.
2. **Monospace font selection** — extends \`settings-appearance.spec.ts\`.
3. **Hotkey conflict detection** — extends \`hotkeys.spec.ts\`.
4. **Unresolved-link create from Outgoing Links panel** — extends \`outgoing-links.spec.ts\`.
5. **Daily-note template seeding** — extends \`daily-notes.spec.ts\` plus a settings-cleanup helper so localStorage pollution between runs no longer leaks.

## Test plan

- [x] \`npm run test:e2e\` — 63 specs, all 63 passing
- [x] \`cargo test --lib\` — 382 passing (no Rust code touched)

## Out of scope

- Fixing the rename-cascade race (issue #378) — needs lifting \`pendingRename\` to Sidebar, separate PR.
- Auto-lock timeout E2E — requires time-stub plumbing not yet present.

### DIFF
--- a/e2e/specs/canvas-context-menus.spec.ts
+++ b/e2e/specs/canvas-context-menus.spec.ts
@@ -2,7 +2,7 @@ import fs from "node:fs";
 import path from "node:path";
 import { createTestVault, type TestVault } from "../helpers/vault.js";
 import { openVaultInApp } from "../helpers/open-vault.js";
-import { textOf } from "../helpers/text.js";
+import { textOf, textsOf } from "../helpers/text.js";
 
 /**
  * E2E for #164 — canvas context menus.
@@ -66,9 +66,14 @@ describe("Canvas context menus (#164)", () => {
   }
 
   async function waitForActiveTab(label: string, timeout = 5000) {
-    const activeLabel = await browser.$(".vc-tab--active .vc-tab-label");
+    // Re-query each iteration — the `.vc-tab--active` element is replaced
+    // on every tab switch and a handle captured before the transition
+    // would silently keep returning the old label.
     await browser.waitUntil(
-      async () => ((await activeLabel.getProperty("textContent")) as string).includes(label),
+      async () => {
+        const labels = await textsOf(await browser.$$(".vc-tab--active .vc-tab-label"));
+        return labels.some((l) => l.includes(label));
+      },
       { timeout, timeoutMsg: `active tab never switched to "${label}"` },
     );
   }
--- a/e2e/specs/canvas-context-menus.spec.ts
+++ b/e2e/specs/canvas-context-menus.spec.ts
@@ -136,18 +136,35 @@ describe("Canvas context menus (#164)", () => {
     const menu = await browser.$(".vc-context-menu");
     await menu.waitForDisplayed({ timeout: 3000 });
 
-    // Find "Add text node" entry and click it.
-    const items = await browser.$$(".vc-context-menu .vc-context-item");
-    let clicked = false;
-    for (const it of items) {
-      const label = (await textOf(it)).trim();
-      if (label === "Add text node") {
-        await it.click();
-        clicked = true;
-        break;
+    // #362: "Add text node" opens an inline shape picker — the node is
+    // only created when a shape is then chosen. Two clicks total, both
+    // via JS dispatch (overlay-click guard).
+    const expanded = await browser.execute(() => {
+      const items = document.querySelectorAll<HTMLElement>(
+        ".vc-context-menu .vc-context-item",
+      );
+      for (const el of Array.from(items)) {
+        if ((el.textContent ?? "").trim() === "Add text node") {
+          el.click();
+          return true;
+        }
       }
-    }
-    expect(clicked).toBe(true);
+      return false;
+    });
+    expect(expanded).toBe(true);
+
+    const picked = await browser.waitUntil(
+      async () => {
+        return (await browser.execute(() => {
+          const row = document.querySelector<HTMLElement>(".vc-shape-picker-row");
+          if (!row) return false;
+          row.click();
+          return true;
+        })) as boolean;
+      },
+      { timeout: 3000, timeoutMsg: "shape picker never expanded after Add text node" },
+    );
+    expect(picked).toBe(true);
 
     // A text node should materialise in the active canvas viewport.
     await browser.waitUntil(
@@ -219,19 +236,21 @@ describe("Canvas context menus (#164)", () => {
       );
     });
 
-    // Menu → click "Add file node…".
+    // Menu → click "Add file node…" via JS dispatch (overlay-click guard).
     const menu = await browser.$(".vc-context-menu");
     await menu.waitForDisplayed({ timeout: 3000 });
-    const items = await browser.$$(".vc-context-menu .vc-context-item");
-    let clicked = false;
-    for (const it of items) {
-      const label = (await textOf(it)).trim();
-      if (label === "Add file node…") {
-        await it.click();
-        clicked = true;
-        break;
+    const clicked = await browser.execute((label: string) => {
+      const items = document.querySelectorAll<HTMLElement>(
+        ".vc-context-menu .vc-context-item",
+      );
+      for (const el of Array.from(items)) {
+        if ((el.textContent ?? "").trim() === label) {
+          el.click();
+          return true;
+        }
       }
-    }
+      return false;
+    }, "Add file node…");
     expect(clicked).toBe(true);
 
     // QuickSwitcher opens — type the target file name and press Enter.
--- a/e2e/specs/canvas-context-menus.spec.ts
+++ b/e2e/specs/canvas-context-menus.spec.ts
@@ -153,17 +153,19 @@ describe("Canvas context menus (#164)", () => {
     });
     expect(expanded).toBe(true);
 
-    const picked = await browser.waitUntil(
-      async () => {
-        return (await browser.execute(() => {
-          const row = document.querySelector<HTMLElement>(".vc-shape-picker-row");
-          if (!row) return false;
-          row.click();
-          return true;
-        })) as boolean;
-      },
+    // Wait for the picker to appear, THEN click — never click inside the
+    // poll predicate (that turns the wait into N rapid-fire clicks).
+    await browser.waitUntil(
+      async () =>
+        (await browser.$$(".vc-shape-picker-row")).length > 0,
       { timeout: 3000, timeoutMsg: "shape picker never expanded after Add text node" },
     );
+    const picked = await browser.execute(() => {
+      const row = document.querySelector<HTMLElement>(".vc-shape-picker-row");
+      if (!row) return false;
+      row.click();
+      return true;
+    });
     expect(picked).toBe(true);
 
     // A text node should materialise in the active canvas viewport.
--- a/e2e/specs/canvas-file-refs.spec.ts
+++ b/e2e/specs/canvas-file-refs.spec.ts
@@ -2,7 +2,7 @@ import fs from "node:fs";
 import path from "node:path";
 import { createTestVault, type TestVault } from "../helpers/vault.js";
 import { openVaultInApp } from "../helpers/open-vault.js";
-import { textOf } from "../helpers/text.js";
+import { textOf, textsOf } from "../helpers/text.js";
 
 /**
  * E2E coverage for #162 — file-reference nodes on canvas:
@@ -80,9 +80,13 @@ describe("Canvas file-reference nodes (#162)", () => {
   }
 
   async function waitForActiveTab(label: string, timeout = 5000) {
-    const activeLabel = await browser.$(".vc-tab--active .vc-tab-label");
+    // Re-query each iteration: `.vc-tab--active` is replaced on every tab
+    // switch, so a handle captured before the transition becomes stale.
     await browser.waitUntil(
-      async () => ((await activeLabel.getProperty("textContent")) as string).includes(label),
+      async () => {
+        const labels = await textsOf(await browser.$$(".vc-tab--active .vc-tab-label"));
+        return labels.some((l) => l.includes(label));
+      },
       { timeout, timeoutMsg: `active tab never switched to "${label}"` },
     );
   }
--- a/e2e/specs/canvas-file-refs.spec.ts
+++ b/e2e/specs/canvas-file-refs.spec.ts
@@ -131,15 +131,13 @@ describe("Canvas file-reference nodes (#162)", () => {
   });
 
   it("clicking the wiki-link opens the target note in a new tab", async () => {
-    // Still on Linked.canvas from the previous test. Use a real mousedown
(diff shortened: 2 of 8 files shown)


e2e/specs/canvas-context-menus.spec.ts
```diff
@@ -136,18 +136,35 @@ describe("Canvas context menus (#164)", () => {
...
+          return true;
+        }
       }
-    }
-    expect(clicked).toBe(true);
+      return false;
+    });
+    expect(expanded).toBe(true);
+
+    const picked = await browser.waitUntil(
```
herox215: `waitUntil` here is not polling a condition — it is clicking `.vc-shape-picker-row` on every iteration until the row exists. If the app renders the row and then immediately dismisses it (animation, focus trap, etc.) the click fires repeatedly on an element that may already be gone, leaving the canvas in an unexpected state before the next assertion. Separate the existence poll from the click: `waitUntil` until `.vc-shape-picker-row` exists, then perform a single JS-dispatched click outside the `waitUntil` callback.

— Aristotle

e2e/specs/canvas-file-refs.spec.ts
```diff
@@ -114,23 +114,36 @@ describe("Canvas file-reference nodes (#162)", () => {
...
+    // Still on Linked.canvas from the previous test. Use a real mousedown
+    // — the canvas click delegation listens on `mousedown` paired with the
+    // [data-wiki-target] hit-testing in CanvasRenderer.
+    const clicked = await browser.execute(() => {
+      const link = document.querySelector<HTMLElement>(".vc-reading-wikilink--resolved");
+      if (!link) return false;
+      link.dispatchEvent(
+        new MouseEvent("mousedown", { bubbles: true, cancelable: true, button: 0 }),
+      );
+      link.click();
```
herox215, line 145: Dispatching `mousedown` followed immediately by `link.click()` fires two separate events into the canvas delegation handler. If `mousedown` alone is sufficient to trigger navigation, the subsequent synthetic `click` will attempt to open the note a second time, potentially creating a duplicate tab or a no-op depending on idempotency. The comment says the handler listens on `mousedown`, so drop `link.click()`.

— Aristotle

e2e/specs/hotkeys.spec.ts
```diff
@@ -77,4 +77,52 @@ describe("Keyboard shortcut rebinding", () => {
...
+      for (let i = 0; i < recordBtns.length; i++) {
+        const row = recordBtns[i]!.closest("tr");
+        const action = row?.querySelector(".vc-shortcut-action")?.textContent?.trim() ?? "";
+        if (action && !action.toLowerCase().includes("new note")) return i;
+      }
+      return -1;
+    });
+    expect(targetIndex).toBeGreaterThanOrEqual(0);
+
+    const recordBtn = (await browser.$$('[data-testid="shortcut-record-btn"]'))[targetIndex]!;
```
herox215: `targetIndex` is resolved inside `browser.execute` (in the page context) and then immediately used to index into `browser.$$()` (a separate WebDriver call). Between those two calls the DOM can repaint and the record-button NodeList can reorder, making `recordBtn` point at the wrong element. Capture the target buttons `data-testid` or another stable attribute inside the same `execute` call, then query it directly with `browser.$()`.

— Aristotle

e2e/specs/hotkeys.spec.ts
```diff
@@ -77,4 +77,52 @@ describe("Keyboard shortcut rebinding", () => {
...
+    await recordBtn.click();
+    await browser.$('[data-testid="shortcut-recording"]').waitForDisplayed({ timeout: 2000 });
+
+    // The recording listener sits on `<svelte:window>`. Dispatch a
+    // synthetic keydown that hotkeyFromEvent maps to { meta: true,
+    // key: "n" } — matches NEW_NOTE's default → conflict path is taken.
+    await browser.execute(() => {
```
herox215, line 113: The comment says the event maps to `{ meta: true, key: "n" }` but the dispatched `KeyboardEvent` sets `ctrlKey: true`, not `metaKey: true`. On macOS, `Cmd+N` (the NEW_NOTE default) requires `metaKey`. If `hotkeyFromEvent` normalises Ctrl→Meta on macOS this still works, but if it does not, the conflict modal will never appear on a Mac runner and the test passes vacuously. Either fix the comment or dispatch both `ctrlKey` and `metaKey`.

— Aristotle

e2e/specs/daily-notes.spec.ts
```diff
@@ -94,4 +108,86 @@ describe("Daily notes", () => {
...
+    await browser.waitUntil(
+      () => Promise.resolve(fs.existsSync(newAbs)),
+      { timeout: 5000, timeoutMsg: `${newAbs} never written` },
+    );
+
+    // Daily note must contain the template marker — proves the seeding
+    // path did the read+write; an empty file means the readFile failed
+    // silently and the user gets a blank note (per AC, that fallback
+    // must not block creation, but the happy path shouldn't take it).
+    await browser.waitUntil(
```
herox215, line 149: `waitUntil` at line 159 confirms the file exists (line 150-153 already did that), then immediately reads it and checks the content, but the two polls have separate 5 s and 3 s budgets with a gap between them. On a slow CI box the file can exist but still be partially flushed when this second poll starts, causing a spurious false from `includes()`. Collapse into a single `waitUntil` that checks both existence and content: `fs.existsSync(newAbs) && fs.readFileSync(newAbs, "utf-8").includes("e2e-daily-template-seed")`.

— Aristotle

e2e/specs/daily-notes.spec.ts
```diff
@@ -94,4 +108,86 @@ describe("Daily notes", () => {
...
+          return Promise.resolve(false);
+        }
+      },
+      { timeout: 3000, timeoutMsg: "daily note was never seeded with the template content" },
+    );
+
+    // Reset daily-notes settings — `settingsStore` is global (not
+    // vault-scoped), so leaving "Daily-Tpl" + a template path in place
+    // would change what the *next* spec session sees on cold start.
+    const reopenBtn = await browser.$('button[aria-label="Einstellungen"]');
```
herox215: The settings-reset block (open settings → clear folder + template → close) appears in full three times: once inside `setDailyFormat` (lines 48-63) and twice inside the new template test (lines 174-191). This is the pattern that was already identified as the cross-spec contamination vector. Extract it into a shared `resetDailySettings()` helper, otherwise a future contributor copying the pattern will miss one of the three reset sites and reintroduce localStorage pollution.

— Aristotle

e2e/specs/encrypted-folders.spec.ts
```diff
@@ -187,4 +194,56 @@ describe("Encrypted folders (#345)", () => {
...
+    await settingsBtn.click();
+    await browser.$('[data-testid="settings-modal"]').waitForDisplayed({ timeout: 3000 });
+
+    const lockAll = await browser.$('[data-testid="settings-lock-all"]');
+    await lockAll.waitForDisplayed({ timeout: 3000 });
+    await lockAll.click();
+
+    // Close the modal so the sidebar is observable again. The button is
+    // disabled when no encrypted folders are present, so the lock-all
+    // round-trip must have completed by the time we close.
```
herox215: The comment "The button is disabled when no encrypted folders are present, so the lock-all round-trip must have completed by the time we close" is the actual correctness invariant being relied on, but it is never asserted. If the app does not disable the button, or disables it prematurely, the test closes the modal before the lock command fires and the final `waitUntil` on the locked icon becomes a race. Add `expect(await lockAll.getAttribute("disabled")).not.toBeNull()` after `lockAll.click()` and before `close.click()`.

— Aristotle

e2e/specs/outgoing-links.spec.ts
```diff
@@ -113,4 +113,67 @@ describe("Outgoing Links panel", () => {
...
+        for (const el of unresolved) {
+          if ((await textOf(el)).includes(dangling)) return true;
+        }
+        return false;
+      },
+      { timeout: 5000, timeoutMsg: "unresolved row never appeared for the new target" },
+    );
+
+    // Click the unresolved row → click-to-create at vault root.
+    await browser.execute((target: string) => {
```
herox215: `match?.click()` is fire-and-forget inside `browser.execute` — if `match` is undefined the call silently no-ops, and the test proceeds to assert the tree and the active tab as if creation had been requested. The preceding `waitUntil` only proves the row exists in the DOM at that moment; a render cycle could still remove it before `execute` runs. Use `browser.$(".vc-outlink-row--unresolved=<dangling>").click()` or throw when `match` is null rather than swallowing the miss with optional chaining.

— Aristotle

e2e/specs/outgoing-links.spec.ts
```diff
@@ -113,4 +113,67 @@ describe("Outgoing Links panel", () => {
...
+      const rows = Array.from(
+        document.querySelectorAll<HTMLElement>(".vc-outlink-row--unresolved"),
+      );
+      const match = rows.find((r) => (r.textContent ?? "").includes(target));
+      match?.click();
+    }, dangling);
+
+    // The new file appears in the tree and a tab opens for it.
+    await browser.waitUntil(
```
herox215, line 189: The tree assertion checks for `"${dangling}.md"` (with `.md` appended), but the spec never verifies what filename the "click-to-create" path actually writes. If the panel creates the file as `"${dangling}.md"` the test is correct; if it writes `"${dangling}"` (no extension) the `names.includes()` check will always return false and `waitUntil` will time out rather than fail with a useful message. Add a disk-level check (`fs.existsSync`) for both candidates and fail fast with an explicit message naming the actual file, so a regression here produces a clear diagnostic rather than a 5 s timeout.

— Aristotle

e2e/specs/canvas-file-refs.spec.ts
```diff
@@ -114,23 +114,34 @@ describe("Canvas file-reference nodes (#162)", () => {
...
+    // could open the target tab twice or trip dedup logic.
+    const clicked = await browser.execute(() => {
+      const link = document.querySelector<HTMLElement>(".vc-reading-wikilink--resolved");
+      if (!link) return false;
+      link.click();
+      return true;
+    });
+    expect(clicked).toBe(true);
 
     await waitForActiveTab("Welcome.md");
```
herox215, line 150: `waitForActiveTab` captures `.vc-tab--active .vc-tab-label` before any tab switch and holds that stale handle across all `waitUntil` ticks. If the active-tab element remounts between call and first poll, every subsequent `getProperty("textContent")` returns an empty string and the wait times out with a misleading message. The outgoing-links PR already fixed this pattern — re-query inside the predicate: `const labels = await textsOf(await browser.$$(".vc-tab--active .vc-tab-label")); return labels.some(l => l.includes(label));`. Same fix is needed in `canvas-context-menus.spec.ts` line 69.

— Aristotle

e2e/specs/outgoing-links.spec.ts
```diff
@@ -113,4 +113,81 @@ describe("Outgoing Links panel", () => {
...
+    // requires reaching into CM6 internals that aren't exposed on
+    // `.cm-editor`.
+    await browser.executeAsync(
+      (target: string, done: () => void) => {
+        window.__e2e__!.typeInActiveEditor(`\n[[${target}]]\n`).then(() => done());
+      },
+      dangling,
+    );
+
+    await activateOutgoingSubtab();
```
herox215, line 157: Missing `await ensureRightSidebar()` before this call. The first three tests gate on it; this new test does not. If a prior spec collapses the right sidebar, `[aria-label="Outgoing Links"][role="tab"]` never exists in the DOM and `activateOutgoingSubtab` silently no-ops (the `?.click()` inside swallows the miss), then the `waitUntil` for `.vc-outlink-row--unresolved` times out 5 s later with a misleading message.

— Aristotle

e2e/specs/daily-notes.spec.ts
```diff
@@ -94,4 +122,43 @@ describe("Daily notes", () => {
...
+          return Promise.resolve(content.includes("e2e-daily-template-seed"));
+        } catch {
+          return Promise.resolve(false);
+        }
+      },
+      { timeout: 8000, timeoutMsg: `daily note ${newAbs} never carried the template marker` },
+    );
+
+    // Reset settings — see helper docstring.
+    await applyDailySettings({ folder: "", template: "" });
```
herox215: Reset omits `format`. The docstring explicitly warns that omitting a field inherits whatever the previous session left. If this test runs first in a fresh session (or after a test that cleared the format), the next `todayFilename()`-dependent test will hit the default `YYYY/MM/DD` format and create a nested-folder path instead of a flat file, causing a flaky failure nowhere near this line. Add `format: "YYYY-MM-DD"` to match `setDailyFormat`'s reset contract.

— Aristotle

herox215:
Round-2 fixes in commit `14755ac` — all 4 inline findings addressed:

1. Stale-handle pattern propagated to `canvas-file-refs` + `canvas-context-menus` waitForActiveTab helpers.
2. Pre-existing positional-index race in outgoing-links "navigates" test folded into a single `browser.execute` block.
3. `ensureRightSidebar()` guard added before `activateOutgoingSubtab()` in the new outgoing-links test.
4. Daily-notes end-of-test reset now also clears `format`.

Per the workflow's 2-iteration cap, this is the final round. Suite (63 specs) green. Merging to main.

PR #381, opened 2026-04-30
fix(sidebar): rename cascade dialog races with watcher tree re-flatten (#378)

### LINKED ISSUE #378: rename-cascade dialog races with watcher tree re-flatten

herox215:
Fixes #378.

## Summary
- Lift `pendingRename` / `pendingMove` cascade state from per-row `TreeRow` up to `Sidebar`. The Sidebar is the single component that survives the watcher-driven tree re-flatten — per-row ownership lost the state the moment `loadRoot` re-keyed `{#each}`, so the cascade-confirm dialog never painted and backlinks ended up dangling.
- `TreeRow` hands the cascade request up via two new callback props (`onRequestRenameCascade` / `onRequestMoveCascade`) the moment the IPC await resolves — *synchronously*, no further `await`. The original `getBacklinks` call (the second await that crossed the unmount window for renames) now runs on Sidebar's always-mounted lifetime instead of TreeRow's per-row one.
- Cascade dialogs now render at Sidebar level. `aria-labelledby` is a stable `rename-heading` / `move-heading` (no per-row suffix) since only one dialog can be open at a time. The existing prefix-match selectors keep working.
- Modal precedence: an incoming cascade request is ignored if either `pendingRename` or `pendingMove` is already set — single-modal expectation.

## Why per-row was wrong
1. User confirms inline rename → `renameFile` IPC mutates disk
2. Backend emits synthetic `FILE_CHANGED_EVENT(kind: rename)` (because `write_ignore` suppressed the natural one)
3. Sidebar's watcher handler runs `invalidateAllFolders()` + `loadRoot()` → `flatRows` recomputes → `{#each windowRows as { row } (row.path)}` re-keys → original `TreeRow` for the renamed file unmounts under the old path, new one mounts under the new path
4. `InlineRename`'s post-IPC continuation calls `onConfirm` on a closure pointing at the dead `TreeRow` instance → mutating `pendingRename` on a destroyed component → dialog never paints

## Test plan
- [x] New Vitest unit `Sidebar.cascade.test.ts` simulates the race: triggers a rename via the IPC mock, fires the watcher event programmatically mid-flight, asserts the dialog stays mounted *and* the original row is gone (`{#each}` re-keyed) while the renamed row is mounted under the new path. Fails on `main`, passes on this branch.
- [x] All 18 sidebar Vitest files green (41 tests).
- [x] `svelte-check` clean for the Sidebar workspace.
- [ ] e2e `rename-cascade.spec.ts` un-skipped — verify on UAT (WebDriver was not running locally).
- [ ] UAT: open a vault with a note that has 1+ backlinks, rename via context menu, confirm dialog appears with German body, click *Aktualisieren* → backlinks rewritten, no dangling links remain.

## Files
- `src/types/sidebar.ts` (new) — `RenameCascadeRequest` / `MoveCascadeRequest` / `PendingRename` / `PendingMove`
- `src/components/Sidebar/Sidebar.svelte` — owns cascade state + dialogs + `getBacklinks`
- `src/components/Sidebar/TreeRow.svelte` — drops cascade state + dialogs, adds two callback props
- `src/components/Sidebar/__tests__/Sidebar.cascade.test.ts` (new) — regression guard
- `e2e/specs/rename-cascade.spec.ts` — un-skipped, comment updated
- 4 existing TreeRow tests — added the two new required props to fixtures

### DIFF
--- a/src/components/Sidebar/Sidebar.svelte
+++ b/src/components/Sidebar/Sidebar.svelte
@@ -38,7 +38,7 @@
   import { resolvedLinksStore } from "../../store/resolvedLinksStore";
   import type {
     RenameCascadeRequest,
-    MoveCascadeRequest,
+    MoveDropRequest,
     PendingRename,
     PendingMove,
   } from "../../types/sidebar";
@@ -561,15 +561,64 @@
     pendingRename = { ...req, fileCount };
   }
 
-  function handleRequestMoveCascade(req: MoveCascadeRequest) {
+  async function handleRequestMoveCascade(req: MoveDropRequest) {
+    // Modal precedence — same rule as rename cascade.
     if (pendingRename || pendingMove) return;
-    pendingMove = req;
+    const vault = $vaultStore.currentPath;
+    if (!vault) return;
+    const { sourcePath, targetDirPath } = req;
+    const sourceRelPath = sourcePath.startsWith(vault + "/")
+      ? sourcePath.slice(vault.length + 1)
+      : sourcePath;
+    const sourceFilename = sourcePath.split("/").pop() ?? sourcePath;
+    const newAbsPath = targetDirPath + "/" + sourceFilename;
+    const newRelPath = newAbsPath.startsWith(vault + "/")
+      ? newAbsPath.slice(vault.length + 1)
+      : newAbsPath;
+
+    let linkCount = 0;
+    let fileCount = 0;
+    try {
+      const backlinks = await getBacklinks(sourceRelPath);
+      linkCount = backlinks.length;
+      fileCount = new Set(backlinks.map((b) => b.sourcePath)).size;
+    } catch {
+      /* proceed without cascade */
+    }
+
+    if (linkCount > 0) {
+      pendingMove = {
+        sourcePath,
+        targetDirPath,
+        sourceRelPath,
+        newRelPath,
+        linkCount,
+        fileCount,
+      };
+      return;
+    }
+
+    // No-cascade direct move — was previously inside TreeRow.handleDrop.
+    try {
+      await moveFile(sourcePath, targetDirPath);
+      void bookmarksStore.renamePath(sourceRelPath, newRelPath, vault);
+      await refreshFolder(targetDirPath);
+      const parent = parentOf(sourcePath);
+      if (parent) await refreshFolder(parent);
+      resolvedLinksStore.requestReload();
+    } catch (err) {
+      const ve = isVaultError(err) ? err : { kind: "Io" as const, message: String(err), data: null };
+      toastStore.push({ variant: "error", message: vaultErrorCopy(ve) });
+    }
   }
 
   async function confirmRenameWithLinks() {
     if (!pendingRename) return;
+    // Snapshot but DO NOT clear pendingRename yet — clearing before the
+    // IPC settles would let a second cascade request slip past the modal
+    // guard in handleRequest*Cascade and trigger concurrent link rewrites.
+    // Keep the dialog open while applying; clear only when the IPC resolves.
     const { oldPath, newPath, oldRelPath, newRelPath } = pendingRename;
-    pendingRename = null;
     handlePathChanged(oldPath, newPath);
     resolvedLinksStore.requestReload();
     try {
@@ -589,6 +638,8 @@
         variant: "error",
         message: "Links konnten nicht aktualisiert werden.",
       });
+    } finally {
+      pendingRename = null;
     }
   }
 
@@ -598,8 +649,9 @@
 
   async function confirmMoveWithLinks() {
     if (!pendingMove) return;
+    // Same re-entrancy reasoning as confirmRenameWithLinks — keep the modal
+    // guard armed across the whole IPC chain.
     const { sourcePath, targetDirPath, sourceRelPath, newRelPath } = pendingMove;
-    pendingMove = null;
     try {
       await moveFile(sourcePath, targetDirPath);
       const vaultForBookmarks = $vaultStore.currentPath;
@@ -624,6 +676,8 @@
     } catch (err) {
       const ve = isVaultError(err) ? err : { kind: "Io" as const, message: String(err), data: null };
       toastStore.push({ variant: "error", message: vaultErrorCopy(ve) });
+    } finally {
+      pendingMove = null;
     }
   }
 
--- a/src/components/Sidebar/Sidebar.svelte
+++ b/src/components/Sidebar/Sidebar.svelte
@@ -544,71 +544,88 @@
   // #378 — cascade-request handlers from TreeRow. The row hands the request
   // up the moment the IPC await resolves (synchronously, no further await),
   // so the closure capture cannot race the row's unmount. Sidebar then runs
-  // any post-IPC work (getBacklinks for fileCount, dialog rendering,
-  // updateLinksAfterRename on confirm) on its own always-mounted lifetime.
+  // any post-IPC work on its own always-mounted lifetime.
+  //
+  // `cascadeClaiming` is a synchronous slot reservation. The handler does an
+  // `await getBacklinks(...)` before it can populate the full `pendingRename`
+  // / `pendingMove` shape; checking only those state fields would let a second
+  // request pass the guard and race for the slot during that await window.
+  // `cascadeClaiming` is set before the await and the synthetic guard reads
+  // it, closing the re-entrancy window without flashing a half-populated
+  // dialog.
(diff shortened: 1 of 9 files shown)


src/components/Sidebar/TreeRow.svelte
```diff
@@ -444,14 +419,14 @@
     }
 
     if (linkCount > 0) {
-      pendingMove = {
+      onRequestMoveCascade({
```
herox215: The `await getBacklinks(sourceRelPath)` call (9 lines above, in `handleDrop`) is still there before this `onRequestMoveCascade` call — the identical unmount race that was fixed for rename survives for drag-drop. A watcher event can re-key `{#each}` during that await and destroy this `TreeRow` instance before this line is reached. Move the `getBacklinks` call into `handleRequestMoveCascade` in Sidebar (same pattern used by `handleRequestRenameCascade`), drop `fileCount` from `MoveCascadeRequest`, and pass `linkCount` alone from `handleDrop`.

— Aristotle

src/components/Sidebar/Sidebar.svelte
```diff
@@ -519,6 +541,96 @@
...
+
+  function handleRequestMoveCascade(req: MoveCascadeRequest) {
+    if (pendingRename || pendingMove) return;
+    pendingMove = req;
+  }
+
+  async function confirmRenameWithLinks() {
+    if (!pendingRename) return;
+    const { oldPath, newPath, oldRelPath, newRelPath } = pendingRename;
+    pendingRename = null;
```
herox215: `pendingRename = null` is set before `await updateLinksAfterRename` — the modal guard (`if (pendingRename || pendingMove) return`) is cleared before the async work finishes. A rapid second rename that lands during the `updateLinksAfterRename` await will pass the guard and open a second dialog, potentially triggering concurrent link rewrites. Null the state only after the await resolves (or keep a separate boolean `isConfirmInFlight` for the guard).

— Aristotle

src/components/Sidebar/__tests__/Sidebar.cascade.test.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+describe("Sidebar rename cascade survives tree re-flatten (#378)", () => {
+  beforeEach(() => {
+    vi.stubGlobal("localStorage", makeLocalStorage());
+    capturedFileChangeHandler = null;
+    vaultStore.reset();
+    bookmarksStore.reset();
+    vaultStore.setReady({ currentPath: VAULT, fileList: [], fileCount: 0 });
+    vi.clearAllMocks();
```
herox215, line 119: `vi.clearAllMocks()` resets the `updateLinksAfterRename` spy whose `.mockResolvedValue(...)` was set in the `vi.mock` factory at line 18 — after this line that mock resolves `undefined`. The current test does not exercise the confirm path so this is latent, but a follow-up that clicks Aktualisieren will get a spurious throw on `result.updatedPaths`. Re-apply the return value here in `beforeEach` alongside the other per-test mocks, or replace `clearAllMocks()` with `resetAllMocks()` which resets call history but keeps implementations.

— Aristotle

src/components/Sidebar/Sidebar.svelte
```diff
@@ -519,6 +541,150 @@
...
+  // #378 — cascade-request handlers from TreeRow. The row hands the request
+  // up the moment the IPC await resolves (synchronously, no further await),
+  // so the closure capture cannot race the row's unmount. Sidebar then runs
+  // any post-IPC work (getBacklinks for fileCount, dialog rendering,
+  // updateLinksAfterRename on confirm) on its own always-mounted lifetime.
+  async function handleRequestRenameCascade(req: RenameCascadeRequest) {
+    // Modal precedence: ignore overlapping cascade requests. Only one dialog
+    // open at a time. The user can re-trigger the suppressed action after
+    // resolving the first dialog.
+    if (pendingRename || pendingMove) return;
```
herox215: The re-entrancy guard is checked at line 553 but `pendingRename` is not set until line 561, after the `getBacklinks` await. A second rename triggering `handleRequestRenameCascade` during that await will pass the guard — both calls read `pendingRename === null` — and the final two writes race to set `pendingRename`, leaving a second dialog (or a silently overwritten first one). Fix: set a sentinel synchronously before the await, e.g. set `pendingRename = { ...req, fileCount: 1 }` (or a dedicated `inflightRename` boolean) before the `try`, then overwrite with the resolved `fileCount` after.

— Aristotle

src/components/Sidebar/Sidebar.svelte
```diff
@@ -519,6 +541,150 @@
...
+      fileCount = new Set(backlinks.map((b) => b.sourcePath)).size || 1;
+    } catch {
+      /* fall back to the linkCount/1 default */
+    }
+    pendingRename = { ...req, fileCount };
+  }
+
+  async function handleRequestMoveCascade(req: MoveDropRequest) {
+    // Modal precedence — same rule as rename cascade.
+    if (pendingRename || pendingMove) return;
```
herox215: Same re-entrancy window as `handleRequestRenameCascade`: the guard fires at line 566 but `pendingMove` is not written until line 590 (or the direct-move path at line 605), after the `getBacklinks` await at line 582. Two drops in rapid succession both read `pendingMove === null` and both proceed; the second `getBacklinks` call overlaps the first pipeline and can double-fire `moveFile`. Set a sentinel before the await.

— Aristotle

src/components/Sidebar/__tests__/Sidebar.cascade.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    expect(body.textContent ?? "").toMatch(/3\s+Links?\s+in\s+\d+\s+Dateien/);
+
+    // Sanity: the row that owned the original `pendingRename` is no longer
+    // mounted under the old path — the {#each} re-keyed it. Without the fix,
+    // this is exactly the moment the dialog would disappear.
+    const oldRow = container.querySelector(`[data-tree-row="${VAULT}/Welcome.md"]`);
+    expect(oldRow).toBeNull();
+    const newRow = container.querySelector(`[data-tree-row="${VAULT}/Welcome Renamed.md"]`);
+    expect(newRow).toBeTruthy();
+  });
```
herox215, line 209: The test stops after asserting the dialog is visible but never clicks "Aktualisieren" — `confirmRenameWithLinks`, `updateLinksAfterRename`, and the `finally { pendingRename = null }` path are all unexercised. The `finally` block was the core of issue #2 in the first review; it now has no regression coverage. Add a second `it` that clicks the confirm button, drains microtasks, and asserts: (a) `updateLinksAfterRename` was called once with the correct args, (b) the dialog is gone from the DOM.

— Aristotle

PR #382, opened 2026-04-30
feat(editor): char-morph transition between text-file tabs (#380)

### LINKED ISSUE #380: feat(editor): char-morph transition between text-file tabs
## Goal

When the user switches between markdown / plain-text tabs in the CodeMirror editor, the outgoing file's characters morph into the incoming file's characters via a canvas overlay scramble. Communicates content change beyond an instant swap, while staying inside VaultCore's minimal, dense aesthetic.

## Behavior

- Trigger: any tab activation that flips `activeTabId` between two tabs where `tabHasEditor(tab)` is true (click, `activateTab`, `cycleTab`, Cmd+1..9). Only when **both** outgoing and incoming `EditorView` already exist in `viewMap` — never during async `mountEditorView`.
- Effect: a canvas overlay sits over `.vc-editor-container` for the duration of the morph. The overlay paints the outgoing glyphs, then each glyph cycles through random characters in a **random per-glyph resolution order**, settling on the corresponding incoming glyph.
- Underneath the canvas, the incoming `EditorView` is already swapped in and fully interactive — typing during the morph works, focus is preserved.
- Mismatched lengths: pad shorter side with whitespace; surplus glyphs on the longer side fade to/from blank during the same window.

## Constraints (from Vitruvius design pass)

- Total time from tab-switch trigger to a fully interactive editor must remain within the < 100 ms open-note budget.
- Morph duration: **120 ms**, hard cut at the end. No tail.
- Suppression: if a second qualifying switch arrives within **200 ms** of the previous one, cancel the in-flight morph and swap instantly. Subsequent switches inside that window also stay instant.
- Editor must remain interactive during the morph: no `pointer-events: none`, `inert`, or `aria-disabled` on any ancestor of the CM6 view. Canvas overlay must use `pointer-events: none`.
- `prefers-reduced-motion: reduce` → instant swap, zero animation.
- Scope: applies only to tabs where `tabHasEditor(tab)` returns true. Graph, image, canvas, PDF, unsupported-preview, and Reading Mode (`ReadingView`) are excluded.
- Only between **already-loaded** tabs. No morph on first-open (empty pane → content) — that path stays the current instant render.
- Colors: only `--color-text` / `--color-text-muted`. No new color tokens.
- No new CSS animation chrome on tab bar, breadcrumb, status bar, sidebar.
- No changes to `tabStore`, `tabLifecycleStore`, `tabLayoutStore`, or backend IPC.
- Canvas is rendered at device pixel ratio. Font metrics read from the same `--vc-font-body` / line-height the editor uses, so glyphs align with the underlying CM6 layout when the morph starts and ends.
- Viewport-only: only the visible region of `.cm-scroller` is painted into the overlay. Off-screen content is not animated.

## Acceptance criteria

- Switching between two open markdown tabs triggers a 120 ms canvas-overlay scramble.
- Holding Cmd+Shift+] to cycle through 5+ tabs in under 1 s shows instant swaps after the first switch (200 ms suppression window).
- With `prefers-reduced-motion: reduce` set in OS, no animation plays.
- During the morph, the user can type and the keystrokes land in the incoming editor.
- Switching to a graph / image / canvas / PDF / unsupported / reading-mode tab shows no morph.
- Switching to a tab whose CM6 view is still mounting shows no morph (instant swap).
- No regression in the < 100 ms open-note budget; no regression in keystroke latency.

## Out of scope

- Tab bar / breadcrumb / status bar animation.
- Non-text tab types.
- First-open morph.
- Split-pane focus changes.
- Per-glyph DOM manipulation inside the CM6 content tree.

## Implementation notes

- New component: `TabMorphOverlay.svelte` mounted inside `.vc-editor-container` in `EditorPane.svelte`. Justified — overlay is a self-contained concern with its own canvas lifecycle.
- New CSS custom property: `--vc-tab-switch-duration` (default `120ms`), overridable via user snippets.
- The overlay reads the outgoing editor's visible text via the existing CM6 view (`view.visibleRanges` + `view.state.doc.sliceString`), takes a snapshot, then drives a single `requestAnimationFrame` loop for 120 ms. No DOM thrash inside the loop — only canvas draw calls.

herox215:
Closes #380.

## Summary
- New `TabMorphOverlay.svelte` mounted once per EditorPane inside `.vc-editor-content`. Snapshots both CM6 views' visible glyphs (via `view.coordsAtPos` + `view.state.doc.sliceString` over `view.visibleRanges`), drives a 120ms rAF canvas scramble, tears the canvas down at the end. Pointer-events: none + aria-hidden so the editor below stays fully interactive.
- Pure-logic split in `src/lib/editor/tabMorph.ts` — suppression-window decision, schedule build, reduced-motion gate, CSS-var duration parser — unit-tested without a DOM (17 tests).
- Trigger lives in EditorPane's existing `paneActiveTabId` effect. Gates: both sides must have a mounted CM6 view, both in edit mode, both `tabHasEditor` true, neither in `mountingIds`. Graph / image / canvas / unsupported / reading-mode all bypass.
- 200ms suppression window so Cmd+Shift+] chord cycles instant-swap after the first morph.
- `prefers-reduced-motion: reduce` → instant swap, zero animation.
- New CSS custom property `--vc-tab-switch-duration` (default 120ms). User snippets can set 0ms to opt out without flipping any JS gate.

## Test plan
- [x] 17 unit tests in `tabMorph.test.ts` (suppression rules, reduced-motion gate, schedule pairing, mismatched-length padding, duration parsing)
- [x] All 315 existing editor unit tests still pass
- [x] svelte-check clean for the Editor workspace (only pre-existing CanvasView error remains)
- [x] `pnpm build` clean
- [ ] UAT: open two markdown notes with different content, switch between them — expect a 120ms scramble. Cmd+Shift+] through 5 tabs in <1s — expect instant swaps. OS reduced-motion → no animation. Switch to a graph / image / PDF / canvas / reading-mode tab → no morph. Type during the morph → keystrokes land in the incoming editor.

## Files
- `src/lib/editor/tabMorph.ts` (new) — pure logic
- `src/lib/editor/__tests__/tabMorph.test.ts` (new) — 17 unit tests
- `src/components/Editor/TabMorphOverlay.svelte` (new) — canvas + rAF
- `src/components/Editor/EditorPane.svelte` — overlay mount + trigger gate
- `src/styles/tailwind.css` — `--vc-tab-switch-duration` token

### DIFF
--- a/src/components/Editor/EditorPane.svelte
+++ b/src/components/Editor/EditorPane.svelte
@@ -27,6 +27,7 @@
   import { parseFrontmatter } from "../../lib/frontmatterIO";
   import EditorGraphBackground from "./EditorGraphBackground.svelte";
   import CountStatusBar from "./CountStatusBar.svelte";
+  import TabMorphOverlay from "./TabMorphOverlay.svelte";
   import { countsStore } from "../../store/countsStore";
   import { computeCounts } from "../../lib/wordCount";
   import { scrollToMatch } from "./flashHighlight";
@@ -547,6 +548,12 @@
     }
   });
 
+  // #380: handle to the per-pane morph overlay. The overlay is mounted
+  // inside .vc-editor-content (so it sits above the editor containers
+  // but below the read-only scrim) and runs a 120ms canvas scramble on
+  // qualifying tab switches.
+  let morphOverlay: { play: (out: EditorView | null, inc: EditorView | null) => void } | null = $state(null);
+
   // Handle scroll save/restore on tab switch — separate effect to avoid
   // coupling with the lifecycle effect above
   $effect(() => {
@@ -563,6 +570,33 @@
               prevView.state.selection.main.head
             );
           } catch (_) { /* view may have been destroyed */ }
+        }
+      }
+      // #380: trigger morph if both sides are already-mounted text editors.
+      // Skipped during async mount (mountingIds), for non-editor tabs
+      // (graph/image/canvas/unsupported), and when either side is in
+      // Reading Mode (no CM6 layout to snapshot).
+      if (
+        prevActiveTabId &&
+        newActiveId &&
+        morphOverlay &&
+        !mountingIds.has(newActiveId)
+      ) {
+        const outTab = allTabs.find((t) => t.id === prevActiveTabId);
+        const inTab = allTabs.find((t) => t.id === newActiveId);
+        const outView = viewMap.get(prevActiveTabId);
+        const inView = viewMap.get(newActiveId);
+        const bothEdit =
+          (outTab?.viewMode ?? "edit") === "edit" &&
+          (inTab?.viewMode ?? "edit") === "edit";
+        if (
+          outView &&
+          inView &&
+          bothEdit &&
+          tabHasEditor(outTab) &&
+          tabHasEditor(inTab)
+        ) {
+          morphOverlay.play(outView, inView);
         }
       }
       // Restore scroll/cursor on activated tab
@@ -987,6 +1021,10 @@
     {/if}
     <!-- #87: ambient local-graph background behind the editor in edit mode -->
     <EditorGraphBackground visible={showGraphBg} relPath={bgRelPath} />
+    <!-- #380: 120ms char-morph overlay between text-file tabs. Sits above
+         the editor containers (z-index: 5) but below the readonly scrim
+         (z-index: 10). pointer-events: none so the editor stays interactive. -->
+    <TabMorphOverlay bind:this={morphOverlay} />
     <!-- Svelte renders one container per tab. Visibility is driven by
          style:display reacting to paneActiveTabId — no manual DOM needed.
          Graph tabs render <GraphView /> instead of a CM6 container so the
--- a/src/components/Editor/EditorPane.svelte
+++ b/src/components/Editor/EditorPane.svelte
@@ -578,8 +578,10 @@
   // #380: handle to the per-pane morph overlay. The overlay is mounted
   // inside .vc-editor-content (so it sits above the editor containers
   // but below the read-only scrim) and runs a 120ms canvas scramble on
-  // qualifying tab switches.
-  let morphOverlay: { playFromSnapshots: (out: import("../../lib/editor/tabMorph").ViewSnapshot | null, inc: import("../../lib/editor/tabMorph").ViewSnapshot | null) => void } | null = $state(null);
+  // qualifying tab switches. Type derived from the component so the
+  // signature here cannot drift from the export.
+  type MorphOverlayHandle = ReturnType<typeof TabMorphOverlay>;
+  let morphOverlay: MorphOverlayHandle | null = $state(null);
 
   // #380: outgoing snapshot must be taken BEFORE Svelte re-flips the
   // `style:display` of the previously-active container — once the old
--- a/src/components/Editor/EditorPane.svelte
+++ b/src/components/Editor/EditorPane.svelte
@@ -28,6 +28,7 @@
   import EditorGraphBackground from "./EditorGraphBackground.svelte";
   import CountStatusBar from "./CountStatusBar.svelte";
   import TabMorphOverlay from "./TabMorphOverlay.svelte";
+  import { snapshotView as snapshotEditorView } from "../../lib/editor/tabMorph";
   import { countsStore } from "../../store/countsStore";
   import { computeCounts } from "../../lib/wordCount";
   import { scrollToMatch } from "./flashHighlight";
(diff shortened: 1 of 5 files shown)


src/components/Editor/EditorPane.svelte
```diff
@@ -565,6 +572,33 @@
...
+          (outTab?.viewMode ?? "edit") === "edit" &&
+          (inTab?.viewMode ?? "edit") === "edit";
+        if (
+          outView &&
+          inView &&
+          bothEdit &&
+          tabHasEditor(outTab) &&
+          tabHasEditor(inTab)
+        ) {
+          morphOverlay.play(outView, inView);
```
herox215: `snapshotView(outgoingView)` is called here, but by the time this `$effect` fires, Svelte 5 has already committed the DOM mutation that switches `paneActiveTabId` — meaning the outgoing tab container is already `display:none`. CM6 measures layout from the DOM, so `view.visibleRanges` on a hidden element returns an empty range and the resulting `glyphs: []` snapshot is always empty. The morph never has any outgoing characters to scramble from. The snapshot of the outgoing view must be taken *before* the active-tab change is reflected in the DOM — either capture it eagerly in the event handler that triggers the tab switch, or snapshot it at the top of this effect before any store writes that derive `paneActiveTabId`.

— Aristotle

src/components/Editor/TabMorphOverlay.svelte
```diff
@@ -0,0 +1,199 @@
...
+  });
+</script>
+
+{#if visible && rect}
+  <canvas
+    bind:this={canvasEl}
+    class="vc-tab-morph-canvas"
+    aria-hidden="true"
+    style:left="{0}px"
+    style:top="{0}px"
```
herox215: `style:left="{0}px"` and `style:top="{0}px"` are dead. The CSS rule already sets `inset: 0` on `.vc-tab-morph-canvas`, which covers all four edges. These two inline style directives add noise and will silently override `inset` on browsers where `inset` expands to `top/right/bottom/left` (since inline style wins the cascade). Remove both attributes.

— Aristotle

src/components/Editor/TabMorphOverlay.svelte
```diff
@@ -0,0 +1,199 @@
...
+    randomGlyph,
+    resolveMorphDuration,
+    snapshotView,
+    type ScheduledGlyph,
+    type ViewSnapshot,
+  } from "../../lib/editor/tabMorph";
+
+  let canvasEl = $state<HTMLCanvasElement | null>(null);
+  let visible = $state(false);
+  let rect = $state<{ x: number; y: number; width: number; height: number } | null>(null);
```
herox215: `rect` is only ever read as a truthy guard in `{#if visible && rect}` — it carries no information the renderer uses (canvas sizing and drawing use `incoming.scrollerRect` directly). Replace the guard with just `visible`; if you want a null-safety gate inside `runMorph`, check `incoming` there. The `rect` state variable is dead weight and its `$state` declaration will trigger unnecessary reactive comparisons on every morph cycle.

— Aristotle

src/lib/editor/tabMorph.ts
```diff
@@ -0,0 +1,220 @@
...
+ * returns an empty snapshot if the view is detached.
+ */
+export function snapshotView(view: EditorView): ViewSnapshot | null {
+  const scrollerEl = view.scrollDOM as HTMLElement;
+  const contentEl = view.contentDOM as HTMLElement;
+  if (!scrollerEl || !contentEl || !scrollerEl.isConnected) return null;
+
+  const scrollerRect = scrollerEl.getBoundingClientRect();
+  const cs = window.getComputedStyle(contentEl);
+  const font = `${cs.fontStyle} ${cs.fontWeight} ${cs.fontSize} / ${cs.lineHeight} ${cs.fontFamily}`;
```
herox215: `cs.fontStyle` can be `"oblique 10deg"` (CSS `font-style: oblique <angle>`) for italic-like text. Concatenating it into a font shorthand produces `"oblique 10deg 700 14px / 20px ..."` — the angle token makes the shorthand unparseable to `ctx.font`, which silently falls back to the browser default (typically `"10px sans-serif"`). All canvas glyphs then render in the wrong font, misaligning with the CM6 grid. Use `ctx.font = \`\${cs.fontWeight} \${cs.fontSize} \${cs.fontFamily}\`` instead; `font-variant` and `font-stretch` default to `normal` which the canvas context already assumes, and dropping `line-height` from the shorthand removes the slash ambiguity entirely.

— Aristotle

src/lib/editor/__tests__/tabMorph.test.ts
```diff
@@ -0,0 +1,168 @@
...
+  it("returns true when matchMedia reports the reduce preference", () => {
+    const stub = vi.fn().mockReturnValue({ matches: true });
+    vi.stubGlobal("matchMedia", stub);
+    Object.defineProperty(window, "matchMedia", { configurable: true, value: stub });
+    expect(prefersReducedMotion()).toBe(true);
+    expect(stub).toHaveBeenCalledWith("(prefers-reduced-motion: reduce)");
+  });
+});
+
+describe("resolveMorphDuration", () => {
```
herox215, line 128: No `afterEach` resets `--vc-tab-switch-duration`. Each test sets the property but never cleans it up; the "falls back to MORPH_DURATION_MS" test only removes the property for its own invocation. If a test runner executes this describe block after any of the value-setting tests (e.g. under `--reporter=random` or when Vitest reorders within a pool), the "falls back" assertion passes for the wrong reason — it sees a previously-set value that happens to be non-parseable, or a prior value from another describe. Add a single `afterEach(() => document.documentElement.style.removeProperty("--vc-tab-switch-duration"))` at the top of this describe.

— Aristotle

src/components/Editor/EditorPane.svelte
```diff
@@ -547,6 +575,22 @@
     }
   });
 
+  // #380: handle to the per-pane morph overlay. The overlay is mounted
+  // inside .vc-editor-content (so it sits above the editor containers
+  // but below the read-only scrim) and runs a 120ms canvas scramble on
+  // qualifying tab switches.
+  let morphOverlay: { playFromSnapshots: (out: import("../../lib/editor/tabMorph").ViewSnapshot | null, inc: import("../../lib/editor/tabMorph").ViewSnapshot | null) => void } | null = $state(null);
```
herox215: `morphOverlay` is typed as an inline structural literal rather than the component's own exported type (`ComponentExports<typeof TabMorphOverlay>`). If `playFromSnapshots` gains or drops a parameter, this declaration silently stays stale and the mismatch only surfaces at the call site — one extra argument away from a runtime no-op. Use `import type { ComponentExports } from "svelte"; let morphOverlay: ComponentExports<typeof TabMorphOverlay> | null = $state(null);` to bind the type to the implementation.

— Aristotle

PR #384, opened 2026-04-30
feat(editor): extend char-morph tab transition to canvas tabs (#383)

### LINKED ISSUE #383: feat(editor): extend char-morph transition to canvas tabs (canvas↔text, canvas↔canvas)
## Context

#380 / #382 shipped a 120ms char-morph transition between text-file tabs (CM6 ↔ CM6) via `TabMorphOverlay.svelte` + `src/lib/editor/tabMorph.ts`. Canvas tabs are currently in the **bypass** list — switching to or from a canvas instant-swaps with no transition.

Goal: cover the remaining tab-switch combinations so the morph is consistent across the editor, not only between markdown files.

## Scope

Extend the morph trigger in `EditorPane.svelte` and the snapshot logic in `TabMorphOverlay.svelte` to handle:

1. **canvas → text** — snapshot the canvas surface (visible nodes + edges as glyphs, or a rasterized text-grid sample of the canvas viewport), morph into the incoming CM6 glyph grid.
2. **text → canvas** — snapshot the outgoing CM6 visible glyphs, morph into a glyph representation of the canvas viewport.
3. **canvas → canvas** — snapshot both canvas viewports as glyph grids, morph between them.

The morph stays purely visual. The underlying view swap is unchanged; the overlay just bridges the two frames.

## Constraints

- Reuse the existing `tabMorph.ts` logic (suppression window, reduced-motion gate, schedule builder, `--vc-tab-switch-duration`). Do **not** fork a second pipeline.
- Keystroke-latency budget (< 16ms) and idle-RAM budget (< 100MB) must not regress. Canvas snapshot must be bounded — sample a fixed glyph grid (e.g. width × height in monospace cells) instead of walking every node.
- Bypass list shrinks but does not vanish: graph view, image, PDF, reading-mode, unsupported types still instant-swap.
- `prefers-reduced-motion: reduce` → instant swap, same as text↔text.
- Pointer-events: none + aria-hidden on the overlay (no interaction regressions during the 120ms window).
- Cmd+Shift+] chord cycling must still hit the 200ms suppression window for canvas tabs (rapid cycle = instant after first morph).

## Open questions (answer before implementation)

- Canvas snapshot strategy: **(a)** rasterize the canvas DOM to a hidden offscreen canvas, then sample pixels into a glyph grid (density-mapped to ASCII shading), or **(b)** walk canvas-node text content and place glyphs at projected viewport coords. (a) is uniform but heavier; (b) is cheaper but ignores non-text nodes. Pick one and justify.
- For canvas → canvas, do we morph the *glyph grids* directly, or do we cross-fade the rasterized snapshots and let the morph engine treat them as character cells? Probably the former for consistency with text↔text.
- Empty canvas (no nodes): does the morph show a blank field scrambling into text, or do we bypass for empty canvases? Lean toward bypass — nothing to morph from.

## Acceptance

- [ ] canvas → text: 120ms scramble, no flash, no input lost.
- [ ] text → canvas: 120ms scramble, canvas interactive immediately after.
- [ ] canvas → canvas: 120ms scramble between two canvases with different node layouts.
- [ ] Bypass list (graph, image, PDF, reading-mode) still instant-swaps.
- [ ] Reduced-motion → instant.
- [ ] Cmd+Shift+] through mixed text + canvas tabs: first switch morphs, subsequent rapid switches instant.
- [ ] `pnpm build` clean, `svelte-check` clean for Editor workspace, unit tests for any new pure logic added to `tabMorph.ts`.
- [ ] No measurable regression on keystroke latency or idle RAM.

## Files (expected)

- `src/components/Editor/TabMorphOverlay.svelte` — canvas-snapshot path
- `src/components/Editor/EditorPane.svelte` — drop canvas from bypass, route through overlay
- `src/lib/editor/tabMorph.ts` (+ tests) — any new pure helpers (e.g. canvas-viewport → glyph-grid sampler signature)
- possibly `src/components/Canvas/*` — minimal hook to expose a viewport-snapshot method

## References

- #380 / PR #382 — text↔text char-morph
- `src/lib/editor/tabMorph.ts`, `src/components/Editor/TabMorphOverlay.svelte`

herox215:
Closes #383.

## Summary

- Extends the 240 ms char-morph from #380/#382 to cover canvas↔text and canvas↔canvas tab switches. Canvas tabs no longer instant-swap when they enter or leave the active slot.
- New `snapshotCanvas` walks text-bearing canvas nodes (text body, link URL, file basename, group label) and projects characters to viewport pixels via the camera transform — bounded at 1500 glyphs, off-screen nodes culled before they consume budget. No DOM rasterization, no async work, no regression on keystroke latency or idle RAM.
- `canvasMorphRegistry` keeps EditorPane and CanvasView decoupled — CanvasView registers a snapshot closure on mount, unregisters on destroy.
- `ViewSnapshot` / `GlyphRef` moved to a shared `src/lib/morphTypes.ts` to avoid `canvas/` importing across into `editor/`.

## Behavior

| Switch | Result |
|---|---|
| canvas → text | 240 ms scramble, no input lost |
| text → canvas | 240 ms scramble, canvas interactive immediately after |
| canvas → canvas | scramble between the two glyph grids |
| empty / unloaded canvas | bypass → instant (per ticket) |
| graph / image / unsupported / PDF | still bypass |
| `prefers-reduced-motion: reduce` | still instant |
| Cmd+Shift+] chord through mixed tabs | first morph plays, rest instant (existing suppression window unchanged) |

## Test plan

- [x] `pnpm vitest run src/lib/canvas/__tests__/canvasTabMorph.test.ts src/lib/canvas/__tests__/canvasMorphRegistry.test.ts src/lib/editor/__tests__/tabMorph.test.ts` — 36 passing
- [x] `pnpm svelte-check` — no new errors (one pre-existing #362 error on CanvasView.svelte:975 untouched)
- [x] `pnpm build` — clean
- [ ] **Manual UAT** (required before merge):
  - Open mixed text + canvas tabs
  - Switch text → canvas, canvas → text, canvas → canvas — expect 240 ms scramble
  - Cmd+Shift+] cycle through 4 mixed tabs — first morph plays, rest instant
  - `prefers-reduced-motion: reduce` — instant for all combinations
  - Open empty canvas, switch to it from text — instant (bypass)
  - Verify no input lost on incoming editor / canvas after morph

### DIFF
--- a/src/components/Canvas/CanvasView.svelte
+++ b/src/components/Canvas/CanvasView.svelte
@@ -69,6 +69,11 @@
     resolveVaultAbs,
     toVaultRel,
   } from "../../lib/canvas/embed";
+  import { snapshotCanvas } from "../../lib/canvas/canvasTabMorph";
+  import {
+    registerCanvasSnapshot,
+    unregisterCanvasSnapshot,
+  } from "../../lib/canvas/canvasMorphRegistry";
   import {
     type PointerMode,
     type DraftEdge,
@@ -203,6 +208,14 @@
 
   onMount(() => {
     void load();
+    // #383: expose a snapshot fn so EditorPane's tab-morph overlay can
+    // capture this canvas during canvas↔text and canvas↔canvas switches.
+    // Closure reads `viewportEl`, `doc`, `camX/Y`, `zoom` at snapshot
+    // time — they're plain `let` (DOM ref) and `$state` (camera + doc),
+    // both of which the closure observes at lookup, not at registration.
+    registerCanvasSnapshot(tabId, () =>
+      viewportEl ? snapshotCanvas(viewportEl, doc, camX, camY, zoom) : null,
+    );
   });
 
   onDestroy(() => {
@@ -212,6 +225,7 @@
       void persist();
     }
     cancelLongpressTimer();
+    unregisterCanvasSnapshot(tabId);
     // #165: tear down preview-invalidation subscriptions so multi-tab
     // open/close cycles don't leak listeners.
     unsubTab();
--- a/src/components/Canvas/CanvasView.svelte
+++ b/src/components/Canvas/CanvasView.svelte
@@ -225,6 +225,10 @@
       void persist();
     }
     cancelLongpressTimer();
+    // #383: unregister BEFORE the rest of the teardown so a tab-switch
+    // racing with this destroy can't read a snapshot fn whose closure
+    // points at `viewportEl` after Svelte has detached it. Ordering is
+    // load-bearing — a stale closure would dereference a null DOM ref.
     unregisterCanvasSnapshot(tabId);
     // #165: tear down preview-invalidation subscriptions so multi-tab
     // open/close cycles don't leak listeners.
--- a/src/components/Editor/EditorPane.svelte
+++ b/src/components/Editor/EditorPane.svelte
@@ -29,6 +29,7 @@
   import CountStatusBar from "./CountStatusBar.svelte";
   import TabMorphOverlay from "./TabMorphOverlay.svelte";
   import { snapshotView as snapshotEditorView } from "../../lib/editor/tabMorph";
+  import { snapshotCanvasTab } from "../../lib/canvas/canvasMorphRegistry";
   import { countsStore } from "../../store/countsStore";
   import { computeCounts } from "../../lib/wordCount";
   import { scrollToMatch } from "./flashHighlight";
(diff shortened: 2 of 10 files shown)


src/lib/morphTypes.ts
```diff
@@ -0,0 +1,28 @@
+// Shared types for the tab-morph effect (#380, #383).
```
herox215, line 1: `TabMorphOverlay` imports `ViewSnapshot` via `../../lib/editor/tabMorph` (its re-export), while `EditorPane` line 619 imports it inline via `../../lib/morphTypes`. Two consumers, two import paths for the same type — if the re-export in `tabMorph.ts` is ever dropped, `TabMorphOverlay` silently breaks. Pick one canonical path (`morphTypes`) and have all consumers use it.

— Aristotle

src/components/Editor/EditorPane.svelte
```diff
@@ -632,20 +659,23 @@
       ) {
         const outTab = allTabs.find((t) => t.id === prevActiveTabId);
         const inTab = allTabs.find((t) => t.id === newActiveId);
-        const inView = viewMap.get(newActiveId);
         const bothEdit =
```
herox215, line 665: `bothEdit` evaluates `(inTab?.viewMode ?? "edit") === "edit"` for canvas tabs, which is always `true` since canvas tabs have no `viewMode` field. This means the check is load-bearing only for text tabs — it never actually gates a canvas↔anything switch. The PR comment acknowledges this but treating a missing field as "edit" is an implicit contract with the Tab type. Assert it explicitly: add `tab.viewer === "canvas"` as a short-circuit in the `bothEdit` evaluation, or add a type-narrowing comment here explaining why the fallback is safe and intentional so the next reader does not silently tighten this.

— Aristotle

src/lib/canvas/canvasTabMorph.ts
```diff
@@ -0,0 +1,171 @@
...
+    case "file": {
+      const file = (node as CanvasFileNode).file;
+      if (!file) return null;
+      const slash = file.lastIndexOf("/");
+      return slash >= 0 ? file.slice(slash + 1) : file;
+    }
+    case "group":
+      return (node as CanvasGroupNode).label || null;
+    default:
+      return null;
```
herox215, line 57: The monospace cell-width approximation (`fontSize * 0.6`) is applied to all canvas node text, including proportional fonts. For group labels or link URLs rendered in a proportional face, glyph positions can be off by a significant fraction of a character width — visible as clustering at wrong positions during the 120ms scramble. The file header calls this acceptable but does not tie it to a ticket. Add a `// TODO(#383): proportional-font metric pass` here so the known inaccuracy is traceable and does not become invisible technical debt.

— Aristotle

src/components/Canvas/CanvasView.svelte
```diff
@@ -212,6 +225,7 @@
       void persist();
     }
     cancelLongpressTimer();
+    unregisterCanvasSnapshot(tabId);
```
herox215, line 232: `unregisterCanvasSnapshot(tabId)` fires before `void persist()` in the destroy sequence, so a snapshot captured immediately after `onDestroy` starts (e.g. a tab switch that races the close gesture) will correctly return `null` from the registry. But `persist()` is fire-and-forget — if the canvas is still in the registry during the async persist write, a concurrent snapshot call could still read live-but-partially-torn state. This is fine in practice because the morph overlay does not run during a close gesture. Document the assumption explicitly so the ordering is not accidentally changed: the unregister must come before any async work.

— Aristotle

src/components/Editor/EditorPane.svelte
```diff
@@ -259,14 +260,38 @@
     return true;
   }
 
+  // #383: tabs that participate in the char-morph tab transition. Superset
+  // of `tabHasEditor` — adds canvas tabs, which are snapshotted via the
+  // canvasMorphRegistry rather than from a CM6 view. Graph / image /
+  // unsupported / PDF still bypass to instant-swap.
+  function tabSupportsMorph(t: Tab | undefined): boolean {
```
herox215, line 268: `tabSupportsMorph` duplicates the exclusion logic of `tabHasEditor` (lines 255–261) and adds only a canvas pass-through — but without calling `tabHasEditor` internally. If a new viewer type is added to `TabViewer` (e.g. `"pdf"`) and someone correctly adds an exclusion to `tabHasEditor`, they must remember to mirror it in `tabSupportsMorph`. The two functions will silently diverge. Either delegate: `tabSupportsMorph(t) = tabHasEditor(t) || t?.viewer === "canvas"`, or consolidate them into one function with a parameter.

— Aristotle

Issue #386, opened 2026-04-30
feat(mobile): #74.2 mobile shell — single-pane layout + drawer + sheet


herox215:
Sub-ticket of #74. Depends on #74.1 (viewport store).

## Goal

Below 700px viewport, collapse `VaultLayout` to a single editor pane with the sidebar as a left drawer and the right-sidebar as a full-screen sheet. Desktop layout unchanged above 700px.

## Approach

The existing 5-column grid stays. Apply a media-query overlay:

```css
@media (max-width: 699px) {
  .vc-vault-layout { grid-template-columns: 1fr; }
  .vc-layout-sidebar {
    position: fixed; inset: 0 auto 0 0;
    width: 280px;
    transform: translateX(-100%);
    transition: transform 240ms ease;
    z-index: 50;
  }
  .vc-layout-sidebar--mobile-open { transform: translateX(0); }
  .vc-layout-divider, .vc-layout-divider-right,
  .vc-layout-divider-hidden, .vc-layout-divider-right-hidden,
  .vc-split-divider { display: none; }
  .vc-layout-right-sidebar {
    position: fixed; inset: 0;
    transform: translateX(100%);
    transition: transform 240ms ease;
  }
  .vc-layout-right-sidebar--mobile-open { transform: translateX(0); }
}
```

Drag-to-resize handlers stay attached unchanged — dividers are `display: none` on mobile so `mousedown` cannot fire. **No `if (isMobile)` branches in event handlers.**

## State

- New flag `mobileDrawerOpen` (separate from desktop `sidebarCollapsed`) — different intent: desktop "collapsed" = no sidebar; mobile "closed" = drawer hidden, can slide back in.
- Topbar collapse-toggle on mobile becomes "open drawer".
- Right-sidebar collapse-toggle on mobile becomes "open sheet".
- Split view (`isSplit`) unconditionally inert on mobile.

## Gestures

- **`src/lib/actions/swipeGesture.ts`** (new) — Svelte action via Pointer Events. Edge-only mode for left-edge sidebar reveal (matches iOS UIKit). Threshold: 50px horizontal in 300ms, vertical drift <30px.
- Scrim tap → close drawer / sheet.
- Swipe-right on drawer → close. Swipe-left from drawer-right-edge → close. Swipe-left from screen-right-edge → open right sheet.

## Vitruvius input required

- Drawer scrim color (light/dark)
- `env(safe-area-inset-*)` integration for iOS notch/island
[...]

Issue #385, opened 2026-04-30
feat(mobile): #74.1 viewport-mode store + hit-target token sweep


herox215:
Sub-ticket of #74. Foundation PR for mobile support — desktop must remain byte-identical.

## Goal

Land the primitives the rest of the mobile work depends on, without changing any visible desktop behavior.

## Scope

- **`src/store/viewportStore.ts`** (new) — Svelte readable store exposing `{ mode: "desktop" | "tablet" | "mobile", isCoarsePointer: boolean, width: number }`. Driven by `matchMedia("(max-width: 699px)")`, `matchMedia("(max-width: 1023px)")`, `matchMedia("(pointer: coarse)")`. Hit-target bumps gate on `isCoarsePointer`; layout collapse gates on `mode`.
- **`--vc-hit-target` CSS custom property** — defined at `:root` as `auto`, overridden to `44px` under `@media (pointer: coarse)`. Each touched control references it via `min-height: var(--vc-hit-target, <desktop-value>)` so the desktop dimension stays as the literal fallback.
- **Hit-target sweep** — bump these to use the token (desktop value preserved as fallback):
  - `vc-tree-row` (24px → 44px on coarse)
  - `vc-tabbar` / Tab (36 → 44)
  - `vc-sidebar-toggle-btn` (32 → 44)
  - `vc-tree-more` (always show on `(hover: none)`, currently hover-only)
  - `vc-context-item`, `vc-confirm-btn`, `vc-right-tab` (36 → 44)
  - Breadcrumbs (28 → 44), settings inputs (28–36 → 44)
  - PropertiesPanel rows (24 → 44), TagRow (28 → 44)
  - FileListRow / RecentVaultRow (32 → 44)
- **`@media (hover: hover)` guards** around `:hover` selectors that visibly stick on touch (cosmetic — accept defer if too invasive, address in #74.2).

## Boy Scout

- `src/test/setup.ts` currently has no `matchMedia` mock; multiple tests stub it ad-hoc. Add a global controllable stub.

## Tests (TDD)

- `viewportStore.test.ts` — mock `matchMedia`, assert mode transitions on width change, `isCoarsePointer` independent of width.
- Snapshot/visual regression: run a representative desktop scenario and assert no class/style diff vs main.

## Out of scope

- Layout topology changes (#74.2)
- Long-press (#74.3)
- Any Tauri config changes — pure CSS/JS PR

herox215:
[...]

Issue #74, opened 2026-04-15
Mobile / tablet support


herox215:
## Summary

Ship VaultCore on iOS and Android via **Tauri 2 Mobile**. Scope is a **"Reader + Quick-Capture" mobile experience** — full read/navigate/search, editable but not feature-parity with desktop. Existing Rust backend (Tantivy, `notify`, FS-layer, all commands) transfers as-is. Frontend gets a responsive shell, touch gestures, and mobile-specific panels via conditional rendering driven by a viewport store.

## Framework Decision

**Tauri 2 Mobile.** Confirmed: only path that preserves the Rust backend (Tantivy/notify/Rust commands) on mobile, keeps performance constraints achievable, and avoids a second codebase. Capacitor / native Swift+Kotlin / React-Native rewrites all rejected.

## Use-Case Profile

**"Reader + Quick-Capture"** — read notes on the go, small edits, capture quickly. Long-form editing stays on desktop. Mobile is not desktop-in-miniature.

- Read-mode is the default on note open (Obsidian-style). Pencil toggle in topbar switches to edit-mode. Edit-mode persists per tab, not globally.
- Quick-capture has **no FAB**. New notes are created via long-press on a folder (or vault root) in the tree → context menu → "New Note here". The long-pressed folder is the target.

## Mobile Navigation Layout (decided 2026-04-30)

Hybrid (Variante C):
- **Topbar**: existing 32px sidebar-toggle button = open File-Tree drawer (left).
- **Bottom-tab-bar**: 3 tabs — Files / Search / More.
- **Burger sheet** (via the More tab): hosts Backlinks, Bookmarks, Outline, Outgoing Links, Properties, Settings.
- **Right-sidebar full-screen sheet on mobile**: dropped (its content moved to the burger sheet).

## Scope (In)

**Phase A — Read & Navigate**
- Vault open via platform Document Picker (iOS Files-App, Android SAF)
- File-tree as left drawer (toggle via topbar button)
- Note open + read (read-mode default)
- Internal-link follow
- Tabs (sequential, no split, horizontally scrollable)
- Theme (light/dark/auto) with `env(safe-area-inset-*)`

**Phase B — Edit**
- CodeMirror 6 with default touch selection
- Pencil-icon toggle in topbar → edit-mode (per-tab persistence)
- Wiki-link autocomplete
- Auto-save (same semantics as desktop)
- Tags panel read-only (Phase D adds write via frontmatter editing)

**Phase C — Search**
[...]

PR #396, opened 2026-04-30
feat(mobile): viewport store + hit-target token sweep (#385)

herox215:
Foundation PR for mobile/tablet support — desktop must remain byte-identical. Refs #385, #74.

## Summary

- **New `src/store/viewportStore.ts`** — Svelte readable store `{ mode, isCoarsePointer, width }` driven by three `matchMedia` listeners: `(max-width: 699px)`, `(max-width: 1023px)`, `(pointer: coarse)`. Layout collapse gates on `mode`; hit-target bumps gate on `isCoarsePointer` (kept independent so a desktop with a touchscreen still gets 44px targets without collapsing the layout). Pattern mirrors `themeStore.ts` (classic writable factory, D-06/RC-01).
- **New `--vc-hit-target` CSS token** — intentionally **left undefined on desktop** so `min-height: var(--vc-hit-target, <px>)` falls back to the literal desktop value at every callsite. Under `@media (pointer: coarse)`, `:root` sets it to `44px` and the fallback is overridden.
- **Hit-target sweep** — TreeRow, Tab/TabBar, sidebar toggle, tree-kebab (also forced visible on `@media (hover: none)`), context-menu items, confirm buttons, right-sidebar tabs, breadcrumbs + mode toggle, settings controls (close, theme radio, vault switch, generic btn), properties panel rows + add button + chips, tag row, file-list row, recent-vault row. Each control adds `min-height: var(--vc-hit-target, <existing-px>)` (and `min-width` on icon-square buttons) **alongside** the existing `height` declaration. On desktop the var is undefined → fallback engages → exact same value as the existing `height` → byte-identical. On coarse, `min-height` becomes 44px and overrides `height` per CSS spec.
- **Boy Scout** — added a global controllable `matchMedia` stub in `src/test/setup.ts` (jsdom does not implement it). Existing per-suite `vi.stubGlobal('matchMedia', …)` overrides keep working.

## Why undefined-token-with-fallback

An earlier draft set `--vc-hit-target: auto` on `:root` and relied on `auto` being invalid for `min-height` to engage the `var()` fallback. That was wrong — `min-height: auto` is the property's initial value (valid). Leaving the custom property *undefined* on desktop is the only approach that guarantees the fallback engages.

## Out of scope (deferred per ticket)

- `@media (hover: hover)` guards on `:hover` selectors (cosmetic — ticket explicitly allows defer; will land in #386).
- Topbar / layout adjustments needed to host the larger toggle button on touch — belongs to #386.
- The `vc-context-item` flex-vs-block decision: kept `display: block` to avoid even cosmetic desktop change; on coarse the menu text sits at the top of a 44px row. Polish in #386.

## Test plan

- [x] `pnpm vitest run src/store/__tests__/viewportStore.test.ts` — 7/7 pass (mode transitions, isCoarsePointer independence, multiple subscribers, SSR fallback).
- [x] `pnpm vitest run` for every test that touches the swept components (Sidebar, Tabs, ContextMenu, Editor, Tags, modal theming) — 124/124 pass.
- [x] `pnpm build` — succeeds.
- [x] `svelte-check` — no new errors (one pre-existing CanvasView error on `main` is unaffected).
- [ ] **Visual desktop verification via `pnpm dev` was NOT run from this environment** — the change is mathematically byte-identical on desktop (token undefined → fallback engages literal px), but a manual visual pass before merge is recommended. Asking for it as part of UAT.
- [ ] Manual coarse-pointer verification (Chromium DevTools: emulate "Touch") — recommended in UAT.

Refs #385, #74.

### DIFF
--- a/src/components/Editor/Breadcrumbs.svelte
+++ b/src/components/Editor/Breadcrumbs.svelte
@@ -138,6 +138,8 @@
     align-items: center;
     gap: 6px;
     height: 28px;
+    /* #385 — bumps to 44px on coarse pointer (overrides `height`). */
+    min-height: var(--vc-hit-target, 28px);
     padding: 0 12px;
     background: var(--color-surface);
     border-bottom: 1px solid var(--color-border);
@@ -172,6 +174,9 @@
     justify-content: center;
     width: 24px;
     height: 22px;
+    /* #385 — square 44px touch target on coarse. */
+    min-width: var(--vc-hit-target, 24px);
+    min-height: var(--vc-hit-target, 22px);
     padding: 0;
     flex-shrink: 0;
     background: transparent;
--- a/src/components/Layout/RightSidebar.svelte
+++ b/src/components/Layout/RightSidebar.svelte
@@ -124,6 +124,8 @@
     align-items: center;
     justify-content: center;
     height: 36px;
+    /* #385 — coarse-pointer min-height overrides height per CSS spec. */
+    min-height: var(--vc-hit-target, 36px);
     background: transparent;
     border: none;
     cursor: pointer;
--- a/src/components/Layout/VaultLayout.svelte
+++ b/src/components/Layout/VaultLayout.svelte
@@ -1064,6 +1064,10 @@
     justify-content: center;
     width: 32px;
     height: 32px;
+    /* #385 — bump to 44x44 on coarse pointer; both axes since the icon
+       button needs a square touch target. */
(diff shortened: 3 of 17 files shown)


herox215:
Closing — Socrates returned REVISE on the plan (5 critical findings: MQL listener leak, stale `width` field, TDD order violation, sweep gaps, plan/code description mismatch). Reworking on a fresh branch with tests-committed-first. New PR will reference the same issues.

PR #398, opened 2026-04-30
feat(mobile): viewport store + hit-target token sweep (#385)

herox215:
Foundation PR for mobile/tablet support. Desktop remains byte-identical. Refs #385, #74.

## Summary

- **New `src/store/viewportStore.ts`** — Svelte readable store `{ mode, isCoarsePointer }` driven by three `matchMedia` listeners: `(max-width: 699px)`, `(max-width: 1023px)`, `(pointer: coarse)`. Layout collapse gates on `mode`; hit-target bumps gate on `isCoarsePointer` (independent so a desktop with a touchscreen still gets 44px targets without collapsing the layout). Implemented via Svelte's `readable(initial, start)` so listener teardown is driven by subscriber count — `start` returns a `stop` callback that removes every MQL listener it added. `matchMedia` calls are guarded by try/catch for sandboxed WebView configs that expose the function but throw on invocation.
- **New `--vc-hit-target` CSS custom property** — intentionally **left undefined on `:root`** so `min-height: var(--vc-hit-target, <px>)` falls back to the literal desktop value at every callsite (byte-identical desktop). Defined as `44px` only under `@media (pointer: coarse)`, where `min-height` overrides any sibling `height` per CSS spec.
- **Boy Scout** — global controllable `matchMedia` stub in `src/test/setup.ts` (jsdom doesn't implement it). Existing per-suite `vi.stubGlobal('matchMedia', …)` overrides keep working.

## Why undefined-token-with-fallback

An earlier draft (closed PR #396) set `--vc-hit-target: auto` on `:root` and relied on `auto` being invalid for `min-height` to engage the `var()` fallback. That was wrong — `min-height: auto` is a valid value (the property's initial value). Leaving the custom property *undefined* on desktop is the only approach that guarantees the per-callsite fallback engages.

## Components touched in the sweep (17 components, 22 controls)

| File | Selectors |
|---|---|
| `src/components/Sidebar/TreeRow.svelte` | `.vc-tree-row` (24), `.vc-tree-more` (forced visible on `(hover: none)` — scoped here), `.vc-confirm-btn` (auto fallback) |
| `src/components/Sidebar/Sidebar.svelte` | `.vc-sidebar-action-btn` (32×32), `.vc-new-menu-item` (auto fallback), `.vc-confirm-btn` (auto fallback) |
| `src/components/Tabs/TabBar.svelte` | `.vc-tabbar` (36) |
| `src/components/Tabs/Tab.svelte` | `.vc-tab` (36) |
| `src/components/Layout/VaultLayout.svelte` | `.vc-sidebar-toggle-btn` (32×32) |
| `src/components/Layout/RightSidebar.svelte` | `.vc-right-tab` (36) |
| `src/components/common/ContextMenu.svelte` | `:global(.vc-context-item)` (auto fallback) |
| `src/components/common/PasswordPromptModal.svelte` | `.vc-password-modal-input` (32), `.vc-password-modal-cancel`, `.vc-password-modal-ok` (auto fallback) |
| `src/components/Editor/Breadcrumbs.svelte` | `.vc-breadcrumbs` (28), `.vc-breadcrumbs-mode-toggle` (24×22) |
| `src/components/Settings/SettingsModal.svelte` | `.vc-settings-close` (28×28), `.vc-theme-radio` (36), `.vc-vault-switch-btn` (32), `.vc-settings-btn` (32) |
| `src/components/Properties/PropertiesPanel.svelte` | `.vc-props-add` (24×24), `.vc-props-key` / `.vc-props-value` (24), `.vc-props-chips` (24) |
| `src/components/Tags/TagRow.svelte` | `.vc-tag-row` (28) |
| `src/components/Welcome/FileListRow.svelte` | `.vc-file-row` (32) |
| `src/components/Welcome/RecentVaultRow.svelte` | `.vc-recent-row` (32) |
| `src/components/Search/QuickSwitcherRow.svelte` | `.vc-qs-row` (auto fallback) |
| `src/components/Search/OmniSearch.svelte` | `.vc-omni-mode` (auto fallback — Filename/Content mode toggles) |
| `src/components/CommandPalette/CommandPalette.svelte` | `.vc-cp-row` (auto fallback) |

Each callsite carries a `/* #385 — ... */` comment explaining the undefined-fallback strategy so future maintainers don't "fix" it to `0` and break the desktop guard.

## TDD discipline (visible in `git log`)

Two RED → GREEN cycles, each test commit landing the failing spec before the corresponding impl commit:

```
31b67f0 feat(viewport): guard matchMedia throw + sweep sidebar action buttons (#385)   ← GREEN cycle 2
12c4ae4 test(viewport): cover throwing matchMedia + value-after-reattach (#385)       ← RED   cycle 2
da8f9c5 feat(mobile): apply --vc-hit-target token across touch-affected controls (#385)
628ecb9 feat(viewport): implement viewportStore via readable(initial, start) (#385)    ← GREEN cycle 1
51e77ae test(viewport): add viewportStore TDD spec + matchMedia setup stub (#385)     ← RED   cycle 1
```

Verified locally — `viewportStore.ts` does not exist at `51e77ae` (test fails to import), and the throwing-matchMedia test fails at `12c4ae4` (no try/catch yet in the impl).

### Automated (already run, all green)

- [x] `pnpm vitest run src/store/__tests__/viewportStore.test.ts` — 10/10 pass: mode transitions, isCoarsePointer independence, multi-subscriber updates, listener cleanup on last unsubscribe, listener re-attach with current MQL state (not stale closure), SSR fallback, throwing-matchMedia fallback.
- [x] `pnpm vitest run` for every test that touches the swept components (Sidebar, Tabs, ContextMenu, Editor, Tags, OmniSearch, CommandPalette, modal theming, tabMorph) — 148/148 pass.
- [x] `pnpm build` — clean.

### Manual UAT (NOT run in implementing agent's environment — please verify before merge)

**Desktop** (default, no coarse pointer): open vault and confirm no visual regression on
- [ ] tree rows (24px row height preserved)
- [ ] tabs / tab bar (36px height preserved)
- [ ] sidebar header buttons (New Note / New Folder / Graph / Docs — 32×32 preserved)
- [ ] sidebar collapse toggles (32×32 preserved)
- [ ] breadcrumbs (28px height) and mode toggle (24×22)
- [ ] settings modal: close button (28×28), theme radios (36), vault switch & generic buttons (32)
- [ ] properties panel: rows (24), add button (24×24), chips
- [ ] password prompt modal (input 32, buttons padding-only)
- [ ] command palette + quick switcher rows (padding-only)

**Coarse pointer** (touchscreen device or Chromium DevTools → Device Toolbar → "Touch"):
- [ ] every gated control feels ≥44px on tap
- [ ] tree-row kebab is visible WITHOUT hover (`(hover: none)` rule — no Vitest coverage, UAT-only)
- [ ] sidebar header action buttons grow to 44×44

## Known caveats / out of scope

- **`vc-sidebar-toggle-btn` and `vc-sidebar-action-btn` overflow the 36px topbar on coarse pointer** when they grow to 44×44. Out of scope per ticket — #386 (Mobile shell) reworks the topbar layout.
- **`@media (hover: none)` for the tree kebab has no Vitest coverage** — jsdom doesn't reliably evaluate `hover` queries. UAT-only verification.
- **`@media (hover: hover)` guards on `:hover` selectors** — cosmetic; ticket explicitly allows defer; will land in #386.
- **`vc-context-item` flex-vs-block** — kept `display: block` to avoid even cosmetic desktop change; on coarse the menu text sits at the top of a 44px row. Polish in #386.
- **`ReindexStatusbar`** — referenced in code comments but no `.svelte` file exists in the tree. If it lands later it will need its own follow-up. `EncryptionStatusbar` is purely a `<div role="status">` notification (no tappable interactives), so no sweep needed.

Refs #385, #74.

### DIFF
--- a/src/components/CommandPalette/CommandPalette.svelte
+++ b/src/components/CommandPalette/CommandPalette.svelte
@@ -237,6 +237,9 @@
     color: var(--color-text);
     font-size: 14px;
     text-align: left;
+    /* #385 — token undefined on desktop → fallback `auto` (byte-identical
+       to current padding-only height); coarse → 44px. */
+    min-height: var(--vc-hit-target, auto);
   }
   .vc-cp-row--selected { background: var(--color-accent-bg); color: var(--color-accent); }
   .vc-cp-row-name { flex: 1; min-width: 0; overflow: hidden; text-overflow: ellipsis; white-space: nowrap; }
--- a/src/components/CommandPalette/CommandPalette.svelte
+++ b/src/components/CommandPalette/CommandPalette.svelte
@@ -237,8 +237,8 @@
     color: var(--color-text);
     font-size: 14px;
     text-align: left;
-    /* #385 — token undefined on desktop → fallback `auto` (byte-identical
-       to current padding-only height); coarse → 44px. */
+    /* #385 — fallback `auto` = no min-height constraint (initial value), so
+       desktop sizing stays padding-driven as before; coarse → 44px. */
     min-height: var(--vc-hit-target, auto);
   }
   .vc-cp-row--selected { background: var(--color-accent-bg); color: var(--color-accent); }
--- a/src/components/Sidebar/Sidebar.svelte
+++ b/src/components/Sidebar/Sidebar.svelte
@@ -1038,8 +1038,8 @@
     cursor: pointer;
     background: var(--color-surface);
     color: var(--color-text);
-    /* #385 — token undefined on desktop → fallback `auto` (byte-identical
-       padding-only height); coarse → 44px. */
+    /* #385 — fallback `auto` = no min-height constraint (initial value), so
+       desktop sizing stays padding-driven as before; coarse → 44px. */
     min-height: var(--vc-hit-target, auto);
   }
 
@@ -1196,8 +1196,8 @@
     border: none;
     cursor: pointer;
     color: var(--color-text);
-    /* #385 — token undefined on desktop → fallback `auto` (byte-identical
-       to current padding-only height); coarse → 44px. */
+    /* #385 — fallback `auto` = no min-height constraint (initial value), so
+       desktop sizing stays padding-driven as before; coarse → 44px. */
     min-height: var(--vc-hit-target, auto);
   }
 
--- a/src/components/Sidebar/Sidebar.svelte
+++ b/src/components/Sidebar/Sidebar.svelte
@@ -1124,6 +1124,9 @@
     justify-content: center;
     width: 32px;
     height: 32px;
+    /* #385 — fallbacks equal width/height (byte-identical); coarse → 44×44. */
+    min-width: var(--vc-hit-target, 32px);
+    min-height: var(--vc-hit-target, 32px);
     background: none;
     border: none;
     border-radius: 4px;
@@ -1193,6 +1196,9 @@
     border: none;
     cursor: pointer;
     color: var(--color-text);
+    /* #385 — token undefined on desktop → fallback `auto` (byte-identical
+       to current padding-only height); coarse → 44px. */
+    min-height: var(--vc-hit-target, auto);
   }
 
   .vc-new-menu-item:hover {
--- a/src/components/Sidebar/Sidebar.svelte
+++ b/src/components/Sidebar/Sidebar.svelte
@@ -1038,6 +1038,9 @@
     cursor: pointer;
     background: var(--color-surface);
     color: var(--color-text);
+    /* #385 — token undefined on desktop → fallback `auto` (byte-identical
+       padding-only height); coarse → 44px. */
+    min-height: var(--vc-hit-target, auto);
   }
 
   .vc-confirm-btn:hover {
--- a/src/store/__tests__/viewportStore.test.ts
+++ b/src/store/__tests__/viewportStore.test.ts
@@ -1,0 +1,217 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { get } from "svelte/store";
+
+import { createViewportStore } from "../viewportStore";
+
+type MqlListener = (ev: MediaQueryListEvent) => void;
+
+interface FakeMql {
+  media: string;
+  matches: boolean;
+  listeners: Set<MqlListener>;
+  addCount: number;
+  removeCount: number;
+  addEventListener: (type: "change", l: MqlListener) => void;
+  removeEventListener: (type: "change", l: MqlListener) => void;
+  // legacy (unused but populated for compat)
+  addListener: (l: MqlListener) => void;
+  removeListener: (l: MqlListener) => void;
+  onchange: null;
+  dispatchEvent: () => boolean;
+}
+
+interface MqlHarness {
+  matchMedia: (q: string) => FakeMql;
+  mqls: Map<string, FakeMql>;
+  /** Flip `matches` and dispatch `change` to all attached listeners. */
+  set: (query: string, matches: boolean) => void;
+  get: (query: string) => FakeMql;
+}
+
+function makeMqlHarness(initial: Record<string, boolean>): MqlHarness {
+  const mqls = new Map<string, FakeMql>();
+
+  function getOrCreate(query: string): FakeMql {
+    const existing = mqls.get(query);
+    if (existing) return existing;
+    const mql: FakeMql = {
+      media: query,
+      matches: initial[query] ?? false,
+      listeners: new Set(),
+      addCount: 0,
+      removeCount: 0,
+      addEventListener: (type, l) => {
+        if (type !== "change") return;
+        mql.addCount++;
+        mql.listeners.add(l);
+      },
+      removeEventListener: (type, l) => {
+        if (type !== "change") return;
+        mql.removeCount++;
+        mql.listeners.delete(l);
+      },
+      addListener: (l) => mql.listeners.add(l),
+      removeListener: (l) => mql.listeners.delete(l),
+      onchange: null,
+      dispatchEvent: () => false,
+    };
+    mqls.set(query, mql);
+    return mql;
+  }
+
+  return {
+    matchMedia: getOrCreate,
+    mqls,
+    get: getOrCreate,
+    set(query, matches) {
+      const mql = getOrCreate(query);
+      mql.matches = matches;
+      const ev = { matches, media: query } as MediaQueryListEvent;
+      for (const l of mql.listeners) l(ev);
+    },
+  };
+}
+
+const Q_MOBILE = "(max-width: 699px)";
+const Q_TABLET = "(max-width: 1023px)";
+const Q_COARSE = "(pointer: coarse)";
+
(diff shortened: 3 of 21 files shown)


src/store/viewportStore.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface ViewportState {
+  mode: ViewportMode;
+  isCoarsePointer: boolean;
+}
+
+const Q_MOBILE = "(max-width: 699px)";
+const Q_TABLET = "(max-width: 1023px)";
+const Q_COARSE = "(pointer: coarse)";
+
+const SSR_DEFAULT: ViewportState = {
```
herox215: `SSR_DEFAULT` is a mutable module-level object passed by reference into `readable`. Every `createViewportStore()` call shares the same initial-value pointer — if any consumer mutates the returned state (possible with some Svelte store patterns), it contaminates the initial value for all future factory calls. Either `Object.freeze(SSR_DEFAULT)` here, or spread it at the `readable` callsite: `readable<ViewportState>({ ...SSR_DEFAULT }, ...)`.

— Aristotle

src/store/viewportStore.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const read = () => {
+      set({
+        mode: modeFor(mqlMobile.matches, mqlTablet.matches),
+        isCoarsePointer: mqlCoarse.matches,
+      });
+    };
+
+    read();
+    const onChange = () => read();
```
herox215: `const onChange = () => read()` is a needless wrapper — `read` already takes no arguments and has exactly this shape. Use `mqlMobile.addEventListener("change", read)` directly. The wrapper prevents the three `removeEventListener` calls from matching the registered reference if ever refactored to pass `onChange` vs `read` inconsistently.

— Aristotle

src/store/__tests__/viewportStore.test.ts
```diff
@@ -0,0 +1,252 @@
...
+}
+
+const Q_MOBILE = "(max-width: 699px)";
+const Q_TABLET = "(max-width: 1023px)";
+const Q_COARSE = "(pointer: coarse)";
+
+function installHarness(initial: Record<string, boolean>): MqlHarness {
+  const harness = makeMqlHarness(initial);
+  vi.stubGlobal("matchMedia", harness.matchMedia);
+  Object.defineProperty(window, "matchMedia", {
```
herox215: `installHarness` calls both `vi.stubGlobal("matchMedia", ...)` AND `Object.defineProperty(window, "matchMedia", ...)`. `vi.stubGlobal` is sufficient and is reversed by `vi.unstubAllGlobals()` in `afterEach`. The `Object.defineProperty` call is redundant and is NOT reversed by `unstubAllGlobals` — it persists across tests if the configurable descriptor is later set to false by some other code path. Drop the `Object.defineProperty` call here and at line 235.

— Aristotle

src/store/__tests__/viewportStore.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    const unsubB = store.subscribe((s) => seenB.push(s.mode));
+
+    harness.set(Q_TABLET, true);
+    expect(seenA[seenA.length - 1]).toBe("tablet");
+    expect(seenB[seenB.length - 1]).toBe("tablet");
+    unsubA();
+    unsubB();
+  });
+
+  it("removes every MQL listener it added once the last subscriber unsubscribes", () => {
```
herox215, line 164: The listener-cleanup test only exercises subscribe A + subscribe B → unsubscribe A → unsubscribe B. It does not cover the subscribe A → unsubscribe A → subscribe B cycle with two distinct subscribers (the re-attach test at line 190 uses a single subscriber per cycle). A multi-subscriber teardown-then-reattach (subscribe A + B, unsubscribe both, subscribe C) is the edge case where `readable`'s internal counter must reach zero before `start` re-runs — omitting it leaves the re-entry path untested.

— Aristotle

src/components/CommandPalette/CommandPalette.svelte
```diff
@@ -237,6 +237,9 @@
     color: var(--color-text);
     font-size: 14px;
     text-align: left;
+    /* #385 — token undefined on desktop → fallback `auto` (byte-identical
+       to current padding-only height); coarse → 44px. */
+    min-height: var(--vc-hit-target, auto);
```
herox215, line 242: The comment says "fallback `auto` (byte-identical to current padding-only height)" but `auto` is the CSS initial value for `min-height`, meaning no minimum constraint — it does not equal any specific px value. The comment is technically misleading: what is byte-identical is that the element continues to be sized by its content/padding, not that `auto` equals the prior height. Change the comment to: `/* #385 — fallback auto = no constraint (initial value), so desktop sizing is padding-driven as before; coarse → 44px. */`

— Aristotle

src/components/Sidebar/Sidebar.svelte
```diff
@@ -1038,6 +1038,9 @@
     cursor: pointer;
     background: var(--color-surface);
     color: var(--color-text);
+    /* #385 — token undefined on desktop → fallback `auto` (byte-identical
+       padding-only height); coarse → 44px. */
+    min-height: var(--vc-hit-target, auto);
```
herox215, line 1043: Same misleading "byte-identical to current padding-only height" phrasing for `auto` fallback — repeated across Sidebar.svelte (`.vc-confirm-btn` at line ~1043, `.vc-new-menu-item` at line ~1201), TreeRow.svelte (`.vc-confirm-btn`), ContextMenu.svelte, PasswordPromptModal.svelte, and QuickSwitcherRow.svelte. The comment here should say the desktop element remains padding-sized (no min-height constraint), not that `auto` equals some prior px value. Fix the wording in all seven `auto`-fallback occurrences consistently.

— Aristotle

src/test/setup.ts
```diff
@@ -6,3 +6,24 @@ export function withFakeTimers() {
   vi.useFakeTimers();
   return () => vi.useRealTimers();
 }
+
+// jsdom does not implement matchMedia. Provide a non-matching stub so
+// modules that touch viewport-aware code (viewportStore, tabMorph's
+// prefers-reduced-motion check, etc.) don't crash on import. Suites that
+// need controllable behavior override via vi.stubGlobal("matchMedia", ...).
+if (typeof window !== 'undefined' && !window.matchMedia) {
+  Object.defineProperty(window, 'matchMedia', {
```
herox215, line 15: `installHarness` (test line 84) warns that `Object.defineProperty` "would NOT be reversed by unstubAllGlobals and would leak across tests." This file uses `Object.defineProperty` for the same property in the same test run. The two are not contradictory — `configurable: true` + `writable: true` means `vi.stubGlobal` can override it and `unstubAllGlobals` restores to this baseline — but the comment in `installHarness` omits that nuance and will mislead the next developer who reads both files together. Qualify the comment: e.g. "An *additional* `Object.defineProperty` inside a test body (without `configurable: true`) would not be reversed…"

— Aristotle

Issue #387, opened 2026-04-30
feat(mobile): #74.3 long-press → context-menu bridge


herox215:
Sub-ticket of #74. Depends on #74.1.

## Goal

Provide a long-press primitive that dispatches existing context-menu handlers, so right-click parity exists on touch. Also wires "New Note here" on folder long-press (the only mobile quick-capture entry point — there is no FAB).

## Approach

**`src/lib/actions/longPress.ts`** (new) — Svelte action via Pointer Events:

```
use:longPress={{ duration: 500, onLongPress: (e) => ..., moveTolerance: 10 }}
```

- `pointerdown` starts a 500ms timer.
- `pointermove` >10px → cancel.
- `pointerup` before threshold → cancel.
- `pointercancel` → cancel.
- Synthesizes `clientX/clientY` and calls consumer callback. Consumers reuse their existing context-menu handlers verbatim.
- **MUST** call `e.preventDefault()` on `pointerdown` AND on any synthetic `contextmenu` to suppress iOS callout / Android selection magnifier.

Pointer Events (not `touchstart`) so mouse-long-press on hybrid devices flows through the same path.

## Wiring

1. `TreeRow.svelte` — `use:longPress` on `<li>`. Existing `oncontextmenu={handleRowContextMenu}` stays for right-click. Coexist.
2. `BookmarksPanel.svelte` — same pattern, calls existing `openContextMenu(e, relPath)`.
3. `EditorPane.svelte` — long-press on selection text dispatches `handleContextMenuRequest(view, x, y)`. CodeMirror's tap-to-select fires harmlessly within the 500ms window.

## "New Note here" item

Folder long-press → existing `TreeNodeContextMenu` shows. Add a `New Note here` action that creates the note in the long-pressed folder (vault root if long-press on the empty tree area).

## Global `contextmenu` suppressor

`VaultLayout.svelte` 669–679 unchanged. Long-press is independent of the `contextmenu` event path.

## Tests

- Vitest: `longPress.test.ts` — synthesize `pointerdown`, fake-timer to threshold, callback fires with synthetic coords. Assert cancel on move/up/cancel.
- Vitest: `TreeRow.longPress.test.ts` — long-press flips `showContextMenu`, sets `menuPos`. Reuse `TreeNodeContextMenu.test.ts` harness.
- Vitest: `TreeRow.newNoteHere.test.ts` — folder long-press → "New Note here" creates note in the long-pressed folder.
[...]

PR #399, opened 2026-04-30
feat(mobile): #74.3 long-press → context-menu bridge (#387)

herox215:
## Summary

Touch parity for the right-click context menu and a vault-root quick-capture entry on mobile (#387, child of #74).

- New `src/lib/actions/longPress.ts` Svelte action — Pointer-Events primitive (500 ms hold, 10 px Euclidean tolerance). `pointermove` / `up` / `cancel` listened on `document` with capture so CodeMirror's `setPointerCapture` cannot redirect events past us mid-hold. Single-pointer enforcement: a second concurrent `pointerdown` cancels both — multi-touch never surfaces a menu.
- One-shot capture-phase `contextmenu` suppressor uses `stopImmediatePropagation()` to defeat CM6's bubble-phase `view.dom` listener (`editorContextMenuExtension`), which is otherwise indifferent to `defaultPrevented`. `AbortController` ties five teardown signals: one-shot fire, 600 ms ceiling (≈6× iOS/Android synthesis window), `window.blur`, next `pointerdown`, action `destroy()`. Cannot leak.
- `e.preventDefault()` on `pointerdown` gated to `pointerType === "touch"` only — pen and mouse pass through, so Surface pen-drag and desktop focus-on-mousedown stay intact.
- `strict: true` option filters bubbling pointerdowns at the action layer; used by the empty-tree-area wiring.
- Wired into `TreeRow` (`.vc-tree-row`, mirrors `oncontextmenu` surface), `BookmarksPanel` (each `.vc-bookmark-row`), `EditorPane` (`.vc-editor-content`, resolves active `EditorView` from `viewMap` at fire-time), and `Sidebar` empty-tree area (`.vc-sidebar-tree` strict, opens a single-item "New note here" menu rooted at the vault path).
- Folder context menu: `handleNewFileHere` → `handleNewNoteHere`; label "New file here" → "New note here" (issue spec wording, sentence-case to match siblings).
- Boy Scout: TreeRow `getVaultRoot()` subscribe-then-unsub anti-pattern → `get(vaultStore).currentPath`; dead vault-prefix ternary in `confirmDelete` cleaned up.
- `Sidebar.handleNewFile` gains an optional `targetFolder?: string | unknown` parameter; type-guarded so the existing `onclick={handleNewFile}` bindings (which pass the `MouseEvent`) still fall through to selection-based fallback.

Pre-existing scope flagged but **not** fixed in this ticket: `getSelectedFolder()` only inspects the top-level `rootEntries` and misses nested-selected folders. Empty-area long-press always targets vault root explicitly, so the bug doesn't bleed into the new path.

## Test plan

- [x] `pnpm vitest run src/lib/actions/__tests__/longPress.test.ts` — 20/20 green (timer, move/up/cancel, preventDefault matrix touch/mouse/pen, multi-touch, CM6 setPointerCapture redirect, document-capture stopImmediatePropagation, 600 ms ceiling, window.blur teardown, next-pointerdown teardown, scroll cancel, strict-target gate, update/destroy).
- [x] `pnpm vitest run src/components/Sidebar/ src/components/Bookmarks/ src/lib/actions/` — 24 files, 79 tests green.
- [x] `pnpm vitest run src/components/Editor/` — 38 files, 317 tests green (no regression in CM6 right-click path or viewer branches).
- [ ] Manual smoke on a real mobile WebView (deferred — `pnpm dev` desktop browser cannot synthesize touch).
- [ ] WDIO touch-sim spec — explicitly out of scope; deferred to a separate infra ticket.

## Architecture notes worth highlighting for review

- The CM6 dual-fire concern is real for the editor surface only. Tree / Bookmarks long-press cannot reach `view.dom` (separate DOM subtree). The pane-level test (`EditorPane.longPress.test.ts`) asserts no parallel CM6 menu after a long-press fire.
- `AbortController` ordering: `disarmSuppressor()` runs **before** `cancelTimer()` resets `activePointerId` on a multi-touch second pointerdown, so re-entry into the action sees a clean state and the suppressor cannot re-arm against the next legit right-click.

### DIFF
--- a/src/components/Editor/__tests__/EditorPane.longPress.test.ts
+++ b/src/components/Editor/__tests__/EditorPane.longPress.test.ts
@@ -1,0 +1,148 @@
+// EditorPane long-press → context menu (#387). Touch parity for the existing
+// right-click context-menu path (#301), implemented by wiring `use:longPress`
+// on `.vc-editor-content`. The synthetic contextmenu that some platforms
+// dispatch after a touch hold must NOT also bubble through CM6's
+// `editorContextMenuExtension`, which would surface the menu twice.
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render } from "@testing-library/svelte";
+import { tick } from "svelte";
+
+const { readFile } = vi.hoisted(() => ({ readFile: vi.fn() }));
+
+vi.mock("../../../ipc/commands", () => ({
+  readFile,
+  writeFile: vi.fn().mockResolvedValue("0".repeat(64)),
+  getFileHash: vi.fn().mockResolvedValue("0".repeat(64)),
+  mergeExternalChange: vi.fn().mockResolvedValue({ outcome: "clean", merged_content: "" }),
+  getResolvedLinks: vi.fn().mockResolvedValue(new Map()),
+  getResolvedAttachments: vi.fn().mockResolvedValue(new Map()),
+  createFile: vi.fn().mockResolvedValue(""),
+  getLinkGraph: vi.fn().mockResolvedValue({ nodes: [], edges: [] }),
+  getLocalGraph: vi.fn().mockResolvedValue({ nodes: [], edges: [] }),
+  getBacklinks: vi.fn().mockResolvedValue([]),
+  getOutgoingLinks: vi.fn().mockResolvedValue([]),
+  getUnresolvedLinks: vi.fn().mockResolvedValue([]),
+  listTags: vi.fn().mockResolvedValue([]),
+  countWikiLinks: vi.fn().mockResolvedValue(0),
+  suggestLinks: vi.fn().mockResolvedValue([]),
+  searchFulltext: vi.fn().mockResolvedValue([]),
+  searchFilename: vi.fn().mockResolvedValue([]),
+  listDirectory: vi.fn().mockResolvedValue([]),
+  invoke: vi.fn(),
+  normalizeError: (e: unknown) => e,
+}));
+
+vi.mock("../../../ipc/events", () => ({
+  listenFileChange: vi.fn().mockResolvedValue(() => {}),
+  listenVaultStatus: vi.fn().mockResolvedValue(() => {}),
+  listenBulkChangeStart: vi.fn().mockResolvedValue(() => {}),
+  listenBulkChangeEnd: vi.fn().mockResolvedValue(() => {}),
+  listenIndexProgress: vi.fn().mockResolvedValue(() => {}),
+}));
+
+vi.mock("../../Graph/GraphView.svelte", async () => {
+  // @ts-ignore
+  const { default: Empty } = await import("./emptyComponent.svelte");
+  return { default: Empty };
+});
+vi.mock("../../Graph/graphRender", () => ({
+  mountGraph: vi.fn(),
+  updateGraph: vi.fn(),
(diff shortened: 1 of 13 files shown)


src/lib/actions/longPress.ts
```diff
@@ -0,0 +1,202 @@
...
+    // synthetic contextmenu can race the consumer's state updates on slow
+    // mobile WebViews; arming first guarantees we win that race.
+    suppressorAbort = new AbortController();
+    const signal = suppressorAbort.signal;
+
+    document.addEventListener(
+      "contextmenu",
+      (e) => {
+        e.preventDefault();
+        e.stopPropagation();
```
herox215: `stopPropagation()` is called at line 103 before `stopImmediatePropagation()` at line 107. Both are no-ops once `stopImmediatePropagation()` runs, but `stopPropagation()` here is at best redundant — capture-phase `stopImmediatePropagation()` already prevents all subsequent listeners. Remove `stopPropagation()` to avoid giving a false impression that the two calls serve different purposes.

— Aristotle

src/lib/actions/longPress.ts
```diff
@@ -0,0 +1,202 @@
...
+        // stopImmediatePropagation in capture-phase aborts dispatch to
+        // any subsequent listener including bubble-phase handlers
+        // attached to the original target (e.g. CM6's view.dom).
+        e.stopImmediatePropagation();
+        disarmSuppressor();
+      },
+      { capture: true, signal },
+    );
+
+    setTimeout(() => disarmSuppressor(), SUPPRESSOR_CEILING_MS);
```
herox215: The 600ms ceiling `setTimeout` at line 113 is not stored in a variable, so it cannot be cancelled in `destroy()`. When a component mounts, fires a long-press, then unmounts within 600ms — a common lifecycle on rapid navigation — this timer fires after `destroy()` and calls `disarmSuppressor()` on an already-null `suppressorAbort`. Harmless in isolation, but every such lifecycle leaks one 600ms timer. Store the return value and `clearTimeout` it in `destroy()` alongside `cancelTimer()`.

— Aristotle

src/components/Editor/__tests__/EditorPane.longPress.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    await flushAsync();
+
+    const surface = container.querySelector(".vc-editor-content") as HTMLElement;
+    expect(surface).toBeTruthy();
+    expect(container.querySelector(".vc-context-menu")).toBeNull();
+
+    surface.dispatchEvent(pointerEvent("pointerdown", { clientX: 120, clientY: 80 }));
+    // Real timers — the action's setTimeout(500) drives this. 600ms gives a
+    // safe margin without making the test sluggish.
+    await new Promise((resolve) => setTimeout(resolve, 600));
```
herox215: Both EditorPane long-press tests wait 600ms of real wall-clock time (`await new Promise(resolve => setTimeout(resolve, 600))`). The rest of the test suite uses `vi.useFakeTimers()` + `vi.advanceTimersByTime()` (see `longPress.test.ts`, `BookmarksPanel.longPress.test.ts`). These two tests add ~1.2s to every local run and are non-deterministic under CPU contention. Use fake timers instead — the longPress action wires a plain `setTimeout`, which fake timers control without issue.

— Aristotle

Issue #63, opened 2026-04-15
Reading Mode vs Edit Mode toggle (Cmd+E)


herox215:
## Summary

Add a per-tab toggle between **Edit Mode** (current CM6 with live preview) and **Reading Mode** (fully rendered markdown, no editor affordances). Bound to Obsidian's standard `Cmd+E` hotkey. The toggle state is persisted per tab across sessions.

## Context

Today every markdown tab mounts a CodeMirror 6 view with the live-preview plugin (`src/components/Editor/livePreview.ts`). Live Preview hides syntax tokens and renders emphasis/headings/links inline, but it never leaves the editor surface — the document stays editable, the cursor is visible, and gutter/selection affordances remain. There is no dedicated read-only HTML view.

Obsidian users rely on `Cmd+E` to flip a note between two visually distinct modes:

- **Edit Mode** — what we already have (CM6 + live preview).
- **Reading Mode** — a fully rendered HTML view of the markdown with no editor chrome. Used for long-form reading, reviewing a finished note, or sharing the screen.

The toggle lives per tab: opening the same file in a new tab should start in the user's default (Edit) unless persisted otherwise; switching tabs restores each tab's own mode.

Key integration points already in place:

- `src/components/Editor/EditorPane.svelte` mounts the CM6 view per tab through `mountEditorView()` and keeps views in a module-level `viewMap` keyed by tab id. This is the branching surface — in Reading Mode the pane should render a different component (HTML renderer) into the same `data-tab-id` container rather than the CM6 host.
- `src/store/tabStore.ts` owns the `Tab` interface. A new `viewMode: \"edit\" | \"read\"` field belongs here, alongside `scrollPos` / `cursorPos` which are already persisted.
- `src/lib/commands/defaultCommands.ts` is the registry entry point for the new `editor:toggle-reading-mode` command. Note the existing `TOGGLE_SIDEBAR_ALT` already claims `Cmd+Shift+E` — **plain `Cmd+E` is still free**, which matches Obsidian's binding exactly.
[...]

PR #401, opened 2026-04-30
feat(mobile): #388 read-mode default on mobile + topbar pencil toggle

### LINKED ISSUE #388: feat(mobile): #74.4 read-mode / edit-mode toggle (Obsidian-style)
Sub-ticket of #74. Depends on #74.1.

## Goal

On mobile, notes default to **read-mode** when opened (Obsidian behavior). A pencil-icon toggle in the topbar switches to edit-mode. Edit-mode persists per-tab, not globally.

## Rationale

Mobile users frequently open notes to glance, not edit. Edit-mode default on a touch device causes accidental edits while scrolling. Per-tab persistence means a user who opens note A in edit-mode keeps it editable when switching back, while a freshly opened note B starts read-only.

## Approach

- Read-mode renders Markdown via the existing markdown rendering pipeline (preview-mode if exists, otherwise add a minimal MDAST → DOM render path).
  - **Open question:** is there an existing read-mode renderer in the codebase, or do we need to add one? Investigation required during planning. If absent, this ticket grows or spawns a sub-ticket.
- Edit-mode renders CodeMirror 6 as today.
- `tabsStore.ts` — add `mode: "read" | "edit"` per tab. Default on mobile = `"read"`. Default on desktop = `"edit"` (no behavior change).
- Topbar pencil-icon toggle — visible on mobile only (`viewportStore.mode === "mobile"`).
- Tap on internal link in read-mode → navigate, target tab opens in read-mode.

## Tests

- Vitest: `tabsStore.mobileReadMode.test.ts` — opening a note on mobile defaults to read; on desktop defaults to edit.
- Vitest: topbar pencil toggle hidden on desktop, visible on mobile.
- Vitest: per-tab persistence — toggle one tab to edit, switch tabs, switch back, still edit.

## Out of scope

- Read-mode renderer feature parity with desktop (links, embeds, Mermaid, math, etc. — only what already renders).
- Inline editing within read-mode (tap-on-paragraph-to-edit). If we want it later, separate ticket.

herox215:
## Summary

Closes #388. Sub-ticket of #74. Implements the Obsidian-style mobile read-mode default plus a topbar pencil/book-open toggle that flips the active tab's mode. Per-tab persistence (already in place from #63) lets users keep tab A in edit while tab B stays read.

**Pre-investigation finding**: read-mode rendering already existed end-to-end from #63 (`ReadingView.svelte` + `markdownRenderer.ts` with markdown-it / DOMPurify / wiki-link / embed / heading-anchor pipelines, plus `Tab.viewMode` state and `tabLifecycleStore.toggleViewMode`). #388 reduced to:
1. UI-boundary helper deciding the default `viewMode` for new tabs based on viewport.
2. Optional `viewMode` hint on `tabStore.openTab` / `openFileTab`, applied on creation only (dedupe preserves the user's last explicit choice — see JSDoc).
3. Routing key user-facing markdown opens through `openFileAsTab` so they pick up the hint.
4. Mobile-only topbar pencil component.

**Architecture**: store is environment-agnostic — viewport awareness lives at `lib/tabKind.ts` (`defaultViewModeForViewport`). Existing `tabLifecycleStore` tests untouched (no matchMedia stub contamination).

## Boy Scout fixes folded in

The new mobile pencil makes both of these reachable for the first time:

- `VaultLayout.toggleActiveReadingMode` previously excluded only `image`/`unsupported` viewers. A `text`-viewer (.json/.txt/etc.) or canvas tab responding to Cmd/Ctrl+E would set `viewMode='read'` on a tab kind that has no ReadingView path → broken state.
- `EditorPane.paneActiveTabSupportsReading` correctly excluded `image`/`unsupported`/`text` but missed canvas. A canvas tab would show the Breadcrumbs reading toggle but clicking it would `setViewMode('read')` while the canvas itself kept rendering.

Both fixed by extracting `tabSupportsReading()` to `lib/tabKind.ts` — single source of truth used by EditorPane, VaultLayout, the new TopbarReadingToggle, and (via the gating chain) Breadcrumbs.

## Decisions (documented in code + tests)

- **NEW notes default to edit on every viewport.** Click-to-create on an unresolved `[[Foo]]` matches `createNewNote` / `openTodayNote` / Sidebar new-file convention — the user typing or clicking a name is creating, presumably to write into. See `EditorPane.svelte` line 432 and `OutgoingLinksPanel.svelte` click-to-create branch. **EXISTING notes default to read on mobile.**
- **BookmarksPanel.handleOpenInSplit kept synchronous on `tabStore.openTab`.** Async `openFileAsTab` would race with the synchronous `tabStore.moveToPane('right')` immediately after. Split is desktop-only (gated by viewport in #386), so the lost mobile-read hint there is moot. Documented inline.
- **GraphView / LocalGraphPanel / CanvasView NOT routed through `openFileAsTab`.** All three call from inside non-async callbacks; promoting to async would create unhandled-promise hazards. Niche click paths — user can flip via the topbar pencil. Documented as known limitation.
- **Breadcrumbs reading toggle hides on mobile** so it never co-renders with the topbar pencil. Strictly more restrictive than today's gating (which is `tabId && viewMode !== undefined`). Desktop and tablet markdown unchanged.
- `defaultViewModeForViewport()` returns `TabViewMode` always (no `undefined` sentinel leaking). Returning `'edit'` on desktop is byte-identical in effect to today, since every reader coalesces `viewMode ?? 'edit'`.

## Heads-up for review

- `Breadcrumbs.svelte` lines 122-123 still have German strings (`"Bearbeitungsmodus"`, `"Lesemodus"`) inherited from the existing #63 toggle. The new `TopbarReadingToggle.svelte` mirrors this convention. Out of scope to translate — no i18n plan exists in the repo.
- Three commits on the branch:
  1. `39d065f` — RED failing tests (TDD)
  2. `d157f34` — `tabKind` helpers + optional viewMode hint on store openers
  3. `dff322f` — callsite routing + Boy Scout guard fix + TopbarReadingToggle + Breadcrumbs hide-on-mobile

## Files

**Source (10)**:
- `src/lib/tabKind.ts` — added `tabSupportsReading()` + `defaultViewModeForViewport()`
- `src/store/tabLifecycleStore.ts` — added optional `viewMode` arg to `openTab` and `openFileTab`
- `src/lib/openFileAsTab.ts` — markdown branch passes the hint
- `src/components/Editor/EditorPane.svelte` — wiki-link follow + click-to-create wired; inline `paneActiveTabSupportsReading` replaced with shared predicate
- `src/components/Editor/Breadcrumbs.svelte` — hide-on-mobile gate
- `src/components/Layout/VaultLayout.svelte` — topbar pencil wiring + `toggleActiveReadingMode` guard fix
- `src/components/Layout/TopbarReadingToggle.svelte` — new mobile-only component
- `src/components/Bookmarks/BookmarksPanel.svelte` — handleRowClick + handleOpenInNewTab routed (handleOpenInSplit stays sync)
- `src/components/Backlinks/BacklinksPanel.svelte` — handleBacklinkClick routed
- `src/components/OutgoingLinks/OutgoingLinksPanel.svelte` — resolved click routed; click-to-create stays direct (NEW = edit rule)

**Tests (6 new + 2 updated for new contract)**:
- `src/lib/__tests__/tabKind.test.ts` — predicate matrix + viewport hint (mocked viewportStore via vi.doMock seam)
- `src/store/tabLifecycleStore.test.ts` — viewMode hint creation + dedupe-preserves
- `src/lib/__tests__/openFileAsTab.test.ts` — dispatcher passes hint for markdown only
- `src/components/Editor/__tests__/EditorPane.wikiLinkMobileMode.test.ts` — wiki-link follow + click-to-create on mobile/desktop
- `src/components/Editor/__tests__/Breadcrumbs.mobileHide.test.ts` — desktop/tablet/mobile visibility
- `src/components/Layout/__tests__/TopbarReadingToggle.test.ts` — visibility matrix + click + per-tab persistence with reactive icon flip
- `src/components/Editor/__tests__/EditorPaneWikiLinkClickResolve.test.ts` — updated 3 spy assertions for new two-arg form
- `src/components/Bookmarks/__tests__/BookmarksPanel.test.ts` — updated spy assertion for the now-routed click path

## Test plan

- [x] Full vitest suite green: 147 files, 1404 tests
- [x] `pnpm typecheck` clean
- [ ] Manual mobile-emulation pass via `pnpm dev` + browser devtools (see Step 8 of plan): open markdown via tree, omnisearch, bookmarks, backlinks, outgoing-links, wiki-link click — all default to read on mobile, edit on desktop. Click-to-create new note → opens in edit. Topbar pencil flips, persists per-tab across switches, icon reflects state. Breadcrumbs toggle absent on mobile, present on desktop and tablet.
- [ ] Coordination with `frontend-impl` (#386) on `VaultLayout.svelte` topbar — pinged before push; placement on right side near backlinks/settings, separate from leading-edge drawer toggle area.

### DIFF
--- a/src/components/Bookmarks/BookmarksPanel.svelte
+++ b/src/components/Bookmarks/BookmarksPanel.svelte
@@ -10,6 +10,7 @@
   import { vaultStore } from "../../store/vaultStore";
   import { tabStore } from "../../store/tabStore";
   import { longPress, type LongPressDetail } from "../../lib/actions/longPress";
+  import { openFileAsTab } from "../../lib/openFileAsTab";
 
   const BOOKMARKS_COLLAPSED_KEY = "vaultcore-bookmarks-panel-collapsed";
 
@@ -57,7 +58,9 @@
   }
 
   function handleRowClick(relPath: string): void {
-    tabStore.openTab(toAbsPath(relPath));
+    // #388 — route through openFileAsTab so the dispatcher applies the
+    // viewport-aware viewMode default (mobile → read, desktop → edit).
+    void openFileAsTab(toAbsPath(relPath));
   }
 
   function handleRowKeydown(e: KeyboardEvent, relPath: string): void {
@@ -83,14 +86,20 @@
 
   function handleOpenInNewTab(relPath: string): void {
     closeContextMenu();
-    // tabStore.openTab dedupes by filePath and focuses existing — call it to
-    // open/focus the note in the active pane.
-    tabStore.openTab(toAbsPath(relPath));
+    // #388 — route through openFileAsTab so the dispatcher applies the
+    // viewport-aware viewMode default. Dedupe semantics (focus existing tab)
+    // are preserved by the underlying tabStore.openTab call.
+    void openFileAsTab(toAbsPath(relPath));
   }
 
   function handleOpenInSplit(relPath: string): void {
     closeContextMenu();
     const absPath = toAbsPath(relPath);
+    // #388 — kept synchronous on purpose: openFileAsTab is async and would
+    // race with the synchronous moveToPane("right") below, which would
+    // potentially move the wrong tab. Split view is desktop-only via the
+    // viewport gating in #386, so the lost mobile-read hint here doesn't
+    // affect the mobile flow.
     tabStore.openTab(absPath);
     tabStore.moveToPane("right");
   }
--- a/src/components/Bookmarks/BookmarksPanel.svelte
+++ b/src/components/Bookmarks/BookmarksPanel.svelte
@@ -95,11 +95,10 @@
   function handleOpenInSplit(relPath: string): void {
     closeContextMenu();
     const absPath = toAbsPath(relPath);
-    // #388 — kept synchronous on purpose: openFileAsTab is async and would
-    // race with the synchronous moveToPane("right") below, which would
-    // potentially move the wrong tab. Split view is desktop-only via the
-    // viewport gating in #386, so the lost mobile-read hint here doesn't
-    // affect the mobile flow.
+    // #388 — stays sync: handleOpenInSplit is a synchronous event handler
+    // and cannot `await openFileAsTab` without restructuring the call site.
+    // The mobile-aware viewMode hint is lost on the split path; split is
+    // desktop-only via viewport gating, so this is moot in practice.
     tabStore.openTab(absPath);
     tabStore.moveToPane("right");
   }
--- a/src/lib/__tests__/openFileAsTab.test.ts
+++ b/src/lib/__tests__/openFileAsTab.test.ts
@@ -91,3 +91,70 @@ describe("openFileAsTab", () => {
     expect(get(tabStore).tabs).toHaveLength(1);
   });
 });
+
+// #388 — viewport-aware default viewMode. Mobile users get markdown opens in
+// read mode; non-markdown viewers (image / text / unsupported) ignore the
+// hint because they have no Reading Mode path. The dispatcher reads the
+// helper from `tabKind.ts`, which itself reads `viewportStore` once per call.
+
+describe("openFileAsTab + viewMode (#388)", () => {
+  beforeEach(() => {
+    tabStore._reset();
+    readFile.mockReset();
+    vi.resetModules();
+    vi.doUnmock("../../store/viewportStore");
+  });
+
+  async function loadDispatcher(
+    mode: "desktop" | "tablet" | "mobile",
+  ): Promise<typeof openFileAsTab> {
+    const { readable } = await import("svelte/store");
+    vi.doMock("../../store/viewportStore", () => ({
+      viewportStore: readable({ mode, isCoarsePointer: mode === "mobile" }),
+    }));
+    // Re-mock the IPC layer for the freshly-imported dispatcher so the
+    // module graph after `vi.resetModules()` shares the same readFile spy.
+    vi.doMock("../../ipc/commands", () => ({ readFile }));
+    const mod = await import("../openFileAsTab");
+    return mod.openFileAsTab;
+  }
+
+  it("opens markdown with viewMode='read' on mobile", async () => {
+    const dispatch = await loadDispatcher("mobile");
+    const { tabStore: ts } = await import("../../store/tabStore");
+    ts._reset();
+    await dispatch("/vault/note.md");
+    const state = get(ts);
+    expect(state.tabs).toHaveLength(1);
+    expect(state.tabs[0]!.viewMode).toBe("read");
+  });
+
+  it("opens markdown with viewMode='edit' on desktop", async () => {
+    const dispatch = await loadDispatcher("desktop");
+    const { tabStore: ts } = await import("../../store/tabStore");
+    ts._reset();
+    await dispatch("/vault/note.md");
+    const state = get(ts);
+    expect(state.tabs[0]!.viewMode).toBe("edit");
+  });
+
+  it("does not pass the hint for image viewers (image has no Reading Mode)", async () => {
+    const dispatch = await loadDispatcher("mobile");
+    const { tabStore: ts } = await import("../../store/tabStore");
+    ts._reset();
+    await dispatch("/vault/photo.png");
+    const state = get(ts);
+    expect(state.tabs[0]!.viewer).toBe("image");
+    expect(state.tabs[0]!.viewMode).toBeUndefined();
+  });
+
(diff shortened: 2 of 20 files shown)


src/lib/tabKind.ts
```diff
@@ -1,6 +1,17 @@
...
 // subsequently loading the file content (or, for "unsupported", skipping it).
+//
+// Also hosts two viewport-aware helpers for the #388 mobile read-mode flow:
+//   - `tabSupportsReading(tab)`: predicate over tab shape, used wherever the
+//     UI needs to know whether Reading Mode applies to the active tab.
+//   - `defaultViewModeForViewport()`: read-once viewport hint passed to
+//     openers when creating a new tab. The store remains environment-agnostic
+//     — viewport awareness lives at this UI boundary.
+
+import { get } from "svelte/store";
```
herox215: `tabKind.ts` is a pure extension classifier ("no IPC, no filesystem access" per its own header comment). Adding a `viewportStore` import turns it into a UI-environment-coupled module — a lib that previously had zero side-effects now reads a Svelte store at import time. The store belongs at the call-site boundary (`openFileAsTab`, `EditorPane`) or in its own `lib/viewport.ts` helper. As written, any test that imports `getTabKind` or `getExtension` pulls in the matchMedia-dependent store initializer, and the module can no longer be used in a non-Svelte context (e.g., a Rust-driven IPC mock or a Node CLI tool). Extract `defaultViewModeForViewport` to a thin `lib/viewport.ts` seam so the classifier stays pure.

— Aristotle

src/lib/__tests__/openFileAsTab.test.ts
```diff
@@ -91,3 +91,70 @@ describe("openFileAsTab", () => {
...
 });
+
+// #388 — viewport-aware default viewMode. Mobile users get markdown opens in
+// read mode; non-markdown viewers (image / text / unsupported) ignore the
+// hint because they have no Reading Mode path. The dispatcher reads the
+// helper from `tabKind.ts`, which itself reads `viewportStore` once per call.
+
+describe("openFileAsTab + viewMode (#388)", () => {
+  beforeEach(() => {
+    tabStore._reset();
```
herox215: The `openFileAsTab + viewMode` describe block calls `vi.resetModules()` in `beforeEach` and then re-imports `tabStore` inside each `it` body. The top-level `tabStore._reset()` in `beforeEach` operates on the *pre-reset* module instance; the `ts._reset()` inside each `it` after `import("../../store/tabStore")` resets the *post-reset* instance. These are not guaranteed to be the same singleton — after `resetModules`, the dynamic import produces a fresh module with its own `_core` writable. If Vitest caches the top-level `tabStore` import (it does, until `resetModules` runs), then `tabStore._reset()` in `beforeEach` is resetting a stale instance that the test never reads from. The fix is to remove the top-level `import { tabStore }` and always import via `const { tabStore: ts } = await import(...)` inside `loadDispatcher`, or to hoist the reset into `loadDispatcher` after the dynamic import rather than in `beforeEach`.

— Aristotle

src/components/Bookmarks/BookmarksPanel.svelte
```diff
@@ -78,14 +81,20 @@
...
+    void openFileAsTab(toAbsPath(relPath));
   }
 
   function handleOpenInSplit(relPath: string): void {
     closeContextMenu();
     const absPath = toAbsPath(relPath);
+    // #388 — kept synchronous on purpose: openFileAsTab is async and would
+    // race with the synchronous moveToPane("right") below, which would
+    // potentially move the wrong tab. Split view is desktop-only via the
+    // viewport gating in #386, so the lost mobile-read hint here doesn't
```
herox215: `handleOpenInSplit` calls `tabStore.openTab(absPath)` without the `viewMode` hint — intentional and documented. However, the comment says the race risk is with `moveToPane("right")`, but the real problem is subtler: `openFileAsTab` is async because of the encryption gate and the UTF-8 probe, not because of `openTab` itself. For a `.md` file with no encryption gate hit, `openFileAsTab` resolves after a single microtask. A future reader trying to make this async should know the race is specifically with the synchronous `moveToPane` that immediately follows — the tab opened by `openFileAsTab` would already be the active tab by the time `moveToPane` fires, so the split would work correctly. The documented justification ("async would race with synchronous moveToPane") is plausible but the mechanism is not fully accurate; the comment may mislead the next person who tries to fix this. Consider: "openFileAsTab cannot be awaited here because the caller is a synchronous event handler; converting to async would require restructuring the call site."

— Aristotle

Issue #397, opened 2026-04-30
feat(mobile): #74.12 mobile burger menu (sheet with secondary panels + settings)


herox215:
Sub-ticket of #74. Depends on #74.1, #74.2, #74.5 (the More tab triggers this).

## Goal

Provide a single bottom-sheet "burger menu" on mobile that hosts all secondary navigation surfaces: Backlinks, Bookmarks, Outline, Outgoing Links, Properties, and Settings. Triggered by the **More** tab in the bottom-tab-bar (#389).

## Rationale

User chose Variante C (hybrid) — bottom-tab-bar reduced to 3 primary tabs (Files / Search / More), and the long tail of secondary panels lives behind a single burger affordance. Avoids cramming 7+ items into the bottom-tab-bar while still giving thumb-reach access.

## Approach

- New `src/components/Layout/MobileBurgerSheet.svelte`.
- Renders as a bottom-sheet (drag handle, swipe-down to dismiss, scrim, `env(safe-area-inset-bottom)`).
- Mounted only when `viewportStore.mode === "mobile"` and `mobileBurgerOpen === true` (new flag in the viewport-aware UI store).
- Triggered by tapping More in `MobileTabBar.svelte`.

## Sheet content (list of action rows)

Each row navigates to a sub-sheet or opens the corresponding existing panel:

1. **Backlinks** → opens Backlinks panel as full bottom-sheet (replaces right-sidebar route)
2. **Bookmarks** → opens BookmarksPanel as full bottom-sheet
3. **Outline** → opens Outline view as full bottom-sheet
4. **Outgoing Links** → opens OutgoingLinks view as full bottom-sheet
5. **Properties** → opens Properties bottom-sheet (#393, separate ticket — burger is just the entry point)
6. **Settings** → opens Settings full-screen sheet (#394, separate ticket — burger is just the entry point)

Burger sheet itself does NOT render the panel content — it's a router/menu. Each item drills into the existing panel component, promoted to a sheet via CSS (same pattern as #386 right-sidebar promotion that's now retired).

## Visual / Interaction

- Sheet height: ~50% of viewport, drag handle at top, scrim above.
- Each row: 56px tall (same as bottom-tab-bar height), icon left + label + chevron-right.
- Icons (Lucide): Link2 (Backlinks), Bookmark (Bookmarks), List (Outline), ArrowUpRight (Outgoing), FileText (Properties), Settings (Settings).
[...]

PR #403, opened 2026-04-30
feat(mobile): bottom tab bar (Files / Search / More) (#389)

### LINKED ISSUE #389: feat(mobile): #389 bottom tab bar (Files / Search / More)
Sub-ticket of #74. Depends on #74.1, #74.2.

## Scope (REVISED — 3 tabs, not 4)

Replace the desktop sidebar header tabs with a bottom-tab-bar on mobile only. **Three tabs**: Files / Search / More. Backlinks and Bookmarks moved into the new Mobile Burger Menu (#74.12).

## Tabs

1. **Files** — opens left sidebar drawer (redundant with topbar button, but standard mobile-app pattern; bottom is thumb-reachable)
2. **Search** — opens OmniSearch / Quick-Switcher as bottom sheet
3. **More** — opens the Mobile Burger Menu (#74.12) as a bottom sheet, containing: Backlinks, Bookmarks, Outline, Outgoing, Properties, Settings

## Approach

- New `src/components/Layout/MobileTabBar.svelte`.
- Renders only when `viewportStore.mode === "mobile"`.
- Fixed bottom, `env(safe-area-inset-bottom)` padding.
- Hit-targets 44px minimum (uses `--vc-hit-target` from #74.1).
- Active tab indicator (color shift + pill background per Vitruvius brief).
- Tab state stored in viewport-aware UI store (not the tabs store).

## Tokens & Constants (from Vitruvius brief)

- Height **56px**.
- Background `var(--color-surface)`. Top border `1px solid var(--color-border)`.
- Padding-bottom: `env(safe-area-inset-bottom)`.
- No shadow, no border-radius on container.
- Icons: Lucide, size 20, strokeWidth 1.5:
  - Files → `Files`
  - Search → `Search`
  - More → `Menu` (the Burger icon; consistent with the More-as-Burger semantic)
- Inactive: `var(--color-text-muted)`. Active: `var(--color-accent)` + pill background `var(--color-accent-bg)`.
- Labels always visible: `font-size: 11px; font-weight: 500; font-family: var(--vc-font-body)`.
- Mobile-only `@media (max-width: 699px)` — harmonized with #386.
- ARIA `role="tablist"`, per-tab `role="tab"` + `aria-selected` + `aria-controls`. Arrow-key navigation.

## Vitruvius input

- "Backlinks" tab dropped (moved to Burger), so the label-truncation question is moot.
- Settings access on mobile → not a tab here; lives inside the Burger (#74.12).

## Tests

- Vitest: `MobileTabBar.test.ts` — visible only on mobile mode, click switches state, More tab opens burger sheet.
- WDIO browser: `bottom-tab-bar.spec.ts` — tab through Files/Search/More, verify active indicator + correct sheet/drawer opens.

## Out of scope

- Burger sheet content + behavior — see #74.12.
- Tab ordering preference / customization.

herox215:
## Summary

Adds the mobile bottom tab bar per #389 (Vitruvius brief in issue comments). Three tabs: Files / Search / More. Renders only on mobile (`viewportStore.mode === "mobile"`).

- **Files** → opens the existing drawer (open-only — native iOS/Android pattern; close stays owned by scrim, swipe-left, Escape).
- **Search** → closes the drawer first, then opens OmniSearch in filename mode. Close-first is required because the drawer scrim sits at z-index 49 above the tab bar (40); without it the scrim consumes the tap and costs the user a second click.
- **More** → closes the drawer and shows a `Mehr-Menü folgt` info toast (placeholder until #397 lands; TODO comment marks the wire-up site).

### Architecture
- `MobileTabBar.svelte` is parent-gated: VaultLayout decides via `{#if isMobile}`. Component does NOT subscribe to `viewportStore` — verified by a source-level test.
- All state (drawer + modals) stays in VaultLayout; tab bar only emits via callbacks.

### ARIA
- `<div role="tablist" aria-label="Main navigation">` (W3C ARIA tabs uses div; Svelte's `a11y_no_noninteractive_element_to_interactive_role` rejects `<nav role="tablist">`).
- `aria-selected` on EVERY tab — never omitted (ARIA tabs §3.23).
- `aria-controls="vc-mobile-drawer"` on Files tab; Search/More omit it (OmniSearch has no stable id; More wires later).
- Roving `tabindex` follows the active tab; Files acts as the entry point when nothing is active.
- Arrow-Left/Right cycle and wrap; Home/End jump to ends.

### Layout
- 56px tall, `position: fixed`, `background: var(--color-surface)`, `border-top: 1px solid var(--color-border)`, `z-index: 40` (below drawer 50, well below modals ≥199).
- `padding-bottom: env(safe-area-inset-bottom)` for notched devices.
- Active state: pill `var(--color-accent-bg)` + color `var(--color-accent)`; inactive `var(--color-text-muted)`. No border accent.
- Lucide icons (Files / Search / Menu) at 20px / strokeWidth 1.5.
- Editor padding-bottom of `calc(56px + env(safe-area-inset-bottom))` inside the existing `@media (max-width: 699px)` block — applied to `.vc-layout-editor` (flex column) because padding on `.vc-vault-layout` would be a no-op (display:grid + height:100vh + overflow:hidden clips it).

### Plan v2 caveats addressed (Socrates v1)
1. Scrim-swallows-tap → Search/More handlers explicitly close the drawer first.
2. `padding-bottom` on grid container is a no-op → moved to `.vc-layout-editor`.
3. Files = open-only → early-return when drawer already open.
4. Self-gating is a layering leak → component takes no viewportStore dep; parent owns the gate.
5. `aria-selected` on every tab → always present.
6. OmniSearch z-index → verified 200/201 (above drawer 50 and tab bar 40).
7. Drawer-open + Search-tap regression → covered by WDIO test "Search tab opens OmniSearch (and closes the drawer first if open)".
8. CSS placement → component `<style>` `@media` block, NOT global tailwind.css.

## Test plan
- [x] `pnpm vitest run src/components/Layout/` — 31/31 green (10 new MobileTabBar cases).
- [x] `pnpm build` — clean.
- [ ] Manual UAT: resize <700px, tab bar appears at the bottom; 3 tabs render with icons + German labels.
- [ ] Manual UAT: Files opens drawer; second tap leaves it open; Files cannot close it (close via scrim/swipe/Escape).
- [ ] Manual UAT: open drawer via Files, tap Search → drawer closes AND OmniSearch opens (single tap).
- [ ] Manual UAT: tap More → "Mehr-Menü folgt" toast.
- [ ] Manual UAT: arrow-key nav across the 3 tabs (focus the bar with Tab first).
- [ ] Manual UAT: resize to desktop → tab bar disappears, editor padding restores.

Closes #389. Refs #74.

### DIFF
--- a/e2e/specs/bottom-tab-bar.spec.ts
+++ b/e2e/specs/bottom-tab-bar.spec.ts
@@ -102,17 +102,29 @@ describe("Mobile bottom tab bar (#389)", () => {
     });
   });
 
-  it("More tab shows the placeholder toast (until #397)", async () => {
+  it("More tab closes the drawer first if open AND shows the placeholder toast", async () => {
+    // Mirror of the Search-tab regression: the close-drawer-first guard in
+    // handleMobileMoreTab must work the same way as Search. Without this
+    // test, the More-tab close-first line could regress silently.
+    await clickFilesTab();
+    await browser.waitUntil(isDrawerOpen, { timeout: 3000 });
     await clickMoreTab();
     // toastStore renders into a host that stays in DOM; selector covers
     // the common conventions in this repo.
     const toast = await browser.$(".vc-toast, [role='status']");
     await toast.waitForDisplayed({ timeout: 3000 });
+    await browser.waitUntil(async () => !(await isDrawerOpen()), {
+      timeout: 3000,
+      timeoutMsg: "Drawer should close when More tab is tapped",
+    });
   });
 
   it("absent at desktop size", async () => {
     await browser.setWindowSize(1280, 900);
-    // Wait for the @media-driven re-render.
+    // Wait for the @media-driven re-render. Window-size restore is owned
+    // by the `after` hook — keeping the restore out of the test body means
+    // a `waitUntil` timeout produces the original assertion failure
+    // cleanly rather than hiding behind an unguarded restore line.
     await browser.waitUntil(async () => {
       const bars = await browser.$$(".vc-mobile-tab-bar");
       return bars.length === 0;
@@ -120,7 +132,5 @@ describe("Mobile bottom tab bar (#389)", () => {
       timeout: 3000,
       timeoutMsg: "Tab bar should be absent at desktop size",
     });
-    // Restore mobile size for the after-hook contract.
-    await browser.setWindowSize(600, 900);
   });
 });
--- a/e2e/specs/bottom-tab-bar.spec.ts
+++ b/e2e/specs/bottom-tab-bar.spec.ts
@@ -1,0 +1,126 @@
+import { createTestVault, type TestVault } from "../helpers/vault.js";
+import { openVaultInApp } from "../helpers/open-vault.js";
+
+/**
+ * #389 — bottom-tab-bar e2e coverage.
+ *
+ * Resizes the WebKitWebDriver window to 600x900 (matches the #386 spec
+ * convention) so the responsive @media (max-width: 699px) branch fires
+ * and the tab bar mounts.
+ *
+ * Each test is self-contained — no test depends on the side effects of
+ * the one before it. afterEach closes the drawer if any test left it open.
+ */
+describe("Mobile bottom tab bar (#389)", () => {
+  let vault: TestVault;
+  let restoreSize: { width: number; height: number } | null = null;
+
+  before(async () => {
+    vault = createTestVault();
+    await openVaultInApp(vault.path);
+    restoreSize = await browser.getWindowSize();
+    await browser.setWindowSize(600, 900);
+  });
+
+  after(async () => {
+    if (restoreSize) {
+      await browser.setWindowSize(restoreSize.width, restoreSize.height);
+    }
+    vault.cleanup();
+  });
+
+  async function isDrawerOpen(): Promise<boolean> {
+    const drawer = await browser.$(".vc-layout-sidebar");
+    const cls = (await drawer.getAttribute("class")) ?? "";
+    return cls.includes("vc-layout-sidebar--mobile-open");
+  }
+
+  async function clickFilesTab(): Promise<void> {
+    await browser.execute(() => {
+      (document.getElementById("vc-mobile-tab-files") as HTMLElement | null)?.click();
+    });
+  }
+
+  async function clickSearchTab(): Promise<void> {
+    await browser.execute(() => {
+      (document.getElementById("vc-mobile-tab-search") as HTMLElement | null)?.click();
+    });
+  }
+
+  async function clickMoreTab(): Promise<void> {
+    await browser.execute(() => {
+      (document.getElementById("vc-mobile-tab-more") as HTMLElement | null)?.click();
+    });
+  }
+
+  afterEach(async () => {
+    if (await isDrawerOpen()) {
+      await browser.keys(["Escape"]);
+      await browser.waitUntil(async () => !(await isDrawerOpen()), { timeout: 3000 }).catch(() => {});
+    }
+    // Close OmniSearch / any modal that a test left open.
+    await browser.keys(["Escape"]);
+  });
+
+  it("renders the tab bar at mobile size", async () => {
+    const bar = await browser.$(".vc-mobile-tab-bar");
+    await bar.waitForDisplayed({ timeout: 3000 });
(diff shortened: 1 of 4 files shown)


e2e/specs/bottom-tab-bar.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    // Wait for the @media-driven re-render.
+    await browser.waitUntil(async () => {
+      const bars = await browser.$$(".vc-mobile-tab-bar");
+      return bars.length === 0;
+    }, {
+      timeout: 3000,
+      timeoutMsg: "Tab bar should be absent at desktop size",
+    });
+    // Restore mobile size for the after-hook contract.
+    await browser.setWindowSize(600, 900);
```
herox215: `setWindowSize(600, 900)` restore is not guarded by `try/finally`. If `waitUntil` times out, the window stays at 1280px through `afterEach`. The in-test restore is stated to be "for the after-hook contract" (comment at line above) but `after()` already handles the true restore via `restoreSize` — this line is only redundant safety, and absent a `try/finally` it provides none. Wrap the restore in a `try/finally` block or remove the comment implying it is load-bearing.

— Aristotle

e2e/specs/bottom-tab-bar.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+    });
+  });
+
+  it("More tab shows the placeholder toast (until #397)", async () => {
+    await clickMoreTab();
+    // toastStore renders into a host that stays in DOM; selector covers
+    // the common conventions in this repo.
+    const toast = await browser.$(".vc-toast, [role='status']");
+    await toast.waitForDisplayed({ timeout: 3000 });
+  });
```
herox215, line 120: The "More tab" test never opens the drawer before tapping More, so `handleMobileMoreTab`'s `mobileDrawerOpen = false` line is never exercised. Test 4 (Search) correctly sets the drawer open first and asserts it closes — the More path needs the same regression check. Without it, removing `mobileDrawerOpen = false` from `handleMobileMoreTab` leaves this test green.

— Aristotle

src/components/Layout/MobileTabBar.svelte
```diff
@@ -0,0 +1,147 @@
...
+  type TabId = "files" | "search" | "more";
+  interface TabSpec {
+    id: TabId;
+    label: string;
+    icon: typeof Files;
+    controls?: string;
+    onSelect: () => void;
+  }
+
+  const tabs: TabSpec[] = $derived([
```
herox215: `tabs` is a static structure — ids, labels, and icons never change at runtime. Wrapping it in `$derived()` causes the entire array to re-create whenever any `$props()` value changes (including the callbacks). The keyed `{#each}` recovers correctly, so there is no visual bug, but `$derived` implies the value is computed from reactive inputs in a meaningful way. Declare `tabs` as a plain `const` and reference the callbacks directly in the `onSelect` slot; only `activeId` genuinely needs `$derived`.

— Aristotle

src/components/Layout/__tests__/MobileTabBar.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    // Verifies the v2 #4 fix: MobileTabBar must not self-gate on viewport.
+    // Read the source instead of relying on a runtime probe (would still
+    // pass if a vi.mock blocked the import at runtime). Match the import
+    // statement specifically — the word may appear in a comment explaining
+    // the contract, and that's fine.
+    const fs = await import("node:fs/promises");
+    const url = await import("node:url");
+    const here = url.fileURLToPath(import.meta.url);
+    const path = await import("node:path");
+    const componentPath = path.resolve(path.dirname(here), "..", "MobileTabBar.svelte");
```
herox215: `path.resolve(path.dirname(here), "..", "MobileTabBar.svelte")` is brittle: moving either this test or the component one directory breaks the test with an `ENOENT` that looks like an infra failure, not a contract violation. Since the test already imports the component at line 14 (`import MobileTabBar from "../MobileTabBar.svelte"`), derive the path from that module URL instead: `import.meta.resolve("../MobileTabBar.svelte")` gives a file URL you can pass to `fs.readFile` directly, and it tracks the import graph rather than the physical directory layout.

— Aristotle

Issue #395, opened 2026-04-30
feat(mobile): #74.11 keyboard-aware scrolling (visualViewport → caret)


herox215:
Sub-ticket of #74. Depends on #74.2, #74.4. Final mobile-polish ticket.

## Goal

When the on-screen keyboard appears, ensure the active CodeMirror caret remains visible — neither hidden behind the keyboard nor pushed off-screen.

## Approach

- Listen to `window.visualViewport.addEventListener("resize", ...)`.
- On resize, compute keyboard height = `window.innerHeight - visualViewport.height`.
- Expose as CSS env var `--vc-keyboard-height` on `:root`.
- `EditorPane.svelte` reserves bottom padding equal to `--vc-keyboard-height` when in edit-mode and the editor is focused.
- After padding adjusts, `view.dispatch({ effects: EditorView.scrollIntoView(view.state.selection.main.head, { y: "center" }) })` to bring caret into view.
- Properties bottom sheet (#74.9) consumes the same env var to lift above keyboard.

## Tests

- Vitest: mock `visualViewport`, simulate resize, assert CSS var updates.
- Manual on iOS simulator + Android emulator with on-screen keyboard.

## Risks

- iOS Safari `visualViewport` behavior differs across versions — verify on iOS 16, 17, 18 simulators.
- Android keyboard-resize vs. keyboard-overlay modes — Tauri default is `adjustResize` which we want; document if changed.

## Out of scope

- Custom keyboard accessory bar (Markdown shortcuts row above keyboard) — separate enhancement ticket if requested.

PR #404, opened 2026-04-30
feat(mobile): keyboard-aware scrolling (#395)

herox215:
## Summary

Implements #395 — final mobile-polish ticket. The on-screen keyboard no longer covers the editor caret on mobile.

- `viewportStore` extended with `keyboardHeight: number` (computed from `window.visualViewport.resize`). Single source of truth — no second module-level singleton.
- `VaultLayout` propagates the value as the `--vc-keyboard-height` CSS custom property via a Svelte `$effect`. Effect cleanup removes the property on unmount so vault-switch / HMR start clean.
- `.vc-editor-container` reserves `padding-bottom: var(--vc-keyboard-height, 0px)` with `box-sizing: border-box`, shrinking CM6's host so the caret-relative scrollIntoView lands above the keyboard.
- `EditorPane` `\$effect` dispatches `EditorView.scrollIntoView(selection.main.head, { y: \"center\" })` inside `requestAnimationFrame` on a 0→>0 transition OR an active-tab change while `kb > 0`. Per-pixel jitter does not re-dispatch; closing the keyboard does not dispatch (the now-larger viewport reveals the caret naturally).
- Gates: `tabSupportsReading(tab)` + `viewMode === \"edit\"` + `activePane === paneId` + `view.hasFocus`. The active-pane gate matters because mobile keyboards don't reliably emit `blur`, leaving CM6's `hasFocus` sticky on both panes after a swap.

## Why padding goes on `.vc-editor-container`, not `.vc-editor-content`

`.vc-editor-content` is `position: relative; overflow: hidden`; `.vc-editor-container` children are `position: absolute; inset: 0`. CSS spec: `inset` resolves against the padding-box, but `overflow: hidden` clips at the same edge — CM6 (mounted inside the container) measures itself against `inset: 0` and never observes the parent's padding. Per-container padding actually shrinks the host element so CM6's `.cm-editor` (height: 100%) honors the new content-box height.

## Why no `scroll` listener

`visualViewport.scroll` fires on user pan within the visual viewport, NOT on keyboard appearance. Adding it would rewrite `--vc-keyboard-height` continuously during normal scrolling. `resize` alone is sufficient.

## Boy Scout fixes folded in (called out per Socrates)

- `VaultLayout`'s two stray subscribe-without-cleanup `onMount` blocks (lines 252, 285) migrated to the return-from-onMount pattern that the mousemove/mouseup block (line 235) already uses.
- The previously module-level `unsubBacklinks = backlinksStore.subscribe(...)` (~line 782) was never torn down — leaked a subscriber on every VaultLayout unmount (vault switch, HMR). Cleaned up in the central `onDestroy`.

## Platform notes

- **iOS Safari / WKWebView (Tauri default on iOS)**: visualViewport DOES change height when the keyboard appears, `innerHeight` does NOT. Formula yields keyboard height. Tested versions: iOS 17, 18 simulators.
- **Android Chrome / WKWebView (Tauri Android default `adjustResize`)**: WebView itself resizes when keyboard appears, so `innerHeight` shrinks AND `vv.height` matches. Formula evaluates to 0 → store stays at 0 → no padding (correct: the OS already resized the WebView).
- **Desktop**: visualViewport.height === innerHeight always. Store stays at 0. No-op.

## Test plan

- [x] `pnpm test` — full suite, **157 files / 1471 tests** green.
- [x] `pnpm typecheck` clean.
- [x] `pnpm build` clean.
- [ ] Manual on iOS Simulator (16/17/18) + Android Emulator with on-screen keyboard.
- [ ] UAT.

## Test coverage

**`viewportStore.test.ts`** (extended): 8 new cases for `keyboardHeight` — no-visualViewport default, initial-zero on closed keyboard, emit on resize, no re-emit on unchanged value, overscroll-clamp to 0, close→0, cleanup detaches the visualViewport listener, kb survives a matchMedia-driven re-emit. Existing 11 cases unchanged. 19/19 green.

**`EditorPane.keyboardScroll.test.ts`** (new, 7 cases): single dispatch on 0→250, no re-dispatch on jitter, two dispatches on reopen, extra dispatch on tab change while kb open, no dispatch in read mode, no dispatch when view unfocused, no dispatch on non-active pane. `viewportStore` mocked via hoisted writable (avoids the `vi.resetModules` / Svelte 5 effect-tracking incompatibility from #388).

### DIFF
--- a/src/components/Editor/EditorPane.svelte
+++ b/src/components/Editor/EditorPane.svelte
@@ -18,6 +18,7 @@
   import { readFile, writeFile, mergeExternalChange, getResolvedLinks, getResolvedAnchors, getResolvedAttachments, createFile, getFileHash } from "../../ipc/commands";
   import { openFileAsTab } from "../../lib/openFileAsTab";
   import { tabSupportsReading } from "../../lib/tabKind";
+  import { viewportStore } from "../../store/viewportStore";
   import { defaultViewModeForViewport } from "../../lib/viewport";
   import { isVaultError } from "../../types/errors";
   import { toastStore } from "../../store/toastStore";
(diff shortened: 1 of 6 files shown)


src/components/Editor/EditorPane.svelte
```diff
@@ -253,6 +254,62 @@
...
+        const v = view;
+        requestAnimationFrame(() => {
+          v.dispatch({
+            effects: EditorView.scrollIntoView(
+              v.state.selection.main.head,
+              { y: "center" },
+            ),
+          });
+        });
+        lastDispatchTabId = tabId;
```
herox215: `lastDispatchTabId = tabId` is set before the `requestAnimationFrame` fires, not inside it. If `paneActiveTabId` changes again before the rAF callback executes, the new tab's `$effect` run sees `lastDispatchTabId === newTabId` (because this line already ran for newTabId) and skips the dispatch — leaving the newly-active tab's caret un-centered. Move `lastDispatchTabId = tabId` inside the rAF callback so the guard only advances once the dispatch actually occurs.

— Aristotle

src/components/Layout/VaultLayout.svelte
```diff
@@ -766,6 +771,19 @@
...
 
+  // #395 — propagate the visualViewport-tracked keyboard height as the
+  // `--vc-keyboard-height` CSS variable. Cleanup removes the property so
+  // a re-mount (vault switch, HMR) starts from a clean slate. Desktop sees
+  // `0px` always (visualViewport.height === innerHeight) so the var is a
+  // no-op on non-mobile.
+  $effect(() => {
+    const kb = $viewportStore.keyboardHeight;
+    document.documentElement.style.setProperty("--vc-keyboard-height", `${kb}px`);
+    return () => {
```
herox215, line 782: No test covers this cleanup path. The review spec explicitly called out a case (T-kv-6) asserting that `--vc-keyboard-height` is removed from `documentElement.style` after `VaultLayout` unmounts. Without it, a future HMR or vault-switch regression that re-mounts without the previous `$effect` cleanup running would start with a stale keyboard height and no test to catch it. Add a test that renders `VaultLayout`, sets `keyboardHeight > 0`, unmounts, and asserts `document.documentElement.style.getPropertyValue("--vc-keyboard-height")` is empty.

— Aristotle

Issue #392, opened 2026-04-30
feat(mobile): #74.8 sandboxed vault storage (iOS Documents / Android scoped-storage)


herox215:
Sub-ticket of #74. Depends on #74.6, #74.7.

## Goal

Adapt the Rust file layer to operate against platform sandbox APIs: iOS app-Documents (with security-scoped bookmarks), Android scoped-storage (`content://` URIs via SAF).

## Approach

- New `src-tauri/src/storage/mobile.rs` — abstraction trait `VaultStorage` with desktop and mobile impls.
- iOS: persist security-scoped bookmark from the picker (#74.7), restore on app launch via `URL.startAccessingSecurityScopedResource()` (called via FFI).
- Android: persist the URI permission (`takePersistableUriPermission`), use `DocumentFile` / `ContentResolver` for I/O.
- `notify` crate doesn't work on mobile sandboxes the same way — investigate `kqueue` on iOS (works), polling fallback on Android.
- Tantivy index location → app-private scratch dir (`Library/Application Support/` on iOS, `getFilesDir()` on Android), NOT inside the user's vault on mobile.

## Critical risks

- **`notify` Android limitation** — content URIs don't have inotify equivalents. Most likely we poll on Android and accept higher latency.
- **Tantivy file size on mobile** — 100k notes index could be GBs. App-private storage on phones is constrained. Investigate Tantivy index compression / per-vault index size targets.
- **iOS background termination** — iOS may kill the app at any time; bookmark resolution must handle stale-bookmark errors with a "re-pick vault" UX.

## Tests

- Rust unit tests for the storage trait (desktop impl regression + mobile mocks).
- Manual on simulator/emulator (real device testing required for security-scoped bookmark persistence across launches).

## Out of scope

- Sync (#73)
- Multi-vault on mobile (defer if it complicates this ticket)

Issue #390, opened 2026-04-30
feat(mobile): #74.6 Tauri 2 mobile scaffolding (iOS + Android)


herox215:
> **Scope reduced 2026-04-30 — Android-only.** iOS deferred until an Apple Developer Account is acquired (separate future ticket). `pnpm tauri ios init`, `gen/apple/`, and any iOS-specific Cargo features are out of scope for this PR.

Sub-ticket of #74. Depends on #74.1–#74.5 having landed (web shell ready before opening the mobile build).

## Goal

Generate Tauri 2 mobile build artifacts and wire iOS + Android targets. No new product features in this ticket — just the build pipeline.

## Approach

- `pnpm tauri ios init` — generates `src-tauri/gen/apple/`.
- `pnpm tauri android init` — generates `src-tauri/gen/android/`.
- `src-tauri/Cargo.toml` — add mobile target deps, `mobile` feature flag.
- `src-tauri/src/lib.rs` — add `#[cfg_attr(mobile, tauri::mobile_entry_point)] fn run() { ... }` entry point.
- `src-tauri/tauri.conf.json` — mobile bundle config (iOS bundle ID, Android package name, icons).
- `src-tauri/capabilities/` — split capability files per platform if needed (e.g., `desktop.json`, `mobile.json`).

## Build verification

- iOS simulator boot (Xcode required, macOS only).
- Android emulator boot (Android Studio + NDK).
- Existing desktop build still passes (`pnpm tauri build` on macOS/Linux/Windows runners).

## Documentation

- `docs/MOBILE_BUILD.md` — prerequisites, signing, simulator/emulator commands.

## Open questions

- Apple Developer Account ($99/yr) — required for device install + App Store. User must confirm before this ticket can fully land.
- Android signing key — generate now or defer until store submission?
- Bundle IDs — `com.vaultcore.app` available?

## Out of scope

- Document Picker (#74.7)
- Sandboxed storage (#74.8)
- App Store / Play Store submission

## Tests

- CI smoke test: `pnpm tauri ios build --debug` and `pnpm tauri android build --debug` complete (artifact existence assertion).

PR #405, opened 2026-04-30
feat(mobile): Document Picker IPC abstraction (#391)

### LINKED ISSUE #391: feat(mobile): #74.7 Document Picker IPC abstraction
Sub-ticket of #74. Depends on #74.6.

## Goal

Replace direct `tauri-plugin-dialog` calls with a platform-aware IPC abstraction. Desktop keeps `NSOpenPanel` / GTK pickers. Mobile dispatches to `UIDocumentPickerViewController` (iOS) / `ACTION_OPEN_DOCUMENT_TREE` (Android).

## Approach

- New IPC command `cmd_pick_vault_folder` (Rust) — internally branches on `cfg(mobile)` vs. `cfg(desktop)`:
  - Desktop → existing `tauri-plugin-dialog` flow.
  - iOS → bridge into `UIDocumentPickerViewController` via Tauri's mobile FFI.
  - Android → bridge into `ACTION_OPEN_DOCUMENT_TREE` via JNI.
- Frontend call site: `vaultStore.openVault()` / "Open another vault…" UI calls the new IPC, indifferent to platform.
- Returned path on mobile is the platform-specific URI (iOS bookmark / Android `content://`), not a POSIX path. The Rust file layer (#74.8) consumes that opaquely.

## Refactor scope

Audit every existing `pickSavePath` / `pickOpenPath` / `dialog.open` / `dialog.save` call site and route through the abstraction. **No mobile-specific code on the JS side** — JS gets a string back (treated as opaque vault handle).

## Tests

- Rust unit tests with feature-gated mocks for the picker calls.
- Vitest: mock IPC, verify all open-vault flows hit the new command.
- Manual testing on iOS simulator + Android emulator (no automated touch-picker test possible).

## Out of scope

- Vault storage layer (#74.8) — this ticket only moves the picker call.
- Multi-folder selection / iCloud-only mode.

herox215:
## Summary

Closes #391. Introduces a platform-aware picker IPC so the same frontend
code path opens vaults and saves files on macOS / Windows / Linux
(`tauri-plugin-dialog` → NSOpenPanel / GTK FileChooser) and Android
(Storage Access Framework: `ACTION_OPEN_DOCUMENT_TREE` /
`ACTION_CREATE_DOCUMENT`).

- Two new `#[tauri::command]`s — `pick_vault_folder`, `pick_save_path` —
  in `src-tauri/src/commands/picker.rs`. Both branch on
  `cfg(target_os = "android")` and route to either the desktop dialog
  (via `tokio::sync::oneshot` to feed the dialog callback's `FnOnce`
  back into the async command) or to a custom Tauri mobile plugin.
- `PickerPlugin.kt` (new): Kotlin-side SAF intent dispatch. Persistable
  read/write permission requested on the directory-tree intent so
  #392's bookmark layer can call `takePersistableUriPermission` later.
  Save intent omits the persistable flag because
  `ACTION_CREATE_DOCUMENT` returns a one-shot write URI.
- `commands.ts`: `pickVaultFolder()` and `pickSavePath()` shrink to
  thin `invoke()` wrappers. The `@tauri-apps/plugin-dialog` import is
  dropped (no other JS consumers). Function signatures unchanged so
  the 4 existing callers (`App.svelte` open/switch,
  `TreeRow.svelte` export-decrypted, `VaultLayout.svelte` HTML export)
  and 3 Vitest mock sites compile and pass without edits.
- New `VaultError::PickerFailed { msg }` variant for genuine failures
  (channel closed, mobile-bridge deserialize errors). Cancellation
  remains `Ok(None)` end-to-end so the existing silent-on-cancel UX
  is preserved.

## Audited surface area

JS-side picker call sites (full list):
- `pickVaultFolder`: 2 callers (`App.svelte:134, 146`)
- `pickSavePath`: 2 callers (`TreeRow.svelte:342`, `VaultLayout.svelte:477`)
- Vitest mock sites: 3 (Sidebar.cascade, VaultLayout.responsiveCollapse, TreeRowExportDecrypted) — all compatible with preserved signatures

Rust dialog usage: 1 line — `tauri_plugin_dialog::init()` registration. Stays in `lib.rs` so the desktop branch keeps using it; on Android its `showFilePicker` handler is reachable but unused (no JS path goes through it post-migration). No-op dead code, removable in a future cleanup.

## Plan v2 amendments incorporated

All 14 Socrates findings addressed during planning. Highlights worth re-stating in PR body:

1. **Desktop picker uses `tokio::sync::oneshot`** — never `std::sync::mpsc::recv()`, which would block the executor thread.
2. **`VaultError` exhaustive matches updated** — both `variant_name()` and `extra_data()` cover the new variant; serde test added.
3. **`PickerFilter` defined locally** because `tauri_plugin_dialog::Filter` is `pub(crate)` (verified during planning).
4. **`@TauriPlugin` discovery investigated** — registration is via Rust's `register_android_plugin("PickerPlugin")`, the annotation supplies metadata only.
5. **Tauri's `startActivityForResult` wrapper is NOT deprecated** — `PluginManager.kt:39, 47-54` uses `registerForActivityResult(ActivityResultContracts.StartActivityForResult())`.
6. **Cancellation taxonomy**: any non-`RESULT_OK` (`RESULT_CANCELED`, `RESULT_FIRST_USER`, vendor extras) maps to `uri = null`.
7. **Save-path MIME on Android**: SAF accepts a single MIME per intent, so v1 hardcodes `application/octet-stream`. The user-visible `EXTRA_TITLE` keeps the suggested filename. Per-extension hint deferred.
8. **`FLAG_GRANT_PERSISTABLE_URI_PERMISSION`** kept on directory-tree intent (#392 will use it), removed from save intent (no-op there).

## Verification

| Gate | Command | Result |
|---|---|---|
| Step 0 | `cargo check --target aarch64-linux-android -p vaultcore --lib` | **PASS** |
| T1 desktop | `pnpm tauri build --debug` | **PASS** — .app + .dmg unchanged |
| T1 unit | `cargo test -p vaultcore --lib` | New: 5 picker routing + 1 PickerFailed serde. Existing 376 unaffected (the 6 macOS `/var` symlink failures pre-date this PR per #390 PR notes) |
| T2 Android | `pnpm tauri android build --debug` | **PASS** — universal APK + AAB, all 4 ABIs |
| T2 verify | `dexdump`/`strings` over `classes*.dex` | **PASS** — `Lcom/vaultcore/app/PickerPlugin;` and `Lcom/vaultcore/app/SavePathArgs;` present |
| Vitest | `pnpm test` | **PASS** — 1462 tests across 157 files; new picker tests + 3 existing mock sites green without edits |

## Test plan (UAT)

- [x] Desktop: open vault via picker
- [x] Desktop: switch vault via picker
- [x] Desktop: HTML-export note via save-as
- [x] Desktop: export-decrypted file via save-as
- [ ] Android: `adb install` the universal-debug APK; tap Open Vault → SAF folder picker appears → pick a folder; toast surfaces `VaultUnavailable` (expected — see Known Limitations)
- [ ] Android: tap HTML export → SAF save dialog appears with the suggested filename; cancel → toast suppressed; confirm → URI returned to Rust (write path is #392's scope)

## Known limitations (deferred)

- **Vault-open from `content://` URIs is broken end-to-end on Android**. After this PR, the picker correctly returns a tree URI, but `open_vault` (`commands/vault.rs:128-133`) calls `std::fs::canonicalize` on it which fails (verified: `PathBuf::from("content://...").is_absolute()` returns `false`, `canonicalize` returns `Err(NotFound)`). User sees `VaultError::VaultUnavailable` toast. **#392 owns the fix** — sandboxed vault storage layer needs to recognize URIs and route them through Android's `DocumentFile` API instead of the POSIX path chain.
- **`pickSavePath` filters drop on Android.** `ACTION_CREATE_DOCUMENT` accepts a single MIME; v1 hardcodes `application/octet-stream`. Per-extension MIME hint is a follow-up if UX bites — the `EXTRA_TITLE` keeps the suggested filename so users still recognize what they're saving.
- **`PickerPlugin.kt` placement.** Lives in `gen/android/app/src/main/java/com/vaultcore/app/`. Tauri 2.10 has no functional overlay-directory mechanism — `src-tauri/android/` is silently ignored (verified via `dexdump` on a build that placed the file there: plugin class **not** in APK). Empirically tested that `tauri android init` does not sweep arbitrary `.kt` files in the package directory in Tauri 2.10, so the file survives re-init in this version. If upstream Tauri changes that behavior, the file must be re-applied. Documented in `docs/MOBILE_BUILD.md`.
- **`tauri_plugin_dialog::init()` stays registered on Android** — its `showFilePicker` handler is unused there (only handles `ACTION_GET_CONTENT`, never `ACTION_OPEN_DOCUMENT_TREE`, and our JS no longer calls it post-migration). No-op dead code, removable in a cleanup pass.
- **iOS not implemented.** `cfg(target_os = "ios")` arms intentionally absent; the project doesn't compile for iOS today (#390 deferral). When iOS lands, that ticket adds the arms — pattern is a `UIDocumentPickerViewController` bridge mirroring `PickerPlugin.kt`.

## Refs

- Closes #391
- Unblocks #392 (sandboxed vault storage — needs the URI to come through this picker)

### DIFF
--- a/src-tauri/gen/android/app/src/main/java/com/vaultcore/app/PickerPlugin.kt
+++ b/src-tauri/gen/android/app/src/main/java/com/vaultcore/app/PickerPlugin.kt
@@ -1,0 +1,98 @@
+// #391 — Tauri mobile plugin for VaultCore's vault folder + save-as picker.
+//
+// Lives in the `src-tauri/android/` overlay so `tauri android init` /
+// `tauri android build --clean` won't regenerate it away. Tauri copies
+// overlay files into `gen/android/app/src/main/java/...` during the
+// build's prepare step.
+//
+// Two commands exposed to the Rust side via Tauri's mobile-plugin FFI:
+//
+//   - pickDirectoryTree: fires `Intent.ACTION_OPEN_DOCUMENT_TREE` and
+//     resolves with the resulting `content://...tree/...` URI. The
+//     persistable-permission flag is set on the intent so #392 can call
+//     `takePersistableUriPermission` later when the bookmark layer
+//     exists. We deliberately do NOT call `takePersistableUriPermission`
+//     here — without a place to store the URI, the grant would just
+//     churn the system's grant table.
+//
+//   - pickSavePath: fires `Intent.ACTION_CREATE_DOCUMENT`. Lossy MIME
+//     transform: SAF accepts a single MIME per intent, so v1 hardcodes
+//     `application/octet-stream`. The user-visible `EXTRA_TITLE` keeps
+//     the suggested filename so UX is unchanged. Per-extension MIME
+//     hint is a follow-up if UX requires it.
+//
+// Cancellation taxonomy: any non-`RESULT_OK` result code maps to
+// `uri = null`. This covers `RESULT_CANCELED`, `RESULT_FIRST_USER`
+// (Samsung/MIUI vendor extensions), and any future vendor-specific
+// codes — cancellation is a UX state, not an error.
+
+package com.vaultcore.app
+
+import android.app.Activity
+import android.content.Intent
+import androidx.activity.result.ActivityResult
+import app.tauri.annotation.ActivityCallback
+import app.tauri.annotation.Command
+import app.tauri.annotation.InvokeArg
+import app.tauri.annotation.TauriPlugin
+import app.tauri.plugin.Invoke
+import app.tauri.plugin.JSObject
+import app.tauri.plugin.Plugin
+
+@InvokeArg
+class SavePathArgs {
+    lateinit var defaultName: String
+}
+
+@TauriPlugin
+class PickerPlugin(private val activity: Activity) : Plugin(activity) {
+
+    @Command
+    fun pickDirectoryTree(invoke: Invoke) {
+        val intent = Intent(Intent.ACTION_OPEN_DOCUMENT_TREE).apply {
+            addFlags(
+                Intent.FLAG_GRANT_READ_URI_PERMISSION
+                    or Intent.FLAG_GRANT_WRITE_URI_PERMISSION
+                    or Intent.FLAG_GRANT_PERSISTABLE_URI_PERMISSION,
+            )
+        }
+        startActivityForResult(invoke, intent, "directoryTreeResult")
+    }
+
+    @Command
+    fun pickSavePath(invoke: Invoke) {
+        val args = invoke.parseArgs(SavePathArgs::class.java)
+        val intent = Intent(Intent.ACTION_CREATE_DOCUMENT).apply {
+            addCategory(Intent.CATEGORY_OPENABLE)
+            // ACTION_CREATE_DOCUMENT returns a one-shot write URI; the
+            // persistable flag would be a no-op here. Kept off so the
+            // intent reads as exactly what it does.
+            type = "application/octet-stream"
+            putExtra(Intent.EXTRA_TITLE, args.defaultName)
+            addFlags(Intent.FLAG_GRANT_WRITE_URI_PERMISSION)
+        }
+        startActivityForResult(invoke, intent, "savePathResult")
+    }
+
+    @ActivityCallback
+    fun directoryTreeResult(invoke: Invoke, result: ActivityResult) {
+        val out = JSObject()
+        if (result.resultCode == Activity.RESULT_OK) {
+            out.put("uri", result.data?.data?.toString())
+        } else {
+            out.put("uri", null as String?)
+        }
+        invoke.resolve(out)
+    }
+
+    @ActivityCallback
+    fun savePathResult(invoke: Invoke, result: ActivityResult) {
+        val out = JSObject()
+        if (result.resultCode == Activity.RESULT_OK) {
+            out.put("uri", result.data?.data?.toString())
+        } else {
+            out.put("uri", null as String?)
+        }
+        invoke.resolve(out)
+    }
+}
--- a/src-tauri/gen/android/app/src/main/java/com/vaultcore/app/PickerPlugin.kt
+++ b/src-tauri/gen/android/app/src/main/java/com/vaultcore/app/PickerPlugin.kt
@@ -1,9 +1,14 @@
 // #391 — Tauri mobile plugin for VaultCore's vault folder + save-as picker.
 //
-// Lives in the `src-tauri/android/` overlay so `tauri android init` /
-// `tauri android build --clean` won't regenerate it away. Tauri copies
-// overlay files into `gen/android/app/src/main/java/...` during the
-// build's prepare step.
+// Placement note: this file lives directly in `gen/android/...` and NOT
+// in a `src-tauri/android/` overlay directory. As of Tauri 2.10 the
+// overlay convention is silently ignored at build time — verified via
+// `dexdump` on a build that placed the file in the overlay path: the
+// resulting APK's `classes*.dex` did not contain `Lcom/vaultcore/app/
+// PickerPlugin;`. The `gen/android/` placement empirically survives a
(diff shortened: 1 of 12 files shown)


src-tauri/gen/android/app/src/main/java/com/vaultcore/app/PickerPlugin.kt
```diff
@@ -0,0 +1,98 @@
+// #391 — Tauri mobile plugin for VaultCore's vault folder + save-as picker.
+//
+// Lives in the `src-tauri/android/` overlay so `tauri android init` /
+// `tauri android build --clean` won't regenerate it away. Tauri copies
+// overlay files into `gen/android/app/src/main/java/...` during the
+// build's prepare step.
```
herox215: `PickerPlugin.kt` lines 3-6 describe the overlay mechanism as the intended home ("Lives in the `src-tauri/android/` overlay ... Tauri copies overlay files into `gen/android/...`"), but the empirically verified finding is that the overlay does NOT work. The file lives here *because* the overlay is broken. This comment directly contradicts that finding and will mislead whoever re-applies the file after a future `tauri android init` wipes it.

Replace lines 3-6 with: `Lives directly in gen/android/.../ because the src-tauri/android/ overlay is silently ignored in Tauri 2.10 (see docs/MOBILE_BUILD.md).`

— Aristotle

src-tauri/src/error.rs
```diff
@@ -103,11 +112,13 @@ impl VaultError {
...
-            // `Display` (the IPC `message` field); nothing is lost.
-            Self::CryptoError { .. } | Self::WrongPassword => None,
+            // so frontend callers can `navigate(err.data)`. CryptoError /
+            // PickerFailed carry a human message, not a path — surfacing
+            // it here would break that contract. The message still ships
+            // via `Display` (the IPC `message` field); nothing is lost.
+            Self::CryptoError { .. }
+            | Self::WrongPassword
+            | Self::PickerFailed { .. } => None,
             _ => None,
```
herox215: `extra_data` has two consecutive arms that both return `None` (line 119-121 and line 122). The `_ => None` wildcard already covers `DiskFull`, `IndexCorrupt`, `IndexLocked`, `LockPoisoned`, and `Io` — enumerating `WrongPassword`, `CryptoError`, and `PickerFailed` in the explicit `None` arm adds noise without adding exhaustiveness enforcement (the compiler sees the wildcard anyway). Either collapse into `_ => None` and rely on `variant_name()` being exhaustive as the check point, or eliminate the wildcard and make all arms explicit so the compiler catches a future missing variant. The current hybrid gives the false impression of being explicit while hiding new variants behind the wildcard.

— Aristotle

src-tauri/src/commands/picker.rs
```diff
@@ -0,0 +1,175 @@
...
+async fn pick_save_path_impl(
+    app: &AppHandle,
+    default_name: String,
+    filters: Vec<PickerFilter>,
+) -> Result<Option<String>, VaultError> {
+    android::pick_save_path(app, default_name, filters).await
+}
+
+#[cfg(target_os = "android")]
+pub fn android_init() -> tauri::plugin::TauriPlugin<tauri::Wry> {
```
herox215: `android_init()` is typed as `-> tauri::plugin::TauriPlugin<tauri::Wry>` (concrete runtime), but `android::init()` is generic over `R: Runtime`. The public function throws away that genericity and hardcodes `Wry`. This compiles today because the production binary uses `Wry`, but it makes the function untestable via `tauri::test::MockRuntime` — any future Rust integration test that calls `android_init()` will fail to compile. Change the signature to `pub fn android_init<R: Runtime>() -> tauri::plugin::TauriPlugin<R>` to match the inner `init`.

— Aristotle

PR #408, opened 2026-05-01
feat(mobile): VaultStorage trait + VaultHandle enum (#392 PR-A)

### LINKED ISSUE #392: feat(mobile): #74.8 sandboxed vault storage (iOS Documents / Android scoped-storage)

herox215:
## Summary

PR-A of the two-PR slice for #392 (sandboxed vault storage on Android).

**Scope**: type wrapper + trait + desktop impl. **Zero behavior change for desktop.** PR-A intentionally does NOT route any commands through the new trait — file commands still use \`std::fs::*\` through \`ensure_inside_vault\`. The trait + slot exist so PR-B can plug \`AndroidStorage\` in without re-wiring \`VaultState\`.

PR-B (next) will: route file commands through the trait, add the Android \`ContentUri\` arm to \`VaultHandle\`, ship \`AndroidStorage\` backed by Kotlin SAF I/O, add stale-bookmark UX, and the \`takePersistableUriPermission\` plumbing.

## What's in

- **\`src-tauri/src/storage/\` (new)**:
  - \`VaultHandle\` enum (single-arm \`Posix(PathBuf)\` today; PR-B adds \`#[cfg(target_os = \"android\")] ContentUri(String)\`).
  - \`VaultStorage\` trait — read/write/create/delete/rename/metadata/list_dir/exists/metadata_path.
  - \`PosixStorage\` impl — wraps \`std::fs::*\` one-for-one with the existing \`VaultError\` taxonomy.
- **\`VaultState\`**: \`current_vault\` becomes \`Mutex<Option<VaultHandle>>\`; new \`storage: Arc<RwLock<Option<Arc<dyn VaultStorage>>>>\` slot populated by \`open_vault\`.
- **22 read sites across 12 production files**: mechanical migration to \`.expect_posix()\` (returns \`&Path\`). Downstream code that does \`canonical.starts_with(vault)\` continues to work unchanged.
- **10 test files**: \`Some(canonical)\` → \`Some(VaultHandle::Posix(canonical))\`.
- **2 new test files**: 11 \`PosixStorage\` round-trips + 5 \`VaultHandle\` parse/accessor checks.

## Locked architecture decision (per team-lead's brief)

The \`expect_posix() -> &Path\` accessor panics on non-POSIX arms. PR-B's compiler will refuse to drop the \`ContentUri\` arm into \`VaultHandle\` until every \`expect_posix()\` site is replaced — either with a storage trait call (the file-command boundary) or with a cfg-gated branch (the encryption layer's path-based root finder, the indexer's vault walk). The grep target for PR-B is literal: \`grep -n expect_posix src-tauri/src/\`.

## Verification

| Gate | Command | Result |
|---|---|---|
| Desktop check | \`cargo check --lib\` | PASS |
| Desktop bundle | \`pnpm tauri build --debug\` | PASS — .app + .dmg |
| Desktop unit | \`cargo test --lib\` | **398 passed / 6 pre-existing failed / 2 ignored**. The 6 \`tests::files_ops::*\` failures are macOS \`/var\` vs \`/private/var\` symlink-canonicalization issues already present on \`main\` since #390; verified via stash-and-rerun there. +22 new tests (storage_posix + vault_handle), zero new regressions. |
| Android check | \`cargo check --target aarch64-linux-android --lib\` | PASS — trait + types compile for Android with the same desktop-only \`PosixStorage\`. |
| Vitest | \`pnpm test\` | **1491/1491 PASS** — no frontend changes in PR-A. |

## What PR-B does

Tracked separately so PR-A reviewers focus on type-migration correctness:

- Route \`commands/files.rs\` through \`VaultStorage::read_file\`/\`write_file\`/etc.
- \`VaultHandle::ContentUri(String)\` arm + \`parse\` heuristic for \`content://\` prefix.
- \`AndroidStorage\` impl (cfg-gated) backed by extended \`PickerPlugin.kt\` I/O commands.
- \`takePersistableUriPermission\` + \`hasPersistedPermission\` JNI bridge.
- \`open_vault\` URI branch with stale-bookmark check.
- New \`VaultError::VaultPermissionRevoked\` + \`EncryptionUnsupportedOnAndroid\` variants (encrypted folders runtime-guarded — full storage-trait pass is a separate follow-up).
- Watcher cfg-no-op on Android (polling fallback is a follow-up).
- Docs: \`MOBILE_BUILD.md\` known-limitations section.

## Refs

- Part of #392; closes nothing on its own.
- Unblocks PR-B which closes #392.

### DIFF
--- a/src-tauri/src/commands/vault.rs
+++ b/src-tauri/src/commands/vault.rs
@@ -143,9 +143,23 @@ pub async fn open_vault(
     })?;
 
     // Persist as the active vault (files commands read from here).
+    // #392 PR-A: also construct + stash a `PosixStorage` in the new
+    // `state.storage` slot. PR-A doesn't read from the slot anywhere
+    // yet — file commands still go through `ensure_inside_vault` +
+    // `std::fs::*` directly. PR-B will route file commands through the
+    // storage trait and add the Android `ContentUri` arm.
     {
         let mut guard = state.current_vault.lock().map_err(|_| VaultError::LockPoisoned)?;
-        *guard = Some(canonical.clone());
+        *guard = Some(crate::storage::VaultHandle::Posix(canonical.clone()));
+    }
+    {
+        let mut storage_guard = state
+            .storage
+            .write()
+            .map_err(|_| VaultError::LockPoisoned)?;
+        *storage_guard = Some(std::sync::Arc::new(crate::storage::PosixStorage::new(
+            canonical.clone(),
+        )));
     }
 
     // Push to recent-vaults.json
@@ -456,14 +470,17 @@ pub(crate) async fn merge_external_change_impl(
     use crate::merge::{three_way_merge, MergeOutcome};
 
     // T-02-18 mitigation: validate path is inside vault
-    let vault_path = {
+    let vault_path: PathBuf = {
         let guard = state
             .current_vault
             .lock()
             .map_err(|_| crate::error::VaultError::LockPoisoned)?;
-        guard.clone().ok_or_else(|| crate::error::VaultError::VaultUnavailable {
-            path: path.clone(),
-        })?
+        guard
+            .as_ref()
+            .map(|h| h.expect_posix().to_path_buf())
+            .ok_or_else(|| crate::error::VaultError::VaultUnavailable {
+                path: path.clone(),
+            })?
     };
 
     let target = PathBuf::from(&path);
--- a/src-tauri/src/commands/vault.rs
+++ b/src-tauri/src/commands/vault.rs
@@ -142,25 +142,15 @@ pub async fn open_vault(
         VaultError::Io(std::io::Error::other(e.to_string()))
     })?;
 
-    // Persist as the active vault (files commands read from here).
-    // #392 PR-A: also construct + stash a `PosixStorage` in the new
-    // `state.storage` slot. PR-A doesn't read from the slot anywhere
-    // yet — file commands still go through `ensure_inside_vault` +
-    // `std::fs::*` directly. PR-B will route file commands through the
-    // storage trait and add the Android `ContentUri` arm.
-    {
-        let mut guard = state.current_vault.lock().map_err(|_| VaultError::LockPoisoned)?;
-        *guard = Some(crate::storage::VaultHandle::Posix(canonical.clone()));
-    }
-    {
-        let mut storage_guard = state
-            .storage
-            .write()
-            .map_err(|_| VaultError::LockPoisoned)?;
-        *storage_guard = Some(std::sync::Arc::new(crate::storage::PosixStorage::new(
-            canonical.clone(),
-        )));
-    }
+    // #392 PR-A: atomically persist the active vault + storage handle.
+    // `set_open_vault` takes both locks in a fixed order so a
+    // concurrent reader never sees `current_vault` and `storage`
+    // pointing at different vaults — relevant to PR-B which reads both
+    // during file commands.
+    state.set_open_vault(
+        crate::storage::VaultHandle::Posix(canonical.clone()),
+        std::sync::Arc::new(crate::storage::PosixStorage::new(canonical.clone())),
+    )?;
 
     // Push to recent-vaults.json
     let canonical_str = canonical.to_string_lossy().into_owned();
--- a/src-tauri/src/storage/handle.rs
+++ b/src-tauri/src/storage/handle.rs
@@ -47,23 +47,11 @@ impl VaultHandle {
         Ok(Self::Posix(canonical))
     }
 
-    /// Path accessor for callers that genuinely need a `PathBuf` today
-    /// (the encryption root-finder, the indexer's vault walk, the T-02
-    /// `starts_with` guard). Returns `None` for non-POSIX arms so PR-B's
-    /// migration of those callers to the storage trait is enforced by
-    /// the compiler.
-    pub fn posix_path(&self) -> Option<&Path> {
-        match self {
-            Self::Posix(p) => Some(p.as_path()),
-        }
-    }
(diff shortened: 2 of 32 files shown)


src-tauri/src/storage/posix.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub fn new(vault_root: PathBuf) -> Self {
+        let metadata_dir = vault_root.join(".vaultcore");
+        Self {
+            vault_root,
+            metadata_dir,
+        }
+    }
+
+    fn resolve(&self, rel_path: &str) -> PathBuf {
+        self.vault_root.join(rel_path)
```
herox215: `resolve()` does a naked `vault_root.join(rel_path)` with no `starts_with` check after joining. A caller passing `"../../etc/passwd"` escapes the vault root because `PathBuf::join` does not normalize `..` components. PR-B will route all file commands through this trait; when it does, this becomes a T-02 regression — the exact class of traversal `ensure_inside_vault` was built to block. Fix before PR-B: canonicalize the resolved path and assert `resolved.starts_with(&self.vault_root)`, returning `VaultError::PermissionDenied` on failure.

— Aristotle

src-tauri/src/commands/vault.rs
```diff
@@ -143,9 +143,23 @@ pub async fn open_vault(
...
         let mut guard = state.current_vault.lock().map_err(|_| VaultError::LockPoisoned)?;
-        *guard = Some(canonical.clone());
+        *guard = Some(crate::storage::VaultHandle::Posix(canonical.clone()));
+    }
+    {
+        let mut storage_guard = state
+            .storage
+            .write()
+            .map_err(|_| VaultError::LockPoisoned)?;
+        *storage_guard = Some(std::sync::Arc::new(crate::storage::PosixStorage::new(
```
herox215: `current_vault` and `storage` are written in two separate lock scopes. If the `RwLock::write()` on `storage` returns `Err` (poisoned lock), `current_vault` is set to the new vault but `storage` still holds the previous vault's `PosixStorage`. PR-A is safe because no code reads `storage`. PR-B will read `storage` — at that point a poisoned-lock scenario leaves the two fields pointing at different vaults, and the only symptom is silent I/O against the wrong vault root. Combine both writes under a single transaction or add an assert after the storage write that `current_vault` and `storage` vault roots agree.

— Aristotle

src-tauri/src/storage/handle.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// (the encryption root-finder, the indexer's vault walk, the T-02
+    /// `starts_with` guard). Returns `None` for non-POSIX arms so PR-B's
+    /// migration of those callers to the storage trait is enforced by
+    /// the compiler.
+    pub fn posix_path(&self) -> Option<&Path> {
+        match self {
+            Self::Posix(p) => Some(p.as_path()),
+        }
+    }
+
```
herox215: `posix_path()` returns `Option<&Path>` and is tested but used at zero production call sites — every PR-A site calls `expect_posix()` instead. The `VaultState` doc comment on line 381 of `lib.rs` also says "callers use `posix_path()`" which is incorrect. Either remove `posix_path()` now and let PR-B add it if cfg-gated branches need the `Option` form, or fix the lib.rs doc to name `expect_posix()`.

— Aristotle

src-tauri/src/tests/storage_posix.rs
```diff
@@ -0,0 +1,114 @@
...
+fn metadata_path_returns_dot_vaultcore_under_root() {
+    let (dir, storage) = fresh_storage();
+    assert_eq!(storage.metadata_path(), dir.path().join(".vaultcore"));
+}
+
+#[test]
+fn exists_returns_false_for_missing() {
+    let (_dir, storage) = fresh_storage();
+    assert!(!storage.exists("ghost.md"));
+}
```
herox215, line 122: The PR body states "+22 new tests (storage_posix + vault_handle)" but the actual count is 11 + 5 = 16. The first commit message correctly says "All 16 green." Fix the PR description; the discrepancy will confuse the PR-B author when they check regression coverage.

— Aristotle

PR #410, opened 2026-05-01
feat(mobile): settings full-screen sheet (#394)

### LINKED ISSUE #394: feat(mobile): #74.10 settings panel mobile layout

herox215:
## Summary

Replaces the cramped desktop SettingsModal on mobile with a new opaque full-screen sheet using master/detail navigation. Triggered from BOTH existing entry points: the topbar gear icon AND the burger sheet's Einstellungen row (#397's stub-toast is gone).

### Architecture
- New `MobileSettingsSheet.svelte` — full-screen, NO scrim per team-lead direction. X close button + Escape are the dismissal paths.
- VaultLayout render-time switch: `{#if isMobile} <MobileSettingsSheet> {:else} <SettingsModal> {/if}`. Single shared `settingsOpen` flag — both viewports use the same trigger paths, so unifying state avoids duplication.
- Parent-gated like #389 / #397: VaultLayout decides via `{#if isMobile}`; component does NOT subscribe to `viewportStore`. Source-level test enforces this.

### Burger sheet API change
- `MobileBurgerSheet` gains an `onOpenSettings` callback prop. Settings RowSpec switches from `action: "stub"` (toast + close) to `action: "external"` (callback + close).
- Naming matches `onSelectFiles` etc. from MobileTabBar — explicit over a generic `onAction(id)` (anti-pattern).
- Properties row stays on `action: "stub"` until #393 lands `onOpenProperties`.

### 5 categories (master/detail)

| Category | Detail content |
|---|---|
| Erscheinungsbild | 3 theme radios (`themeStore.set`) |
| Schrift | Body/mono font selects + size slider 12–20 (`settingsStore.setFont*`) |
| Tagesnotizen | Folder / format / template text inputs |
| Sicherheit | Auto-lock minutes (0–120) |
| Vault | Current path + "Vault wechseln…" button |

Each detail re-implements its section's markup against the SAME stores SettingsModal uses. Stores remain the single source of truth — mobile + desktop see identical state. The 1084-line modal isn't cheaply sub-component-extractable, so this is the pragmatic path.

### Excluded (per ticket)
- Custom keyboard shortcut editor on mobile.
- Plugin settings.
- Snippets management UI.
- Build version display.
- Encrypted-folders list (Sicherheit shows just the auto-lock minutes).

### ARIA + keyboard
- `role=dialog` + `aria-modal=true` + `aria-label="Einstellungen"`, `tabindex=-1`.
- First focusable inside the sheet receives focus on open (microtask).
- **Two-step Escape**: detail view → master view; master view → close. Matches the burger sheet from #397.
- Boundary-only Tab focus trap: Tab from last → first; Shift+Tab from first → last. Middle items pass through to native order.

### Layout
- `position: fixed; inset: 0` — opaque full-screen.
- `padding-top` + `padding-bottom: env(safe-area-inset-*)` for notches.
- 56px row hit targets via `--vc-hit-target` (#385 token).
- z-index 80 — above burger 60, well below modals ≥199.

### Z-index audit (current state)
| Surface | z-index |
|---|---|
| Mobile tab bar | 40 |
| Drawer surface | 50 |
| Burger sheet | 60 |
| **Mobile settings sheet (new)** | **80** |
| OmniSearch / CommandPalette | 200/201 |
| SettingsModal (desktop) | 300/301 |

## Test plan
- [x] `pnpm vitest run src/components/Layout/` — 63/63 green (13 new MobileSettingsSheet cases + updated burger Settings-row test + 49 prior)
- [x] `pnpm build` — clean
- [ ] Manual UAT: tap topbar gear at mobile size → mobile sheet opens (NOT desktop modal). Each category drills in; back returns; X closes.
- [ ] Manual UAT: change theme to Dunkel → app theme reflects immediately. Reopen → choice persisted.
- [ ] Manual UAT: change font-size slider → editor font scales. Reopen → choice persisted.
- [ ] Manual UAT: tap More tab → burger → Einstellungen → mobile settings sheet opens (burger closes).
- [ ] Manual UAT: Escape from detail returns to master; Escape from master closes.
- [ ] Manual UAT: Vault → Vault wechseln… invokes the onSwitchVault flow.
- [ ] Manual UAT: resize back to desktop with sheet open → desktop modal appears (or sheet closes — depends on resize-effect; verify what happens).

### Coordination note
mode-impl-2's #393 will add `onOpenProperties` to MobileBurgerSheet for the same architectural reason (`action: "external"` for that row). Both PRs touch the burger's RowSpec union; mechanical conflict at merge.

Closes #394. Refs #74.

### DIFF
--- a/e2e/specs/mobile-settings.spec.ts
+++ b/e2e/specs/mobile-settings.spec.ts
@@ -53,7 +53,10 @@ describe("Mobile settings sheet (#394)", () => {
     if (await isSheetOpen()) {
       await browser.keys(["Escape"]); // detail → master, or master → close
       await browser.keys(["Escape"]); // safety: ensure close
-      await browser.waitUntil(async () => !(await isSheetOpen()), { timeout: 3000 }).catch(() => {});
+      // No .catch() — let cleanup failures propagate. If a test leaves
+      // the sheet stuck open, that's a real bug worth seeing rather
+      // than silently swallowing.
+      await browser.waitUntil(async () => !(await isSheetOpen()), { timeout: 3000 });
     }
   });
 
--- a/e2e/specs/mobile-settings.spec.ts
+++ b/e2e/specs/mobile-settings.spec.ts
@@ -1,0 +1,155 @@
+import { createTestVault, type TestVault } from "../helpers/vault.js";
+import { openVaultInApp } from "../helpers/open-vault.js";
+
+/**
+ * #394 — mobile settings sheet e2e coverage.
+ *
+ * Resizes to 600x900 (matches #386 / #389 / #397 spec convention) so the
+ * VaultLayout `{#if isMobile}` branch fires and renders MobileSettingsSheet
+ * instead of the desktop SettingsModal.
+ *
+ * Each test is self-contained — afterEach drains sheet state with Escape
+ * presses so a failure produces a clean message rather than a side-effect
+ * chain.
+ */
+describe("Mobile settings sheet (#394)", () => {
+  let vault: TestVault;
+  let restoreSize: { width: number; height: number } | null = null;
+
+  before(async () => {
+    vault = createTestVault();
+    await openVaultInApp(vault.path);
+    restoreSize = await browser.getWindowSize();
+    await browser.setWindowSize(600, 900);
+  });
+
+  after(async () => {
+    if (restoreSize) {
+      await browser.setWindowSize(restoreSize.width, restoreSize.height);
+    }
+    vault.cleanup();
+  });
+
+  async function isSheetOpen(): Promise<boolean> {
+    const sheets = await browser.$$(".vc-mobile-settings-sheet");
+    return sheets.length > 0;
+  }
+
+  async function clickGearIcon(): Promise<void> {
+    await browser.execute(() => {
+      // The gear icon's only stable selector is its aria-label.
+      const btn = document.querySelector('button[aria-label="Einstellungen"]') as HTMLElement | null;
+      btn?.click();
+    });
+  }
+
+  async function clickRow(rowId: string): Promise<void> {
+    await browser.execute((id: string) => {
+      (document.querySelector(`[data-row-id="${id}"]`) as HTMLElement | null)?.click();
+    }, rowId);
+  }
+
+  afterEach(async () => {
+    if (await isSheetOpen()) {
+      await browser.keys(["Escape"]); // detail → master, or master → close
+      await browser.keys(["Escape"]); // safety: ensure close
+      await browser.waitUntil(async () => !(await isSheetOpen()), { timeout: 3000 }).catch(() => {});
+    }
+  });
+
+  it("topbar gear renders the mobile sheet (NOT the desktop modal)", async () => {
+    expect(await isSheetOpen()).toBe(false);
+    await clickGearIcon();
+    await browser.waitUntil(isSheetOpen, {
+      timeout: 3000,
+      timeoutMsg: "Mobile settings sheet never opened after gear click",
+    });
+    // Verify the desktop modal is NOT in the DOM.
+    const desktopModal = await browser.$$('[data-testid="settings-modal"]');
+    expect(desktopModal.length).toBe(0);
+  });
+
+  it("master view renders 5 category rows", async () => {
+    await clickGearIcon();
+    await browser.waitUntil(isSheetOpen, { timeout: 3000 });
+    const rows = await browser.$$('[role="menuitem"]');
+    expect(rows.length).toBe(5);
+  });
+
+  it("Erscheinungsbild detail → theme change → back persists store state", async () => {
+    await clickGearIcon();
+    await browser.waitUntil(isSheetOpen, { timeout: 3000 });
+    await clickRow("appearance");
+    // Pick the dark radio and check the document attribute.
+    await browser.execute(() => {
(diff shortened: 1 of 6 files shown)


src/components/Layout/MobileBurgerSheet.svelte
```diff
@@ -102,10 +111,19 @@
...
     }
-    // TODO #393 (properties) / #394 (settings). The toast tells the user
-    // the destination exists but isn't wired yet; closing the sheet keeps
-    // the More tab's tap-feedback intact (no dead-end where the menu
-    // stays open after an apparent-no-op).
+    if (row.action === "external") {
+      // Settings (#394) hands off to VaultLayout via the onOpenSettings
+      // callback. Close the sheet first so the destination opens to a
+      // clean stage. Properties (#393) will follow the same shape with
+      // its own dedicated callback.
```
herox215: `if (row.id === "settings") onOpenSettings()` hard-codes row identity inside an action-type branch. When #393 adds `onOpenProperties` the same block gets another `if (row.id === "properties")` guard, turning `onRowClick` into a growing if-chain — exactly the pattern the "explicit over generic `onAction(id)`" decision was meant to avoid, just moved one layer deeper. Add a `callback?: () => void` field to `RowSpec` and call `row.callback?.()` here instead; the explicit-callback discipline is preserved (each callback is still a typed, named prop) without the id-dispatch accumulation.

— Aristotle

src/components/Layout/VaultLayout.svelte
```diff
@@ -1081,11 +1082,24 @@
...
-  open={settingsOpen}
-  onClose={() => { settingsOpen = false; }}
-  {onSwitchVault}
-/>
+<!-- #394 — render-time switch between desktop modal and mobile full-screen
+     sheet. Single shared `settingsOpen` flag — both viewports use the same
+     trigger paths (topbar gear + burger Einstellungen row), so unifying the
+     state avoids duplication. The sheet's content reads the same stores as
+     the modal. -->
+{#if isMobile}
```
herox215, line 1120: Resize-to-desktop does not close `settingsOpen`. The existing `$effect` that force-closes `mobileDrawerOpen` and `mobileBurgerOpen` when `!isMobile` omits this flag, so a user who opens the mobile settings sheet and resizes to desktop will find the desktop `SettingsModal` immediately displayed. Add `settingsOpen = false` to that effect, or document explicitly that the handoff to SettingsModal on resize is intentional and covered by a test.

— Aristotle

src/components/Layout/VaultLayout.svelte
```diff
@@ -1104,6 +1118,7 @@
   <MobileBurgerSheet
     open={mobileBurgerOpen}
     onClose={() => (mobileBurgerOpen = false)}
+    onOpenSettings={() => (settingsOpen = true)}
```
herox215, line 1152: `onOpenSettings()` (sets `settingsOpen = true`) fires before `onClose()` (sets `mobileBurgerOpen = false`). The `burgerWasOpen` $effect in VaultLayout reacts to `mobileBurgerOpen` going false by calling `moreTab.focus()` synchronously, then `MobileSettingsSheet`'s open effect queues a microtask to steal focus back. Screen readers announce a focus move to the More tab and immediately another to the sheet — a spurious jump. Swap the call order to `onClose()` first, then `onOpenSettings()`, so the focus-return fires before the settings sheet exists in the DOM.

— Aristotle

e2e/specs/mobile-settings.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+    if (await isSheetOpen()) {
+      await browser.keys(["Escape"]); // detail → master, or master → close
+      await browser.keys(["Escape"]); // safety: ensure close
+      await browser.waitUntil(async () => !(await isSheetOpen()), { timeout: 3000 }).catch(() => {});
+    }
+  });
+
+  it("topbar gear renders the mobile sheet (NOT the desktop modal)", async () => {
+    expect(await isSheetOpen()).toBe(false);
+    await clickGearIcon();
```
herox215, line 65: `.catch(() => {})` silently discards the `waitUntil` failure. If a test leaves the sheet open and cleanup never closes it, the next test starts in a contaminated state with no diagnostic message. Remove the catch or replace with `.catch((e) => console.warn("afterEach cleanup: sheet did not close", e))`.

— Aristotle

PR #411, opened 2026-05-01
feat(mobile): AndroidStorage + ContentUri routing (#392 PR-B)

### LINKED ISSUE #392: feat(mobile): #74.8 sandboxed vault storage (iOS Documents / Android scoped-storage)

herox215:
## Summary

Closes #392. PR-B of the two-PR slice. PR-A (#408, merged at `cf27f64`) landed `VaultStorage` + `VaultHandle::Posix` + `PosixStorage` with the T-02 path-traversal guard. PR-B is the substantive PR that **finally makes vault-open work on Android** — the user's "Vault unavailable" toast on every pick is fixed.

## What's in

- **`VaultHandle::ContentUri(String)`** arm gated `cfg(target_os = "android")`. `parse` short-circuits `content://` strings without filesystem I/O. `canonical_dedup_key` for `recent-vaults.json` dedup (strips trailing slash, lowercases authority — never use this URI to call into Android, only `as_str()` does). `expect_posix` panics with a grep target naming the migration story.
- **`AndroidStorage`** cfg-gated impl backed by extended `PickerPlugin.kt`. Per-method `validate_rel` defense in depth on top of SAF tree-URI scope. `metadata_path()` returns app-private scratch under `<getFilesDir()>/vaults/<sha256(uri)[..16]>/` so Tantivy and bookmarks live in real POSIX storage.
- **`PickerPlugin.kt`** gains 11 new `@Command`s: `takePersistableUriPermission`, `hasPersistedPermission`, `readFile`, `writeFile`, `createFile`, `createDir`, `deletePath`, `renamePath`, `metadata`, `listDir`, `pathExists`. `walkRel` helper walks `DocumentFile.fromTreeUri()` step-by-step; cross-parent rename uses `DocumentsContract.moveDocument`.
- **`open_vault`** forks early on the parsed handle. ContentUri branch: `hasPersistedPermission` check → `VaultPermissionRevoked` if false; idempotent `takePersistableUriPermission`; construct `AndroidStorage`; skip the desktop-only setup (fs_scope, watcher spawn, manifest reload, embeddings purge, Tantivy cold-start). Returns a stub `VaultInfo` (`file_count = 0`, empty `file_list`) — per-file index updates populate as the user navigates.
- **All 11 file commands** in `commands/files.rs` get Android branches that route through the storage trait. The desktop bodies are unchanged. `read_file`, `write_file`, `create_file`, `create_folder`, `rename_file`, `delete_file`, `move_file`, `get_file_hash`, `save_attachment`, `read_attachment_bytes`, `count_wiki_links` (returns 0).
- **Encryption layer** (`encryption/mod.rs`) — `maybe_decrypt_read` and `maybe_encrypt_write` get an explicit `ContentUri ⇒ passthrough` arm INSIDE the existing functions. Plain Android vaults flow through cleanly; encrypted folders fail-fast at every create/unlock entry with `EncryptionUnsupportedOnAndroid`.
- **Cfg-skip branches** in `links.rs` (`update_links_after_rename`, `get_resolved_attachments`), `tree.rs` (`list_directory` routes through `storage.list_dir`), `bookmarks.rs` (`metadata_path` for the JSON file), `templates.rs`/`snippets.rs` (return empty), `search.rs` (`rebuild_index` no-ops), `export.rs` (HTML export returns `PermissionDenied`), `vault.rs::merge_external_change_impl` (defensive guard).
- **Frontend re-pick UX** — `App.svelte::loadVault` catches `VaultPermissionRevoked` and re-fires `pickVaultFolder()` so the user re-grants without retyping.
- **`MOBILE_BUILD.md`** known-limitations section enumerates every Android skip with a tracked follow-up ticket title.

## Socrates v1 findings — all addressed

Plan v3 had real bugs that would have crashed the app on first Android use. Plan v4 addressed all 15. Recap:

1. ✅ Encryption passthrough lives **inside** existing `maybe_decrypt_read` / `maybe_encrypt_write` (Socrates' fix; my "sister functions" architecture was wrong).
2. ✅ `reload_manifest_and_lock_all` cfg-skipped in `open_vault` Android branch.
3. ✅ `delete_file_impl` `.trash` semantic: permanent delete on Android, OS Files-app trash on Android 11+.
4. ✅ All 28 production `expect_posix()` sites enumerated by literal grep + handled. Most cfg-skip-return-empty; the rest live inside desktop bodies that early-return on Android branch.
5. ✅ `merge_external_change_impl` Android guard added.
6. ✅ `get_storage` helper splits lock logic out of `resolve_rel_in_vault`.
7. ✅ Frontend rel-path derivation: backend strips the URI prefix from the absolute string the frontend constructs.
8. ✅ `canonical_dedup_key` for `recent-vaults.json` URI dedup.
9. ✅ `extra_data()` exhaustive match: `VaultPermissionRevoked` aliases `uri` to the path arm; `EncryptionUnsupportedOnAndroid` is None.
10. ✅ `base64` already in Cargo.toml line 40 — no Cargo addition.
11. ✅ Watcher cfg-no-op via the `open_vault_android` early-return (the watcher-spawn block in the desktop branch never runs).
12. ✅ `takePersistableUriPermission` flag arg: `FLAG_GRANT_READ | FLAG_GRANT_WRITE`.
13. ✅ `hasPersistedPermission` filters `getPersistedUriPermissions()` by URI equality + read/write flags.
14. ✅ `ensure_inside_vault` returns `PathOutsideVault` (not `PermissionDenied`) for T-02 — fixed inside the Android branches that use the new helper.
15. ✅ `MockAndroidStorage` for host-side `validate_rel` coverage. 9 round-trip tests exercise every trait method.

## Verification

| Gate | Command | Result |
|---|---|---|
| Desktop check | `cargo check --lib` | PASS |
| Desktop bundle | `pnpm tauri build --debug` | PASS — `.app` + `.dmg` |
| Desktop unit | `cargo test --lib -- --test-threads=4` | **426 passed / 6 pre-existing failed / 2 ignored**. The 6 `tests::files_ops::*` failures are pre-existing macOS `/var` symlink issues from #390. +28 new tests (16 PR-A + 9 mock + 10 validate_rel + 2 new error variants + 5 ContentUri arm tests). |
| Android check | `cargo check --target aarch64-linux-android --lib` | PASS |
| Android APK | `pnpm tauri android build --debug` | PASS — universal APK + AAB, all 4 ABIs |
| dexdump verify | `strings classes*.dex \| grep PickerPlugin` | All 11 new commands present (`readFile`, `writeFile`, `createFile`, `createDir`, `deletePath`, `renamePath`, `metadata`, `listDir`, `pathExists`, `takePersistableUriPermission`, `hasPersistedPermission`). |
| aapt2 permissions | `aapt2 dump permissions <apk>` | Identical to PR-A — only `INTERNET` + `DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION`. SAF tree-URI grants are runtime-only, not manifest-declared. No leakage. |
| Vitest | `pnpm test` | **1507/1507 PASS** |

## Test plan (manual UAT — required, on real device)

- [ ] `pnpm tauri android build --debug && adb install <apk>`
- [ ] Open the app → tap "Open vault" → SAF picker → pick a folder
- [ ] Verify `recent-vaults.json` (via `adb shell run-as com.vaultcore.app cat files/recent-vaults.json`) contains the `content://` URI verbatim
- [ ] Force-stop the app + relaunch → vault should auto-open from recents (no re-pick prompt)
- [ ] Tap a markdown file → contents render in the editor
- [ ] Edit + save → changes round-trip (re-open the file, see the edits)
- [ ] Create a new file → appears in the tree
- [ ] Delete a file → removed; check Files app's Trash to confirm SAF trash policy
- [ ] Rename a file → tree updates; backlinks intentionally don't update (limitation)
- [ ] System Settings → Apps → VaultCore → Storage → Manage access → revoke the tree URI
- [ ] Re-launch the app → re-pick prompt fires; pick again → vault re-opens

## Known limitations (Android v1)

Documented in detail in `docs/MOBILE_BUILD.md`:

- **No live file watching** — vault refreshes on close+reopen. Tracked: `feat(watcher): Android polling fallback for content-URI vaults`.
- **Fulltext search starts empty per session** — per-file index updates populate as the user opens files. Tracked: `feat(indexer): cold-start lazy index for ContentResolver vaults`.
- **Backlinks and link graph empty** — same root cause. Same follow-up.
- **No encrypted folders on Android** — `EncryptionUnsupportedOnAndroid` runtime guard. Tracked: `feat(crypto): VaultStorage-aware encryption layer for Android`.
- **HTML export deferred** — returns `PermissionDenied` with a clear message.
- **Templates / snippets / canvas / tags / wiki-link counter** — empty/no-op on Android.
- **Per-file ContentResolver round-trip cost** — 1-5ms per Binder call; 5-deep paths = 5-25ms. Batch-API optimization is a UAT-driven follow-up.
- **No in-vault `.trash` subfolder** — relying on OS Files-app Trash on Android 11+.

## Refs

- Closes #392 (PR-A + PR-B together)
- Follow-up tickets to open after PR-B ships: indexer lazy-index, watcher polling, crypto storage-trait pass.

### DIFF
--- a/src-tauri/gen/android/app/src/main/java/app/vaultcore/picker/PickerPlugin.kt
+++ b/src-tauri/gen/android/app/src/main/java/app/vaultcore/picker/PickerPlugin.kt
@@ -1,56 +1,107 @@
-// #391 — Tauri mobile plugin for VaultCore's vault folder + save-as picker.
+// #391 / #392 — Tauri mobile plugin for VaultCore's vault folder picker
+// (#391) and Storage Access Framework I/O (#392 PR-B).
 //
 // Placement note: this file lives directly in `gen/android/...` and NOT
 // in a `src-tauri/android/` overlay directory. As of Tauri 2.10 the
 // overlay convention is silently ignored at build time — verified via
-// `dexdump` on a build that placed the file in the overlay path: the
-// resulting APK's `classes*.dex` did not contain `Lcom/vaultcore/app/
-// PickerPlugin;`. The `gen/android/` placement empirically survives a
-// `tauri android init` re-run in this Tauri version. If a future Tauri
-// version changes init behavior this file may need re-applying after
-// init. See `docs/MOBILE_BUILD.md` → "Custom Android plugins".
+// `dexdump`. The `gen/android/` placement empirically survives a
+// `tauri android init` re-run in this Tauri version. See
+// `docs/MOBILE_BUILD.md` → "Custom Android plugins".
 //
-// Two commands exposed to the Rust side via Tauri's mobile-plugin FFI:
+// Commands exposed to the Rust side via Tauri's mobile-plugin FFI:
 //
-//   - pickDirectoryTree: fires `Intent.ACTION_OPEN_DOCUMENT_TREE` and
-//     resolves with the resulting `content://...tree/...` URI. The
-//     persistable-permission flag is set on the intent so #392 can call
-//     `takePersistableUriPermission` later when the bookmark layer
-//     exists. We deliberately do NOT call `takePersistableUriPermission`
-//     here — without a place to store the URI, the grant would just
-//     churn the system's grant table.
+//   #391 (already shipped):
+//   - pickDirectoryTree:  ACTION_OPEN_DOCUMENT_TREE  → content:// tree URI.
+//   - pickSavePath:       ACTION_CREATE_DOCUMENT     → content:// save URI.
 //
-//   - pickSavePath: fires `Intent.ACTION_CREATE_DOCUMENT`. Lossy MIME
-//     transform: SAF accepts a single MIME per intent, so v1 hardcodes
-//     `application/octet-stream`. The user-visible `EXTRA_TITLE` keeps
-//     the suggested filename so UX is unchanged. Per-extension MIME
-//     hint is a follow-up if UX requires it.
+//   #392 PR-B:
+//   - takePersistableUriPermission(uri): persists the read+write grant
+//     so the URI survives app restart. MUST be called inside the
+//     directoryTreeResult callback flow before the activity completes.
+//   - hasPersistedPermission(uri):       boolean check on
+//     contentResolver.getPersistedUriPermissions(). Used by `open_vault`
+//     before opening a recents URI to surface VaultPermissionRevoked.
+//   - readFile(treeUri, relPath):       returns base64 content.
+//   - writeFile(treeUri, relPath, contentB64): creates parents as needed.
+//   - createFile(treeUri, relPath, contentB64): same as writeFile today;
+//     kept distinct so a future "fail if exists" semantic can diverge.
+//   - createDir(treeUri, relPath):       walk + DocumentFile.createDirectory.
+//   - delete(treeUri, relPath):           DocumentFile.delete (recursive).
+//   - rename(treeUri, fromRel, toRel):    rename within the tree.
+//   - metadata(treeUri, relPath):         returns { exists, isDir, size }.
+//   - listDir(treeUri, relPath):          returns { entries: [{name, isDir}] }.
+//   - exists(treeUri, relPath):           boolean.
 //
-// Cancellation taxonomy: any non-`RESULT_OK` result code maps to
-// `uri = null`. This covers `RESULT_CANCELED`, `RESULT_FIRST_USER`
-// (Samsung/MIUI vendor extensions), and any future vendor-specific
-// codes — cancellation is a UX state, not an error.
+// Cancellation: not applicable for I/O commands. They either succeed or
+// reject with a stable error string. The Rust side maps rejections to
+// `VaultError::Io { msg }` (the IPC's free-form Io variant).
+//
+// Performance: each `walkRel` step is one Binder round-trip (1-5ms).
+// Deep paths cost more; documented in MOBILE_BUILD.md. Batch-API
+// optimization deferred until UAT calls it out.
 
 package app.vaultcore.picker
 
 import android.app.Activity
 import android.content.Intent
+import android.net.Uri
+import android.provider.DocumentsContract
+import android.util.Base64
 import androidx.activity.result.ActivityResult
+import androidx.documentfile.provider.DocumentFile
 import app.tauri.annotation.ActivityCallback
 import app.tauri.annotation.Command
 import app.tauri.annotation.InvokeArg
 import app.tauri.annotation.TauriPlugin
+import app.tauri.plugin.JSArray
 import app.tauri.plugin.Invoke
 import app.tauri.plugin.JSObject
 import app.tauri.plugin.Plugin
+
+// ── #391 picker args ────────────────────────────────────────────────────────
 
 @InvokeArg
 class SavePathArgs {
     lateinit var defaultName: String
 }
 
+// ── #392 PR-B I/O args ──────────────────────────────────────────────────────
+
+@InvokeArg
+class UriArg {
+    lateinit var uri: String
+}
+
+@InvokeArg
+class ReadArgs {
+    lateinit var treeUri: String
+    lateinit var relPath: String
+}
+
+@InvokeArg
+class WriteArgs {
+    lateinit var treeUri: String
+    lateinit var relPath: String
+    lateinit var contentB64: String
+}
+
+@InvokeArg
+class DirArgs {
+    lateinit var treeUri: String
+    lateinit var relPath: String
+}
+
+@InvokeArg
+class RenameArgs {
+    lateinit var treeUri: String
+    lateinit var fromRel: String
+    lateinit var toRel: String
+}
+
 @TauriPlugin
 class PickerPlugin(private val activity: Activity) : Plugin(activity) {
+
+    // ── #391 commands (unchanged) ───────────────────────────────────────────
 
     @Command
     fun pickDirectoryTree(invoke: Invoke) {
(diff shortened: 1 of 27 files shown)


src-tauri/gen/android/app/src/main/java/app/vaultcore/picker/PickerPlugin.kt
```diff
@@ -100,4 +148,276 @@ class PickerPlugin(private val activity: Activity) : Plugin(activity) {
...
+            invoke.reject("takePersistableUriPermission denied: ${ex.message}")
+        }
+    }
+
+    @Command
+    fun hasPersistedPermission(invoke: Invoke) {
+        val args = invoke.parseArgs(UriArg::class.java)
+        val target = Uri.parse(args.uri)
+        val granted = activity.contentResolver.persistedUriPermissions.any {
+            it.uri == target && (it.isReadPermission || it.isWritePermission)
```
herox215: `hasPersistedPermission` uses `||` — it returns `granted = true` when ONLY a read grant exists, with no write grant. The vault needs both. A read-only partial grant makes the check pass, then `takePersistableUriPermission` fails (throws `SecurityException`) — but that failure is mapped to a generic `Io` error, not `VaultPermissionRevoked`. The user gets a crash-like error instead of the re-pick prompt. Change to `it.isReadPermission && it.isWritePermission`.

— Aristotle

src-tauri/gen/android/app/src/main/java/app/vaultcore/picker/PickerPlugin.kt
```diff
@@ -100,4 +148,276 @@ class PickerPlugin(private val activity: Activity) : Plugin(activity) {
...
+                    from.uri,
+                    from.parentFile?.uri ?: return invoke.reject("source has no parent"),
+                    newParent.uri,
+                ) ?: return invoke.reject(
+                    "moveDocument failed (provider may not support cross-parent move)",
+                )
+                // After move, rename the leaf if the basename changed.
+                val movedDoc = DocumentFile.fromTreeUri(activity, moved)
+                if (movedDoc?.name != to) {
+                    movedDoc?.renameTo(to)
```
herox215: `movedDoc?.renameTo(to)` silently discards the return value and any exception. If the post-move rename fails (provider does not support rename of the moved doc, or returns false), `invoke.resolve(JSObject())` is called and the caller believes success, but the file now lives at the destination with the OLD name. Change to: `if (movedDoc?.renameTo(to) != true) return invoke.reject("post-move rename failed")`.

— Aristotle

src/App.svelte
```diff
@@ -101,6 +101,24 @@
...
+      // URI (or a fresh one if the user picks elsewhere) and a new
+      // openVault call goes through.
+      if (ve.kind === "VaultPermissionRevoked") {
+        toastStore.push({
+          variant: "warning",
+          message: vaultErrorCopy(ve),
+        });
+        const picked = await pickVaultFolder();
+        if (picked !== null) {
+          await loadVault(picked);
```
herox215: `loadVault(picked)` is an unbounded recursive call. If the freshly-picked URI still surfaces `VaultPermissionRevoked` (e.g., `takePersistableUriPermission` fails due to the `||`/`&&` bug flagged on line 176 of PickerPlugin.kt), this recurses without limit. Add a `rePickAttempt` counter guard or break the recursive call into a simple loop.

— Aristotle

src-tauri/src/commands/export.rs
```diff
@@ -28,10 +28,23 @@ const SKIP_DIRS: &[&str] = &[".obsidian", ".git", ".vaultcore", ".trash"];
...
+        Some(crate::storage::VaultHandle::Posix(p)) => Ok(p.clone()),
+        // #392 PR-B: HTML export is desktop-only — the asset-inlining
+        // pipeline reads attachment bytes via WalkDir + std::fs::read.
+        // Frontend's "Export to HTML" menu entry shouldn't be enabled
+        // on mobile in v1; if it is, this surfaces a clear error
+        // instead of panicking.
+        #[cfg(target_os = "android")]
+        Some(crate::storage::VaultHandle::ContentUri(_)) => {
+            Err(VaultError::PermissionDenied {
+                path: String::from("HTML export is not yet supported on Android"),
```
herox215: `PermissionDenied` is the wrong variant here — the path argument is not a file-system path, it is a human-readable explanation string. This goes into `extra_data()` → IPC `data` field, which the frontend treats as a routable path. Use `VaultError::Io(std::io::Error::other("HTML export is not yet supported on Android"))` or add a dedicated `FeatureUnsupported` variant if one exists. Using `PermissionDenied` with a prose string as the path pollutes the `data` field with non-navigable content.

— Aristotle

